PR #4, opened 2018-08-24
Fix syntax highlighting in README

jwilk:
* `bash` lexer is for Bash source code.
* `console` is for transcripts of Bash interactive sessions.

You want the latter in README.


jwilk:
> we ask that you sign our Contributor License Agreement

That's not going to happen.

PR #24, opened 2018-08-29
The introduction of golangci-lint during the CI

mattiabertorello:
This PR introduce golangci-lint in the pipeline to improve the code quality through the code static analysis.
I fixed the problems reported by the linters:
- misspell
- gofmt
- unconvert
- nakedret
- ineffassign
- interfacer
- varcheck
- lll
- golint except one problem where I have to modify also vendor dependencies 

These linters should be enable after fix the problems 
 - deadcode
  - dupl
  - errcheck
  - goconst
  - gocyclo
  - govet
  - maligned
  - megacheck
  - unparam

### DIFF
--- a/arduino/cores/cores.go
+++ b/arduino/cores/cores.go
@@ -111,9 +111,10 @@ func (platform *Platform) GetOrCreateRelease(version *semver.Version) (*Platform
 	return release, nil
 }
 
-// GetRelease returns the specified release corresponding the provided version,
+// GetReleaseRelaxedVersion returns the specified release corresponding the provided version,
 // or nil if not found.
-func (platform *Platform) GetRelease(version *semver.Version) *PlatformRelease {
+func (platform *Platform) GetReleaseVersion(version *semver.Version) *PlatformRelease {
+	// use as an fmt.Stringer
 	return platform.Releases[version.String()]
 }
 
@@ -124,7 +125,7 @@ func (platform *Platform) GetLatestRelease() *PlatformRelease {
 	if latestVersion == nil {
 		return nil
 	}
-	return platform.GetRelease(latestVersion)
+	return platform.GetReleaseVersion(latestVersion)
 }
 
 // GetAllReleasesVersions returns all the version numbers in this Platform Package.
--- a/arduino/cores/cores.go
+++ b/arduino/cores/cores.go
@@ -23,7 +23,7 @@ import (
 	"github.com/arduino/go-paths-helper"
 
 	"github.com/arduino/arduino-cli/arduino/resources"
-	properties "github.com/arduino/go-properties-map"
+	"github.com/arduino/go-properties-map"
 	"go.bug.st/relaxed-semver"
 )
 
@@ -111,9 +111,9 @@ func (platform *Platform) GetOrCreateRelease(version *semver.Version) (*Platform
 	return release, nil
 }
 
-// GetReleaseRelaxedVersion returns the specified release corresponding the provided version,
+// FindReleaseWithRelaxedVersion returns the specified release corresponding the provided version,
 // or nil if not found.
-func (platform *Platform) GetReleaseVersion(version *semver.Version) *PlatformRelease {
+func (platform *Platform) FindReleaseWithVersion(version *semver.Version) *PlatformRelease {
 	// use as an fmt.Stringer
 	return platform.Releases[version.String()]
 }
@@ -125,7 +125,7 @@ func (platform *Platform) GetLatestRelease() *PlatformRelease {
 	if latestVersion == nil {
 		return nil
 	}
-	return platform.GetReleaseVersion(latestVersion)
+	return platform.FindReleaseWithVersion(latestVersion)
 }
 
 // GetAllReleasesVersions returns all the version numbers in this Platform Package.
--- a/arduino/cores/status.go
+++ b/arduino/cores/status.go
@@ -113,11 +113,11 @@ func (targetPackage *Package) String() string {
 func (tdep ToolDependency) extractTool(sc Packages) (*Tool, error) {
 	pkg, exists := sc.Packages[tdep.ToolPackager]
 	if !exists {
-		return nil, errors.New("Package not found")
+		return nil, errors.New("package not found")
 	}
 	tool, exists := pkg.Tools[tdep.ToolName]
 	if !exists {
-		return nil, errors.New("Tool not found")
+		return nil, errors.New("tool not found")
 	}
 	return tool, nil
 }
@@ -129,7 +129,7 @@ func (tdep ToolDependency) extractRelease(sc Packages) (*ToolRelease, error) {
 	}
 	release, exists := tool.Releases[tdep.ToolVersion.String()]
 	if !exists {
-		return nil, errors.New("Release Not Found")
+		return nil, errors.New("release Not Found")
 	}
 	return release, nil
 }
@@ -143,15 +143,15 @@ func (packages *Packages) GetDepsOfPlatformRelease(release *PlatformRelease) ([]
 	for _, dep := range release.Dependencies {
 		pkg, exists := packages.Packages[dep.ToolPackager]
 		if !exists {
-			return nil, fmt.Errorf("Package %s not found", dep.ToolPackager)
+			return nil, fmt.Errorf("package %s not found", dep.ToolPackager)
 		}
 		tool, exists := pkg.Tools[dep.ToolName]
 		if !exists {
-			return nil, fmt.Errorf("Tool %s not found", dep.ToolName)
+			return nil, fmt.Errorf("tool %s not found", dep.ToolName)
 		}
 		toolRelease, exists := tool.Releases[dep.ToolVersion.String()]
 		if !exists {
-			return nil, fmt.Errorf("Tool version %s not found", dep.ToolVersion)
+			return nil, fmt.Errorf("tool version %s not found", dep.ToolVersion)
 		}
 		ret = append(ret, toolRelease)
 	}
--- a/arduino/cores/status.go
+++ b/arduino/cores/status.go
@@ -129,7 +129,7 @@ func (tdep ToolDependency) extractRelease(sc Packages) (*ToolRelease, error) {
 	}
 	release, exists := tool.Releases[tdep.ToolVersion.String()]
 	if !exists {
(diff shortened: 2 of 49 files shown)


arduino/cores/status.go
```diff
@@ -129,7 +129,7 @@ func (tdep ToolDependency) extractRelease(sc Packages) (*ToolRelease, error) {
 	}
 	release, exists := tool.Releases[tdep.ToolVersion.String()]
 	if !exists {
-		return nil, errors.New("Release Not Found")
+		return nil, errors.New("release Not Found")
```
cmaglie: this one could be all lower case I guess... don't know why it has been written this way...

arduino/resources/helpers.go
```diff
@@ -41,12 +41,12 @@ func (r *DownloadResource) IsCached(downloadDir *paths.Path) (bool, error) {
...
-	if exist, err := archivePath.Exist(); err != nil {
+	exist, err := archivePath.Exist()
+	if err != nil {
 		return false, fmt.Errorf("checking archive existence: %s", err)
-	} else {
-		return exist, nil
 	}
+
+	return exist, nil
 }
```
cmaglie: I don't know if I like better this syntax or the previous one.... but if the linter says that this is better... :-)

builder_client_helpers/examples.go
```diff
@@ -30,16 +30,16 @@ func ListExamplesPath() string {
 }
 
 // ListExamples provides a list of all the builtin examples
-func (c *Client) ListExamples(ctx context.Context, path string, maintainer *string, type_ *string) (*http.Response, error) {
```
cmaglie: Use `libType` instead of `type1`

commands/commands_test.go
```diff
@@ -35,7 +35,7 @@ import (
 
 // Redirecting stdOut so we can analyze output line by
 // line and check with what we want.
-var stdOut *os.File = os.Stdout
+var stdOut = os.Stdout // *os.File
```
cmaglie: Remove also the comment `// *os.File`

commands/commands_test.go
```diff
@@ -102,7 +102,7 @@ func executeWithArgs(t *testing.T, args ...string) (exitCode int, output []byte)
 	cmd.SetArgs(args)
 	cmd.Execute()
 
-	return
+	return exitCode, output
```
cmaglie: This removes the naked return but not the possibility to add more.
Maybe we should also remove the names from the return values to properly fix this?

commands/compile/compile.go
```diff
@@ -106,7 +106,7 @@ func run(cmd *cobra.Command, args []string) {
 
 	// Check for ctags tool
 	loadBuiltinCtagsMetadata(pm)
-	ctags, err := getBuiltinCtagsTool(pm)
+	ctags, _ := getBuiltinCtagsTool(pm)
```
cmaglie: May this function fail? Probably not but I think we should check the error anyway in case the behavior change in the future.

commands/compile/compile.go
```diff
@@ -116,7 +116,7 @@ func run(cmd *cobra.Command, args []string) {
 			formatter.PrintError(err, "Could not load hardware packages.")
 			os.Exit(commands.ErrCoreConfig)
 		}
-		ctags, err = getBuiltinCtagsTool(pm)
+		ctags, _ = getBuiltinCtagsTool(pm)
```
cmaglie: same here

arduino/cores/cores.go
```diff
@@ -111,9 +111,10 @@ func (platform *Platform) GetOrCreateRelease(version *semver.Version) (*Platform
 	return release, nil
 }
 
-// GetRelease returns the specified release corresponding the provided version,
+// GetReleaseRelaxedVersion returns the specified release corresponding the provided version,
 // or nil if not found.
-func (platform *Platform) GetRelease(version *semver.Version) *PlatformRelease {
+func (platform *Platform) GetReleaseVersion(version *semver.Version) *PlatformRelease {
```
cmaglie: I don't understand this change, may you explain it?

commands/config/dump.go
```diff
@@ -38,10 +38,11 @@ func initDumpCommand() *cobra.Command {
...
 
-var dumpFlags struct {
-	_default bool   // If false, ask questions to the user about setting configuration properties, otherwise use default configuration.
-	location string // The custom location of the file to create.
-}
+// This struct is unused
+//var dumpFlags struct {
+//	_default bool   // If false, ask questions to the user about setting configuration properties, otherwise use default configuration.
+//	location string // The custom location of the file to create.
+//}
```
cmaglie: this can be removed

commands/lib/uninstall.go
```diff
@@ -27,10 +27,11 @@ import (
...
 
-const (
-	versionAll    string = "all"
-	versionLatest string = "latest"
-)
+// These const are unused
+//const (
+//	versionAll    string = "all"
+//	versionLatest string = "latest"
+//)
```
cmaglie: remove

commands/config/init.go
```diff
@@ -39,8 +39,14 @@ func initInitCommand() *cobra.Command {
...
-	initCommand.Flags().StringVar(&initFlags.location, "save-as", "", "Sets where to save the configuration file [default is ./.cli-config.yml].")
+	initCommand.Flags().BoolVar(&initFlags._default,
+		"default",
+		false,
+		"If omitted, ask questions to the user about setting configuration properties, otherwise use default configuration.")
+	initCommand.Flags().StringVar(&initFlags.location,
+		"save-as",
+		"",
+		"Sets where to save the configuration file [default is ./.cli-config.yml].")
 	return initCommand
```
cmaglie: All these newline changes makes the commands `Flags()` initialization different with 3 different styles across the codebase:

STYLE 1 (all in one line):
```go
	command.Flags().StringVar(&flags.vidPid, "vid-pid", "", "When specified, VID/PID specific build properties are used, if boards supports them.")
```

STYLE 2 (break on the description):
```go
	command.Flags().StringVar(&flags.vidPid, "vid-pid", "",
		"When specified, VID/PID specific build properties are used, if boards supports them.")
```

STYLE 3 (break on all parameters):
```go
	initCommand.Flags().BoolVar(&initFlags._default,
￼ 		"default",
￼ 		false,
￼ 		"If omitted, ask questions to the user about setting configuration properties, otherwise use default configuration.")
```

IMHO we should change all of them, in the `commands/xxxx` modules (maybe to STYLE 2 since the line break is likely to happen on the description).

arduino/cores/status.go
mattiabertorello: I don't know I try to resolve only the problem that the linter reported me. In any case this is why should not be capitalized.
"Error strings should not be capitalized (unless beginning with proper nouns or acronyms) or end with punctuation, since they are usually printed following other context"
https://github.com/golang/go/wiki/CodeReviewComments#error-strings

commands/compile/compile.go
mattiabertorello: If you enable the errorcheck linter probably It will reported a miss check

arduino/cores/cores.go
mattiabertorello: Because I suppress the warning othewise I have to replace the type *semver.Version with fmt.Stringer
https://github.com/mvdan/interfacer#suppressing-warnings

commands/lib/uninstall.go
mattiabertorello: Done 833ded1ac65b7cf4ff5444fc77a26aa90259c104

commands/config/dump.go
mattiabertorello: Done 833ded1ac65b7cf4ff5444fc77a26aa90259c104

commands/commands_test.go
mattiabertorello: Done 3b2517935dda8b02d5eb42c7e42100b0841c89db

builder_client_helpers/examples.go
mattiabertorello: Done 3b2517935dda8b02d5eb42c7e42100b0841c89db

commands/config/init.go
mattiabertorello: Done 191cfc37d955d11e4700e8b1007e0dfa51ebc171

commands/commands_test.go
mattiabertorello: I don't know, I can remove it in declare as a normal variable, for me that function in general is difficult to read

mattiabertorello: Done 157bbbb6b0e39b064e5b1452648d34c42c009715

arduino/cores/cores.go
cmaglie: I see that the method has been renamed from `GetRelease` to `GetReleaseVersion`, what's exactly the warning you suppressed with this change?

arduino/cores/status.go
cmaglie: I understand, but since you lower-cased the `R` I think you want to lower case also the `N` and the `F`:
`release Not Found` -> `release not found`

mattiabertorello: Done 1da1d44d4858afc5f12e9b95d5cc2c6fb966a675

arduino/cores/cores.go
mattiabertorello: This is the error form the linter interfacer
```bash
arduino/cores/cores.go:116:38: `version` can be `fmt.Stringer` (interfacer)
func (platform *Platform) GetRelease(version *semver.Version) *PlatformRelease {
```
If you want to remove this error you have to use the interface instead the specific struct.
In this way
```go
func (platform *Platform) GetRelease(version fmt.Stringer) *PlatformRelease {
	return platform.Releases[version.String()]
}
```
Choose the solution that seems better for you, I thought that is necessary use the semver.Version so I prefer to rename the method

cmaglie: Ok, now I understand, changing to fmt.Stringer doesn't look like a great improvement to me.

My problem is that `GetReleaseVersion` looks like a method to get the `Version` of a `Platform`, instead this is a method to find a `PlatformRelease` with a specific version. Let's call it `FindReleaseWithVersion`.

mattiabertorello: Done 9ecd7dd688d0af6194971590e9a7598093384bd2

Issue #30, opened 2018-09-04
arduino-cli standalone install looks for .cli-config.yml in wrong location


uffejakobsen:
arduino-cli standalone install looks for .cli-config.yml in wrong location

 arduino-cli looks for .cli-config.yml in wrong location when installing a standalone arduino-cli binary into /usr/bin/
In the above case it looks for /usr/bin/.cli-config.yml - and does not even try to look inside my home dir

See the debug output:
```
# arduino-cli board listall --debug
INFO[0000] Initiating configuration                     
INFO[0000] Unserializing configurations from /usr/bin/.cli-config.yml 
WARN[0000] Error reading config, using default configuration  error="open /usr/bin/.cli-config.yml: no such file or directory"
WARN[0000] Did not manage to get config file, using default configuration  error="open /usr/bin/.cli-config.yml: no such file or directory"
INFO[0000] Checking if CLI is Bundled into the IDE      
INFO[0000] Candidate IDE Directory: /usr/bin            
INFO[0000] CLI is not bundled into the IDE              
INFO[0000] Configuration set                            
INFO[0000] Starting root command preparation (`arduino`) 
INFO[0000] Formatter set                                
INFO[0000] Loading the default Package index            
Board Name	FQBN
```

ubidefeo:
I believe the first place for `arduino-cli` to look for a config file should be the sketch directory.
I like to have custom config on a per-sketch basis.
Now I'm forced to pass it to the command line and it's a downer.

`ino compile -b arduino:samd:mkrzero --config-file ./cli-config.yml` (mine is aliased as `ino`, c'est plus facile)

my priority order would be
`.`
`$HOME/`
`/usr/local/etc`
and whatever comes after that :)

bxparks:
I think `cli-config.yml` should *not* be retrieved from the current directory at all, because it can cause confusing behavior.

The `arduino-cli compile` command allows the target `*.ino` file to be specified directly, instead of implicitly assumed to be in the current directory:
```
$ arduino-cli compile -b {board} path/to/file.ino
```

If reading from the `./cli-config.yml` is allowed, it seems more correct for the config file in the target directory `path/to/` directory to be used, in other words `path/to/.cli-config.yml`, instead of using the `./.cli-config.yml` file in the current directory from where the `arduino-cli` command was executed.

[...]

PR #140, opened 2019-01-31
Change the directory of the configuration files

matteosuppo:
If you install the arduino cli, the default directory
of the configuration file is the location of the executable
This is not an ideal configuration for linux or windows

With this commit the configuration file is searched in the
default configuration folder for each system


cmaglie:
I see that, on linux, the library searches in `~/.config/....` and  `/etc/xdg/...`.
`~/.config/arduino15` may have been a valid alterantive if used from the beginning but since we used `~/.arduino15` to store everything I'd like to continue to use that one. Also `/etc/xdg` seems a bit overkill to me, why the extra `xdg` sub-folder? `/etc/arduino-cli.conf` seems enough...

cmaglie:
Let's implement here some ideas from: https://github.com/arduino/arduino-cli/issues/30
To recap some points:

1) We should look for a configuration in some predefined places, for example on Linux:
- `/etc/`
- `~/.arduino15/`
- `${pwd}`

those directories are listed in increasing order of priority and any config file found should override the previous one (I mean the options defined in `${pwd}` should take priority on the ones in `/etc` for example).

2) we can navigate parent folders for `${pwd}` to search for a config file, this would allow to make a per-project configuration.

and this is added by me:

3) we should consider a better future-proof name for the config file like: `arduino-cli.conf` so we have a nice `/etc/arduino-cli.conf` instead of `/etc/.cli-config.yaml`

bxparks:
Do we need to worry about `$arduino_ide_dir/portable`, as described in https://www.arduino.cc/en/Guide/PortableIDE ? I don't use `~/.arduino15`, because I keep multiple versions of the IDE around for testing, and I don't want one version of the IDE clobbering the files and configs of another version.

cmaglie:
@bxparks we pushed forward portable support here: https://github.com/arduino/arduino-cli/pull/123

Also we started bundling the cli in the official Beta IDE, at the moment `arduino-cli` detects if it's bundled with the IDE and also if there is a portable folder:

```
~/Code/arduino/build/linux/work$ ls -l
total 29272
-rwxr-xr-x  1 cmaglie cmaglie      882 feb  1 10:05 arduino
-rwxr-xr-x  1 cmaglie cmaglie 13822120 feb  1 10:05 arduino-builder
-rwxr-xr-x  1 cmaglie cmaglie 16000485 feb  1 10:05 arduino-cli
-rwxr-xr-x  1 cmaglie cmaglie     5161 feb  1 10:05 arduino-linux-setup.sh
drwxr-xr-x 13 cmaglie cmaglie     4096 feb  1 10:05 examples
drwxr-xr-x  4 cmaglie cmaglie     4096 feb  1 10:05 hardware
-rwxr-xr-x  1 cmaglie cmaglie     9796 feb  1 10:05 install.sh
drwxr-xr-x  4 cmaglie cmaglie     4096 feb  1 10:05 lib
drwxr-xr-x 21 cmaglie cmaglie     4096 feb  1 10:05 libraries
drwxr-xr-x  6 cmaglie cmaglie     4096 feb  1 10:05 reference
-rw-r--r--  1 cmaglie cmaglie    87826 feb  1 10:05 revisions.txt
drwxr-xr-x  4 cmaglie cmaglie     4096 feb  1 10:05 tools
drwxr-xr-x  4 cmaglie cmaglie     4096 gen  3 17:15 tools-builder
-rwxr-xr-x  1 cmaglie cmaglie       86 feb  1 10:05 uninstall.sh
~/Code/arduino/build/linux/work$ ./arduino-cli config dump
proxy_type: auto
sketchbook_path: /home/cmaglie/Workspace/sketchbook-cores-beta
arduino_data: /home/cmaglie/.arduino15
board_manager:
  additional_urls:
  - http://arduino.esp8266.com/stable/package_esp8266com_index.json

~/Code/arduino/build/linux/work$ mkdir portable
~/Code/arduino/build/linux/work$ ./arduino-cli config dump
proxy_type: auto
sketchbook_path: /home/cmaglie/Code/arduino/build/linux/work/portable/sketchbook
arduino_data: /home/cmaglie/Code/arduino/build/linux/work/portable
board_manager: null
```

BTW we barely tested the "portable" scenario, (for example now I noticed the `board_mager: null` in the `config dump`) so it may require some more work.

matteosuppo:
In this latest batch of commits I implemented the following:

> we can navigate parent folders for `${pwd}` to search for a config file, this would allow to make a per-project configuration.

commands/config/init.go
```diff
@@ -65,7 +65,14 @@ func runInitCommand(cmd *cobra.Command, args []string) {
 	if filepath == "" {
 		filepath = commands.Config.ConfigFile.String()
 	}
-	err := commands.Config.SaveToYAML(filepath)
+
+	err := os.MkdirAll(commands.Config.ConfigFile.Parent().String(), 0766)
```
cmaglie: There should be a `MkdirAll` method directly from the `*paths.Path` object returned from `Parent()`.

configs/directories.go
```diff
@@ -19,23 +19,31 @@ package configs
...
-	if err != nil {
-		executablePath = "."
+	configDirs := configdir.New("arduino", "arduino-cli")
+
+	// Search for a suitable configuration file
+	path := configDirs.QueryFolderContainsFile("arduino-cli.yaml")
+	if path != nil {
+		return paths.New(path.Path, "arduino-cli.yaml")
 	}
-	return paths.New(executablePath).Parent().Join(".cli-config.yml")
```
cmaglie: Shall we warn that we changed the name? Maybe if an old config file is found arduino-cli should exit with an error asking to rename it?

configs/navigate.go
```diff
@@ -0,0 +1,24 @@
+package configs
```
cmaglie: the license is missing

configs/navigate_test.go
```diff
@@ -0,0 +1,55 @@
+package configs_test
```
cmaglie: the license is missing

configs/yaml_serializer.go
```diff
@@ -113,10 +109,10 @@ func (config *Configuration) SerializeToYAML() ([]byte, error) {
 			Password: config.ProxyPassword,
 		}
 	}
-	if len(config.BoardManagerAdditionalUrls) > 1 {
```
cmaglie: I've written `1` because when a new configuration is created with `configs.NewConfiguration` the `BoardManagerAdditionalUrls` is filled with the default `package_index.json` url.
I don't know if it's better to output the default index or not, maybe we should treat it as a special case and leave only 3rd party urls in `BoardManagerAdditionalUrls`?

matteosuppo: I mean it's called AdditionalUrls, so it makes sense to avoid it?

configs/directories.go
matteosuppo: Or we rename it silently?

cmaglie: I'd avoid such automatisms.

matteosuppo: Frankly we're still in alpha and no one is using it, so I wouldn't bother even returning an error

configs/yaml_serializer.go
cmaglie: I agree, but then we should check when it's used and add an extra call for the default URL.

matteosuppo: Sounds like something for another PR, I just restored the previous behaviour

PR #169, opened 2019-03-11
Use $HOME environment for user's home directory first

phd:
Use $HOME environment variable for user's home directory first.
Only then fallback to os/user (parsing /etc/passwd).

This allows user to use a different directory by exporting own value for the $HOME variable.

### DIFF
--- a/configs/directories.go
+++ b/configs/directories.go
@@ -19,6 +19,7 @@ package configs
 
 import (
 	"fmt"
+	"os"
 	"os/user"
 	"runtime"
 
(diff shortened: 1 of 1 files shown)


ultr:
bump

masci:
Hi @phd and thanks for the contribution!

We recently switched to go 1.12 and we can use https://golang.org/pkg/os/#UserHomeDir now, can you change your patch so we can get the home path in a cross platform fashion?

phd:
Hello @masci 
If I have access to my PC anytime soon, I will modify this patch and resubmit it.

phd:
I have created a new pull request with the updated patch:
https://github.com/arduino/arduino-cli/pull/239

Unfortunately this one chance makes arduino-cli compilation fail on Go 1.10.
And sadly Go 1.12 is not widely adopted yet, ex. Ubuntu has Go 1.10 available by default in 18.04, 18.10 and latest 19.04 releases: https://packages.ubuntu.com/search?keywords=golang&searchon=names&exact=1&suite=all&section=all (yes, you can easily install Go 1.12, still it's not very convenient)

masci:
Superseded by #239

Issue #205, opened 2019-05-10
Documentation issue with core install command


jeff-winn:
The core install documentation that is displayed within the console displays conflicting information when attempting to install a specific version.

>  \# download a specific version (in this case 1.6.9).
>  arduino-cli.exe core install arduino:samd=1.6.9

In the above documentation that shows when core install is used, the documentation indicates to use =1.6.9 whereas the command actually needs @1.6.9. This difference can be misleading as executing the command with =1.6.9 ends up displaying an error about a particular version not being found.

masci:
Fixed in #238

PR #206, opened 2019-05-10
Fixing documentation issue with core install command

jeff-winn:
Resolves issue #205 wherein the command documentation shows an = instead of an @ symbol as expected during execution.


masci:
Hi @jeff-winn sorry for the late reply, this fell through the cracks.

Function was moved under the `cli` package, do you think you can move the patch there? That would be great!

jeff-winn:
Not sure I follow what you’re asking of me, could you elaborate?

Thanks!

jeff-winn:
Disregard, I found what you were talking about. I’ll redo it and submit another pull request.

PR #223, opened 2019-06-03
Migration to go.mod for dependency management

cmaglie:
- use go.mod instead of `dep` for dependency management
- removed the `vendor` folder
- copied the `cli-inception` branch of arduino-builder in the `legacy/builder` folder (in the future this module will be migrated in the cli as a proper idiomatic go package)


.travis.yml
```diff
@@ -1,26 +1,18 @@
 language: go
 
 go:
-  - 1.11.x
+  - 1.12.x
 
 env:
-  - DEP_VERSION="0.5.0"
+  - GO111MODULE=on
```
masci: do we need this provided that Go version is 1.12 now?

common/formatter/text.go
```diff
@@ -22,8 +22,8 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/cmaglie/pb"
```
masci: the fork is diverging from upstream, any chance we can contribute instead of forking?

.travis.yml
cmaglie: it is needed for travis, it will be `on` by default from 1.13 AFAIK

common/formatter/text.go
cmaglie: Upstream library went from 1.x to 2.x (with API change) I don't think this patch will be accepted.
Maybe we can port directly to 2.x. Anyway I'm not bound to this particular library, we can evaluate other options. The features I missed much are:

* the progress bar should delete itself when reaching 100% so we have a clean output
* it should have no race conditions! (especially if there is a "rendering" go routine that runs in parallel with the download)

masci: Thanks for the explanation, we can have a look at v2 or alternatives later I guess

PR #226, opened 2019-06-05
Global build vars

rsora:
This PR aims to refactor a group of global variables in a `global` package in order to inject values at build time and use them in the CLI section of the cli

In order to have the values injected the use of `-ldflags` param is required in the `go build` phase otherwise a `missing` value will be shown (so an update in our CI is mandatory)

i.e. to inject the version value 

```
go build -ldflags "-X github.com/arduino/arduino-cli/global.version=0.3.6-alpha.preview"
```

### DIFF
--- a/global/global.go
+++ b/global/global.go
@@ -27,7 +27,7 @@ var appName = filepath.Base(os.Args[0])
 
 var (
 	application = "arduino-cli"
-	version     = "missing"
+	version     = "0.3.6-alpha.preview"
 	commit      = "missing"
 	cvsRef      = "missing"
 	buildDate   = "missing"
--- a/global/global.go
+++ b/global/global.go
@@ -9,14 +9,13 @@ import (
 var appName = filepath.Base(os.Args[0])
 
 var (
-	application= "arduino-cli"
-	version    = "missing"
-	commit     = "missing"
-	cvsRef     = "missing"
-	buildDate  = "missing"
-	repository = "missing"
+	application = "arduino-cli"
+	version     = "missing"
+	commit      = "missing"
+	cvsRef      = "missing"
+	buildDate   = "missing"
+	repository  = "missing"
 )
-
 
 func GetAppName() string {
 	return appName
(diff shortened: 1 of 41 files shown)


global/global.go
```diff
@@ -0,0 +1,57 @@
+package global
```
cmaglie: The license is missing

global/global.go
```diff
@@ -0,0 +1,57 @@
...
+import (
+	"os"
+	"path/filepath"
+)
+
+// appName is the command line name of the Arduino CLI executable on the user system (users may change it)
+var appName = filepath.Base(os.Args[0])
+
+var (
+	application= "arduino-cli"
```
cmaglie: uhm this file doesn't look go fmt-ed:
https://travis-ci.org/arduino/arduino-cli/builds/541797053#L499

I see also that there are some other errors from the linter:
https://travis-ci.org/arduino/arduino-cli/builds/541797053#L502

cmaglie:
I'm not sure about injecting values via ldflags, this is not very idiomatic.

If someone builds the cli from source with `go build github.com/arduino/arudino-cli` the version information is lost (well, on the other hand hardcoding the version gives only a rough approximation but at least you know from which version you're basing on).

Maybe we can keep the version hardcoded? it could be overriden by the CI anyway, so we get the best of the two approaches.

global/global.go
rsora: sorry, solved in 18c38772a02471e367ee7c0f5f553bd6e3f1ce44

global/global.go
rsora: sorry again, solved in 4d432faeea68f7b85335abf2560d3235d0106dfb

rsora:
@cmaglie in accordance to your idea

>  keep the version hardcoded? it could be overriden by the CI anyway, so we get the best of the two approaches.

added hardcoded fallback version in  0408a774874564284dd70cb271704fdc2fa02525

rsora:
PR https://github.com/arduino/arduino-cli/pull/227 is built upon this PR with heavy refactoring of the content.

In agreement with the colleagues this PR will be closed

PR #229, opened 2019-06-06
Introduce new testing layout

masci:
## Overview

The project currently has a robust test suite that exercises most of the functionalities against the actual http API. This is not very suitable to iterate fast on the code, thus this PR introduces the concept of unit tests in addition to the existing suite that will stay as it is but will be marked as "integration tests".

## Implementation

Integration test functions will contain the string `Integration` in the name, so that the suite can be run with `go test -run integration ./...`. Integration tests will be skipped when running unit tests and we use the `short` option from the Go standard lib to make this easy: each test function will begin like this:
```
	if testing.Short() {
		t.Skip("skip integration test")
	}
```
This way, unit tests will be run with just `go test -short ./...`

To make easier for contributors and the CI to run either of the suites, or both in sequence, a task runner was added. `task` is written in Go and can be easily installed by just `go get` so it should be fair to ask developers to have it in their dev env. Available tasks are defined in the `Taskfile.yml` file, with this PR you can do `task test-unit` to run all the unit tests, or (for example) `task test-unit TARGETS=./auth` to run all the unit tests for the `auth` package only. `task test-integration` is symmetric while `task test` will run both integration and unit tests.

Note that `go test ./...` will keep running all the tests so the use of `task` is not mandatory.

### DIFF
--- a/auth/auth_test.go
+++ b/auth/auth_test.go
@@ -57,7 +57,7 @@ func TestTokenIntegration(t *testing.T) {
 	}
 
 	// Obtain info
-	req, err := http.NewRequest("GET", "https://ddauth.arduino.cc/v1/users/byID/me", nil)
+	req, err := http.NewRequest("GET", "https://auth.arduino.cc/v1/users/byID/me", nil)
 	if err != nil {
 		t.Fatal(err)
 	}
--- a/auth/auth_test.go
+++ b/auth/auth_test.go
@@ -19,13 +19,13 @@ package auth_test
 
 import (
 	"encoding/json"
-	"flag"
 	"io/ioutil"
 	"net/http"
 	"os"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


auth/auth_test.go
```diff
@@ -49,7 +57,7 @@ func TestToken(t *testing.T) {
 	}
 
 	// Obtain info
-	req, err := http.NewRequest("GET", "https://auth.arduino.cc/v1/users/byID/me", nil)
+	req, err := http.NewRequest("GET", "https://ddauth.arduino.cc/v1/users/byID/me", nil)
```
matteosuppo: I think this is a typo

masci: good catch!

masci: (for context, tests are green because this specific test is skipped on Travis, found out when it was too late)

PR #239, opened 2019-06-09
Use os.UserHomeDir() for obtaining user's home directory

phd:
This allows user to use a different directory by exporting own value for the $HOME variable.

Follow up of https://github.com/arduino/arduino-cli/pull/169

### DIFF
--- a/configs/directories.go
+++ b/configs/directories.go
@@ -19,7 +19,7 @@ package configs
 
 import (
 	"fmt"
-	"os/user"
+	"os"
 	"runtime"
 
 	"github.com/arduino/go-paths-helper"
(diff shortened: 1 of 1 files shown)


phd:
Unfortunately this one chance makes arduino-cli compilation fail on Go 1.10.
And sadly Go 1.12 is not widely adopted yet, ex. Ubuntu has Go 1.10 available by default in 18.04, 18.10 and latest 19.04 releases: https://packages.ubuntu.com/search?keywords=golang&searchon=names&exact=1&suite=all&section=all (yes, you can easily install Go 1.12, still it's not very convenient)

masci:
Hi @phd thanks for the contribution and thanks for caring about compatibility. 

We'll be distributing binaries for the stable releases and we'll pick a specific Go version so the builds can be reproducible. The fact that `master` doesn't build anymore with an old version of Go has an impact on developers and contributors, not users. IMO asking a bit of extra effort to contributors to install a recent version of Go on systems where it's not yet available is a price we should pay to get a cleaner and more robust codebase.

Hope it makes sense :)

PR #246, opened 2019-06-13
Move i18n package outside legacy

masci:
Package is used in `arduino` root package, let's be explicity about it not being legacy.


cmaglie:
mmmh, IMHO just moving `i18n` outside of `legacy` is not enough to promote it to "non-legacy", there is a lot of non-idiomatic golang code there (the failing linter is also a signal of this).

Where is it referenced in `arduino/*`? I can't find it...

masci:
@cmaglie https://github.com/arduino/arduino-cli/blob/master/commands/compile/compile.go#L35

cmaglie:
That is the `commands/compile` (not `arduino`) that depends on `legacy/builder` to perform the build. The dependency tree is:

`commands/compile` -> `legacy/builder/*` -> `arduino/*`

Again, `legacy/builder` has a ton of non trivial stuff that needs a bit of refactoring before being integrated in `arduino/*`.

masci:
going to tackle this more organically, closing for now

PR #262, opened 2019-07-02
Move types package under arduino

masci:
Part of the undergoing work to clean up and extract code from the `legacy` package, this PR moves the `types` package under the `arduino` one along with:
* removing dead code
* adding unit tests for each type that was moved under `arduino/types`
* adding docstrings for the exported types


masci:
Closing in favor of a complete refactoring of the Command pattern

PR #268, opened 2019-07-09
Core search/list now return boards in a platform

32leaves:
This PR adds the list of boards supported by a platform, as well some more metadata to the response of `Core/List` and `Core/Search`. Further, does it harmonise the structures used by both operations to de-duplicate conversion code converting from the internal structures to the RPC messages.


commands/core/core.go
```diff
@@ -0,0 +1,37 @@
+package core
```
masci: Can you add license and copyright header?

commands/core/core.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"github.com/arduino/arduino-cli/arduino/cores"
+	rpc "github.com/arduino/arduino-cli/rpc/commands"
+)
+
+// platformReleaseToRPC converts our internal structure to the RPC structure.
+// Note: this function does not touch the "Installed" field of rpc.Platform as it's not always clear that the
+//       platformRelease we're currently converting is actually installed.
```
masci: nit: there's an extra tab at the beginning of this line

commands/core/core.go
```diff
@@ -0,0 +1,37 @@
...
+func platformReleaseToRPC(platformRelease *cores.PlatformRelease) *rpc.Platform {
+	boards := make([]*rpc.Board, len(platformRelease.Boards))
+	i := 0
+	for _, b := range platformRelease.Boards {
```
masci: let's do `for i, b := range platformRelease.Boards` and let `range` deal with the index increments

rpc/commands/core.proto
```diff
@@ -73,13 +73,7 @@ message PlatformSearchReq {
...
 
 message PlatformSearchResp {
-	repeated SearchOutput search_output = 1;
-}
-
-message SearchOutput {
-	string ID = 1;
-	string Version = 2;
-	string Name = 3;
+	repeated Platform search_output = 1;
```
masci: 🍰

commands/core/core.go
32leaves: That would be my preferred way, too. `Boards` is a `map[string]..` though, thus `i` would not be numeric index here unfortunately.

masci: oops my bad, didn't check what's been iterated here, scratch my comment!

PR #282, opened 2019-07-16
Added integration test

mastrolinux:
Cleaned up the whole mess with git.


test/test_main.py
```diff
@@ -0,0 +1,67 @@
...
+
+# Useful reference: 
+# http://docs.pyinvoke.org/en/1.2/api/runners.html#invoke.runners.Result
+
+
+def cli_line(*args):
+    # Accept a list of arguments cli_line('lib list --format json')
+    # Return a full command line string e.g. 'arduino-cli help --format json'
+    cli_full_line = ' '.join([cli_path, ' '.join(str(arg) for arg in args)])
+    # print(cli_full_line)
```
masci: let's remove this

test/test_main.py
```diff
@@ -0,0 +1,67 @@
...
+    result = run(cli_line(*args), echo=False, hide='out')
+    return result
+
+
+def test_command_help():
+    result = run_command('help')
+    assert result.ok
+    assert result.stderr == ''
+    assert 'Usage' in result.stdout
+    # result.out
```
masci: let's remove it

test/test_main.py
```diff
@@ -0,0 +1,103 @@
+from invoke import run, Responder
+import os
+import json
+import pytest
+import semver
```
rsora: `semver` has to be included in the `requirements.txt`

test/conftest.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+def pytest_collection_modifyitems(config, items):
+    if config.getoption("--runslow"):
+        # --runslow given in cli: do not skip slow tests
+        return
+    skip_slow = pytest.mark.skip(reason="need --runslow option to run")
+    for item in items:
+        if "slow" in item.keywords:
+            item.add_marker(skip_slow)
```
masci: why not using markers metadata directly? You can just do `pytest -m "not slow"` where slow tests need to be skipped

test/test_main.py
```diff
@@ -37,6 +38,9 @@ def test_command_lib_list():
     result = run_command('lib list', '--format json')
     assert '{}' == result.stdout
 
+# def test_command_lib_download():
+#     result = run_command('lib download')
+
```
masci: let's remove this before merging

test/test_main.py
```diff
@@ -46,10 +50,31 @@ def test_command_lib_install():
...
+    result = run_command('lib search')
+    out_lines = result.stdout.splitlines()
+    libs = []
+    # Create an array with just the name of the vars
+    for line in out_lines:
+        if 'Name: ' in line:
+            libs.append(line.split()[1].strip('\"'))
+    number_of_libs = len(libs)
+    # It would be strange to have less than 2000 Arduino Libs published
+    assert number_of_libs > 2000
```
masci: any way to avoid using a magic number? Can we search for a specific query instead and check for the expected number of items?

mastrolinux: Libraries are users' maintained and subject to changes or removal from the index. I would not put that in the tests. Maybe @cmaglie can help here. There could be some never changing libs

mastrolinux: solved, I added a check based on some libs found as output that are not going to change names in the next few years.

PR #286, opened 2019-07-17
gRPC interface to monitors

masci:
## What
This PR adds a bidirectional stream implementing a serial monitor through gRPC

## How
The concept of `Monitor` is introduced in the `arduino` package. A `Monitor` is an interface capable of reading a writing to a certain "target":
```
type Monitor interface {
	Close() error
	Read(bytes []byte) (int, error)
	Write(bytes []byte) (int, error)
}
```
A `Monitor` is used by a new gRPC service, different from the one exposing the cli commands, to let users pass a `stream` object that can be used to read/write from/to the monitor at the same time (bidirectional stream):
```
service Monitor {
  rpc StreamingOpen(stream StreamingOpenReq)
      returns (stream StreamingOpenResp) {}
}
```
In order to initialize a Monitor target (think about opening a serial port, or connecting to a socket) the first message of the stream **must** contain a special message containing the type of the target (for now only SERIAL), a string containing the target name, and a special field called `additionalConfig` that implements a JSON-like generic object, to be used to pass configuration options that are peculiar for specific targets:
```
message MonitorConfig {
  enum TargetType { SERIAL = 0; }

  string target = 1;
  TargetType type = 2;
  google.protobuf.Struct additionalConfig = 3;
}
```
Subsequent message will just contain data in the form of an array of bytes. Messages sent back from the Monitor will be similar, just containing data:
```
message StreamingOpenResp {
    bytes data = 1;
} 
```
## Serial monitor
This PR also includes a concrete implementation of the `Monitor` interface, specific to talk with a serial port:
```
// SerialMonitor is a monitor for serial ports
type SerialMonitor struct {
	port serial.Port
}
```
Upon receiving the first message, a serial port will be opened according to the configuration parameters, and 2 goroutines will be started to interact with the gRPC stream and the serial port respectively. The serial port will be gracefully closed in case the stream is closed from the client side.

## Example
To test a server started out of this branch you can use a sketch and client programs from https://gist.github.com/masci/804ba2679c8598630ca05919713ab106

## TODO
- [x] Add tests where possible
- [x] Double check for race conditions


masci:
Tests are failing because of a missing dep on the Docker image we use in the CI, fixing in a separate PR.

PR #293, opened 2019-07-19
Changed AdditionalSketchFilesCopier legacy command into new builder API function

masci:
Part of the ongoing refactoring of the `legacy` package.

Also changed few functions names in the new API to match the convention `builder.Sketch*` to better group parts of the API that operates on specific objects like a Sketch in this case (no need of a stricter design, it's just for readability).


arduino/builder/sketch.go
```diff
@@ -155,3 +156,60 @@ func MergeSketchSources(sketch *sketch.Sketch) (int, string) {
 
 	return lineOffset, mergedSource
 }
+
+// SketchCopyAdditionalFiles copies the additional files for a sketch to the
+// specified destination directory.
+func SketchCopyAdditionalFiles(sketch *sketch.Sketch, destPath string) error {
+	if err := os.MkdirAll(destPath, os.FileMode(0755)); err != nil {
```
cmaglie: at some point we may think to replace those path strings with `paths.Path` objects that have a much cleaner API to handle file operations.

arduino/builder/sketch.go
```diff
@@ -133,8 +134,8 @@ func LoadSketch(sketchPath, buildPath string) (*sketch.Sketch, error) {
 	return sketch.New(sketchFolder, mainSketchFile, buildPath, files)
 }
 
-// MergeSketchSources merges all the source files included in a sketch
-func MergeSketchSources(sketch *sketch.Sketch) (int, string) {
+// SketchMergeSources merges all the source files included in a sketch
+func SketchMergeSources(sketch *sketch.Sketch) (int, string) {
```
cmaglie: I'm wondering if these functions (`SketchMergeSources`, `SketchCopyAdditionalFiles`, etc.) may become methods of `sketch.Sketch`?

Issue #96, opened 2018-12-02
JSON Response to lib Commands


GItThatSparrow:
**AS A** User

**I NEED TO** determine the result of a call to the `arduino-cli` based on the response
   **AND** the command specifies the `--format=json` parameter

**SO THAT** I can distinguish the difference in between a successful operation and a failed operation

--------

# Example
I issue a command like
```
>> arduino-cli lib install AudioZero@1.0.0 --format=json
```

But the response is empty, like:
```
>> arduino-cli lib install AudioZero@1.0.0 --format=json

>>
```
## Equivalent example `download`
Same result when I issue a command like
```
>> arduino-cli lib download AudioZero --format=json   # or...
>> arduino-cli lib download AudioZero@1.0.0 --format=json
```

But the response is empty, like:
```
>> arduino-cli lib download AudioZero --format=json # or....

>>
>> arduino-cli lib download AudioZero@1.0.0 --format=json

>>
```
## Equivalent example `list`
Same result when I issue a command like
```
>> arduino-cli lib list --format=json
```

But the response is empty, like:
```
>> arduino-cli lib list --format=json

>>
```

## Equivalent example `uninstall`
Same result when I issue a command like
```
>> arduino-cli lib uninstall AudioZero@1.0.0 --format=json
```

But the response is empty, like:
```
>> arduino-cli lib uninstall AudioZero@1.0.0 --format=json

>>
```

## Equivalent example `update-index`
Same result when I issue a command like
```
>> arduino-cli lib update-index --format=json
```

But the response is empty, like:
```
>> arduino-cli lib update-index --format=json

>>
```

## Equivalent example `upgrade`
Same result when I issue a command like
```
>> arduino-cli lib upgrade --format=json
```

But the response is empty, like:
```
>> arduino-cli lib upgrade --format=json

>>
```

mastrolinux:
@philippejadin you can probably now use the new GRPC interface. This bug is still confirmed and I am not use if best to return an empty json or something like `{'success': 'true', 'message': 'some message here or empty string'}`

masci:
I think when invoking the CLI process the source of truth should always be the process exit code.

If the exit code is `0` operation was successful, otherwise you should make no assumptions about what's in the standard error/output: instead, you would look at the meaning of the exit code. Think about the CLI exiting because of a `panic` when likely the program wasn't able to properly print any output message, let alone a JSON structured message.

[...]

PR #297, opened 2019-07-22
Assert output is empty

masci:
Following up discussion at https://github.com/arduino/arduino-cli/issues/96 we ensure that when command has no output nothing will be printed out, even if the selected format was JSON (`docker` cli style).

### DIFF
--- a/test/test_main.py
+++ b/test/test_main.py
@@ -33,11 +33,14 @@ def test_command_help():
 
 
 def test_command_lib_list():
+    """
+    When ouput is empty, nothing is printed out, no matter the output format
+    """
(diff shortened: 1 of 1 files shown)


test/test_main.py
```diff
@@ -33,11 +33,14 @@ def test_command_help():
 
 
 def test_command_lib_list():
+    """
+    When ouput is empty, nothing is printed out, no matter the output format
+    """
     result = run_command('lib list')
     assert result.ok
     assert result.stderr == ''
```
mastrolinux: also change this to 

`assert '' == result.stderr`

to keep the style consistent

PR #302, opened 2019-07-23
Run integration tests in isolation

masci:
To avoid messing with the filesystem of the user running the tests, execute the suite in a temporary folder that'll be wiped after the tests finish.

### DIFF
--- a/test/test_main.py
+++ b/test/test_main.py
@@ -20,7 +20,7 @@ def test_command_lib_list(run_command):
     result = run_command('lib list')
     assert result.ok
     assert '' == result.stderr
-    result = run_command('lib list', '--format json')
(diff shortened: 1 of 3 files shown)


test/test_main.py
```diff
@@ -5,64 +5,45 @@
...
 
-def test_command_lib_list():
+def test_command_lib_list(run_command):
     """
     When ouput is empty, nothing is printed out, no matter the output format
     """
     result = run_command('lib list')
     assert result.ok
     assert '' == result.stderr
-    result = run_command('lib list', '--format json')
```
masci: this was the only function calling `run_command` with more than one arg, I've merged the two into a single list to match the rest of the code

PR #304, opened 2019-07-23
ctags arm64 osx64 support

rsora:
This PR updates the ctags metadata for the following platforms:
- `aarch64-linux` (arm 64 bit)
- `x86_64-apple` (macOS 64 bit)
In order to force the arduino-cli to download the correct version of the Arduino ctags binary tool (https://github.com/arduino/ctags) for each platform

***regarding the arm 64 bit platforms***
The PR completes the arm64 support effort started in https://github.com/arduino/arduino-cli/pull/204, solving the following error that happened when using the cli built and executed on an arm64 board (i.e. NVIDIA Jetson) 

```
downloading builtin:ctags@5.8-arduino11 tool: tool not available for your OS
```

This PR also solves https://github.com/arduino/arduino-cli/issues/292 together with Arduino CI pipeline configuration to enable the arm64 build.

***regarding the macOS 64 bit platforms***
This PR updates also the metadata for the macOS platform in order to download a ctag binary that complies with the 64bit transition for this OS, and stopping the "optimization" popup warning during the arduino-cli utilization on macOS High Sierra 10.13.4 and later. 
image

PR #307, opened 2019-07-24
updated README.md download links with linuxarm32 and linuxarm64 links

rsora:
This PR follows https://github.com/arduino/arduino-cli/pull/304

Readme links updated to include `linuxarm64` binaries links and renaming `linuxarm` to `linuxarm32` to have a clear distinction between the two architectures

Updated also instructions as the binaries in the archive are, from now on, always named `arduino-cli`


rsora:
@ArduinoBot build this please

Issue #306, opened 2019-07-24
[gRPC][discovery] panic: JSON decoder out of sync - data changing underfoot?


kittaakos:
### Current behavior

I am using the CLI via gRPC from a JS-based application. My client code performs mainly the following requests:
 - `PlatformListReq`,
 - `PlatformSearchReq`, and
 - `BoardListReq`.

Due to the async nature of the client, I am working on, and these requests might run in parallel which from time to time results in a panic error from the discovery code. After the failure, the CLI crashes, and the daemon process terminates.

```
panic: JSON decoder out of sync - data changing underfoot?
goroutine 42 [running]:
encoding/json.(*decodeState).value(0xc000130188, 0x46e4da0, 0xc001ffa390, 0x16, 0xa, 0xa)
    /usr/local/go/src/encoding/json/decode.go:367 +0x23f
encoding/json.(*decodeState).unmarshal(0xc000130188, 0x46e4da0, 0xc001ffa390, 0x0, 0xc0004b2480)
    /usr/local/go/src/encoding/json/decode.go:179 +0x209
encoding/json.(*Decoder).Decode(0xc000130160, 0x46e4da0, 0xc001ffa390, 0xc001ff6055, 0xc000eba9c0)
    /usr/local/go/src/encoding/json/stream.go:73 +0x187
github.com/arduino/arduino-cli/arduino/discovery.(*Discovery).List(0xc000eba9c0, 0xc001ffa330, 0xc0000cc1c0, 0x1, 0x1, 0x4e8a7c0)
    /Users/akos.kitta/git/arduino-cli/arduino/discovery/discovery.go:105 +0x230
github.com/arduino/arduino-cli/commands/board.List(0x4933f80, 0xc001ffa300, 0xc001ffa330, 0x48064e0, 0xc001ff8100, 0x49412c0)
    /Users/akos.kitta/git/arduino-cli/commands/board/list.go:38 +0x609
github.com/arduino/arduino-cli/commands/daemon.(*ArduinoCoreServerImpl).BoardList(0xc000208200, 0x4933f80, 0xc001ffa300, 0xc001ffa330, 0xc000208200, 0xc001ffa300, 0xc000075ba8)
    /Users/akos.kitta/git/arduino-cli/commands/daemon/daemon.go:52 +0x3f
[...]

PR #308, opened 2019-07-24
Sync board list to avoid race conditions

masci:
Fixes https://github.com/arduino/arduino-cli/issues/306

### DIFF
--- a/arduino/discovery/discovery.go
+++ b/arduino/discovery/discovery.go
@@ -23,6 +23,7 @@ import (
 	"io"
 	"os/exec"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/arduino/arduino-cli/arduino/cores"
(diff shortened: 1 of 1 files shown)


masci:
@ArduinoBot build this please

arduino/discovery/discovery.go
```diff
@@ -88,6 +90,7 @@ func (d *Discovery) Start() error {
 
 // List retrieve the port list from this discovery
 func (d *Discovery) List() ([]*BoardPort, error) {
+	d.Lock()
 	if _, err := d.in.Write([]byte("LIST\n")); err != nil {
 		return nil, fmt.Errorf("sending LIST command to discovery: %s", err)
```
cmaglie: If the method returns here with an error the structure is left locked.
It's preferable to use `defer` to unlock, since it automatically unlocks, no matter what the code flow.

```
{
   d.Lock()
   defer d.Unlock()
   ...do stuff with d...
}
```

masci:
@ArduinoBot build this please

PR #315, opened 2019-07-31
Update single libraries

masci:
Adding the ability to upgrade single libs vs. just upgrading all or nothing. This will be now possible:
```
arduino-cli lib upgrade WiFiNINA
```

Also in this PR (sorry it got too large)
* fix a bug preventing to show the latest version in `arduino-cli lib list` (an integration test will be added separately to avoid regressions)
* changed the API to avoid passing around useless parameters

### DIFF
--- a/cli/lib/upgrade.go
+++ b/cli/lib/upgrade.go
@@ -34,11 +34,14 @@ func initUpgradeCommand() *cobra.Command {
 	listCommand := &cobra.Command{
 		Use:   "upgrade",
 		Short: "Upgrades installed libraries.",
-		Long: "This command ungrades all installed libraries to the latest available version." +
-			"To upgrade a single library use the 'install' command.",
-		Example: "  " + os.Args[0] + " lib upgrade",
-		Args:    cobra.ArbitraryArgs,
-		Run:     runUpgradeCommand,
+		Long: "This command upgrades an installed library to the latest available version. " +
+			"If no arguments are provided, the command will upgrade all the installed libraries " +
+			"where an update is available.",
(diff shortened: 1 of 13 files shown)


cli/lib/upgrade.go
```diff
@@ -39,7 +37,7 @@ func initUpgradeCommand() *cobra.Command {
 		Long: "This command ungrades all installed libraries to the latest available version." +
```
rsora: (ultra-nitpicking)
There is a typo here: `ungrades` that should be `upgrades`.
Plus this string should be joined to the next one with a space because I have it printed in my terminal like this:
`This command ungrades all installed libraries to the latest available version.To upgrade a single library use the 'install' command.`
instead of be like this:
`This command ungrades all installed libraries to the latest available version. To upgrade a single library use the 'install' command.`
In addition, in order to differentiate from the `lib install` command mentioned in the message, we could add something like `To upgrade multiple libraries in a single command add the name of the libraries as parameters and separate them by a space` to clearly state that update multiple libraries is possible

cli/lib/upgrade.go
```diff
@@ -39,7 +37,7 @@ func initUpgradeCommand() *cobra.Command {
...
 			"To upgrade a single library use the 'install' command.",
 		Example: "  " + os.Args[0] + " lib upgrade",
```
rsora: The example could be enriched showing something like 
```
Examples:
 ./arduino-cli lib upgrade
./arduino-cli lib upgrade Audio
 ./arduino-cli lib upgrade Audio ArduinoJson
```
in order to show clearly that the update for multiple libraries at once is supported

PR #326, opened 2019-08-06
Create a compile+upload combo command

zmoog:
This PR adds something similar to the IDE’s “compile on upload” feature.

## Idea

The intended purpose is to simplify the workflow of the board user and offer a single-step command to compile the sketch and upload to the board.

For example:

```
$ arduino-cli compile --fqbn arduino:samd:nano_33_iot ~/Arduino/Blink --upload --port /dev/cu.usbmodem14101 --verify
```

## Design & Implementation

I've tried to reuse the existing implementation of both compile and upload command composing them in the Cobra commands.

Since I'm new to both the arduino-cli project and the Go programming language any feedback will be really, really (I mean it!) welcome.


test/test_compile.py
```diff
@@ -0,0 +1,120 @@
...
+import pytest
+import json
+import os
+
+from .common import running_on_ci
+
+
+def test_compile_without_fqbn(run_command):
+    # Init the environment explicitly
+    result = run_command("core update-index")
```
mastrolinux: Here you are mixing testing the "core" command and the compile command. It does not sound good to me. Eventually we should have a setup_test function running a `core update-index` just once. As of now we are running 
```    result = run_command("core update-index")
    assert result.ok

     # # Download latest AVR
    result = run_command("core install arduino:avr")
    assert result.ok```

over and over again. Please remove duplication of code making the tests running significantly slower too.

File: test/test_compile.py [hunk 2]
```diff
@@ -0,0 +1,120 @@
...
+    for port in ports:
+        boards = port.get('boards')
+        assert isinstance(boards, list)
+        for board in boards:
+            detected_boards.append(dict(address=port.get('address'), fqbn=board.get('FQBN')))
+
+    assert len(detected_boards) >= 1, "There are no boards available for testing"
+
+    # Build sketch for each detected board
+    for board in detected_boards:
```
mastrolinux: Here we compile for all the connected boards but we do not check if the proper core is installed: 
`run_command("core install arduino:avr")`
So we either install multiple cores or we decide to run the tests on specific AVR boards.

test/test_compile.py
zmoog: Yep, I agree.. and as said in person last week I'm currently moving the shared testing code into a dedicated fixture in pytest. This is happening in a separated branch with a general review of the integration test suite.

test/test_compile.py
zmoog: I'm updating the board auto-detection to install the core required to use te connected board(s). This is also happening in the separated branch with the general review of the integration test suite.

zmoog:
We're closing this PR to ship the compile+upload combo and the suggested improvements to the integration tests will be addressed in the dedicated PR (I will link it here as soon created).

PR #327, opened 2019-08-06
Clean up plugglable discovery and refactor serial-discovery usage

masci:
This PR removes the partial work to support pluggable discovery so it can be added back later with a dedicated series of PRs.

Since `serial-discovery` is the only discoverer used at the moment, its usage is now more explicit.

The API was cleaned up, removing unused parameters like the context.


commands/board/list.go
```diff
@@ -18,44 +18,50 @@
...
+		return nil, errors.Wrap(err, "unable to instance serial-discovery")
+	}
+
+	if err := serialDiscovery.Start(); err != nil {
+		return nil, errors.Wrap(err, "unable to start serial-discovery")
+	}
+	defer serialDiscovery.Close()
+
 	resp := &rpc.BoardListResp{Ports: []*rpc.DetectedPort{}}
-	for _, disc := range commands.GetDiscoveries(req) {
```
masci: instead of going through a dynamic loading mechanism that will always return just the serial discovery, we use it directly

PR #329, opened 2019-08-07
[skip changelog] Remove protobuf workaround on AppVeyor 

zmoog:
We can now switch back to a current version of the protocol since the
PR https://github.com/protocolbuffers/protobuf/pull/5044 has been
merged to master a released.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -29,7 +29,7 @@ install:
   # protobuf tooling needed at test time. We use a very old version of the compiler
   # because of this: https://github.com/protocolbuffers/protobuf/issues/3957
   - go get github.com/golang/protobuf/protoc-gen-go
-  - curl -o protoc.zip -LO https://github.com/protocolbuffers/protobuf/releases/download/v3.4.0/protoc-3.4.0-win32.zip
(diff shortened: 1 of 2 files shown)


zmoog:
@rsora the [build on AppVeyor](https://ci.appveyor.com/project/Arduino/arduino-cli/builds/26525763) looks fine with protobuf v3.9.1 .. what do you think?

appveyor.yml
```diff
@@ -29,7 +29,7 @@ install:
   # protobuf tooling needed at test time. We use a very old version of the compiler
   # because of this: https://github.com/protocolbuffers/protobuf/issues/3957
```
masci: Let's remove this notice now that we can get latest

masci:
I've edited PR title to avoid this to end in the changelog (it doesn't affect users)

appveyor.yml
zmoog: Should we get the latest also on the Docker image used for the unix builds on drone.io? Now it's using the v3.8.0.

masci: yeah let's consolidate versions across different platforms :+1:

PR #336, opened 2019-08-09
Query the backend to get the fqbn when a core is not installed

masci:
This PR makes the command `arduino-cli board list` properly work even when a core for the connected board is not installed. It works by querying the Arduino Cloud backend passing VID and PID, hence the CLI must be able to access the Internet for this to work.

### DIFF
--- a/commands/board/list.go
+++ b/commands/board/list.go
@@ -28,6 +28,37 @@ import (
 	"github.com/pkg/errors"
 )
 
+func apiByVidPid(url string) ([]*rpc.BoardListItem, error) {
+	retVal := []*rpc.BoardListItem{}
+	req, _ := http.NewRequest("GET", url, nil)
+	if res, err := http.DefaultClient.Do(req); err == nil {
+		body, _ := ioutil.ReadAll(res.Body)
+		res.Body.Close()
+
+		var dat map[string]interface{}
+		err = json.Unmarshal(body, &dat)
+		if err != nil {
+			return nil, errors.Wrap(err, "error processing response from server")
+		}
+
+		name, nameFound := dat["name"].(string)
+		fqbn, fbqnFound := dat["fqbn"].(string)
+
+		if !nameFound || !fbqnFound {
+			return nil, errors.New("wrong format in server response")
+		}
+
+		retVal = append(retVal, &rpc.BoardListItem{
+			Name: name,
+			FQBN: fqbn,
+		})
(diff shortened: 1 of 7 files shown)


commands/board/list.go
```diff
@@ -40,29 +92,51 @@ func List(instanceID int32) (*rpc.BoardListResp, error) {
...
+			url := fmt.Sprintf("https://builder.arduino.cc/v3/boards/byVidPid/%s/%s",
+				port.IdentificationPrefs.Get("vid"),
+				port.IdentificationPrefs.Get("pid"))
+			items, err := apiByVidPid(url)
+			if err == ErrNotFound {
+				// the board couldn't be detected, keep going with the next port
+				continue
+			} else if err != nil {
+				// this is bad, bail out
+				return nil, errors.Wrap(err, "error getting bard info from Arduino Cloud")
```
zmoog:
```suggestion
				return nil, errors.Wrap(err, "error getting board info from Arduino Cloud")
```

commands/board/list.go
```diff
@@ -18,13 +18,65 @@
...
+var (
+	// ErrNotFound is returned when the API returns 404
+	ErrNotFound = errors.New("board not found")
+)
+
+func apiByVidPid(url string) ([]*rpc.BoardListItem, error) {
+	retVal := []*rpc.BoardListItem{}
+	req, _ := http.NewRequest("GET", url, nil)
+	req.Header = globals.HTTPClientHeader
+	req.Header.Set("Content-Type", "application/json")
```
mattiabertorello: Please could you put the user agent and a user id if you have one
An example could be this one https://github.com/arduino/Arduino/blob/324a9bcbbbe540d133604767b5c4d30c5f62c753/arduino-core/src/cc/arduino/utils/network/HttpConnectionManager.java#L59
Example
```
final String defaultUserAgent = String.format(
      "ArduinoIDE/%s (%s; %s; %s; %s) Java/%s (%s)",
      BaseNoGui.VERSION_NAME,
      System.getProperty("os.name"),
      System.getProperty("os.version"),
      System.getProperty("os.arch"),
      System.getProperty("user.language"),
      System.getProperty("java.version"),
      System.getProperty("java.vendor")
    );
```

masci: The user agent is already there, see how `globals.HTTPClientHeader` is initialized

PR #345, opened 2019-08-13
Rewrite the output system

### LINKED ISSUE #34: GPIO-based DTR reset needed for upload on Raspberry Pi/UP2 board etc.
First of all, thanks for working on this project, I have been wanting a Golang-based CLI for the Arduino toolchain for the longest time.

It would appear that in order to upload code to an Arduino that is connected directly to the `ttyAMA0` serial port via the pins on a Raspberry Pi, UP2 board or other board, a GPIO pin-based DTR reset needs to be implemented.

Such like an implementation could be an additional flag named something like `gpioreset` that is conditional similar to what https://github.com/arduino/arduino-cli/blob/master/commands/upload/upload.go#L212 does, but that pulls the GPIO pin high/low similar to https://github.com/platformio/platformio-core/pull/687/files#diff-9df12ed969b3177c694a1020f4de5cc8R82 but of course allowing a specific GPIO pin to be specified.

For example:

```
arduino-cli upload -p /dev/ttyAMA0 -gpioreset 4 --fqbn atmel-avr-xminis:avr:atmega328 Arduino/MyFirstSketch
```

would use GPIO4 to perform the reset before attempting the upload.

Any thoughts on this?

masci:
This PR rewrites the whole logging and feedback system of the CLI. It also fixes #34 

# Rationale

This PR implements a blueprint where the logging system is detached from the output produced in response to user actions.

## Logging

The logging system is used by low level components to leave marks at any step of the code execution and it's intended to be mainly consumed by machines. Structured logging will be encouraged and a review of what we're logging right now, the logging levels and the structure of the messages will be eventually needed. To avoid passing around a logger object, the default `logrus` instance is used (this hasn't changed, it was already the case before this PR).

## Feedback

We call "Feedback" the system used to report an output in response to an user action. For the time being, errors go to the standard error and any other information goes to standard output. To keep consistency across the source code, a new object called `Feedback` was added to provide a `fmt`-like interface the CLI commands can use to show messages to the user. The object interface is purposely minimal and there's almost no logic. To fill the gap with logging, when an error is sent back to the user through the Feedback API, the message is also forwarded to the default logger.

## User interface

To put the user in control of the logging and feedback systems, CLI options changed as following:

* `-v`, `--verbose` option was added. When passed, the output of the logging system is displayed in the standard output. The logger output has colors to ease human browsing.
* `--log-file` option was added. When passed, the logger will write messages to the given file.
* --log-level` option was added to control which messages the log should forward. This applies to both the standard output when `-v` is passed, and the file on disk when `--log-file` is passed.
* `--debug` was removed. The closer option to that is now `arduino-cli -v --log-level debug`

## Additional notes

Almost 1k lines of code were removed for different reasons, most notably:

* ~ascii tables are now provided by https://github.com/cheynewallace/tabby that's just 70 lines of code and relies on the standard `tabwriter`~
* the `formatter` was only partially used and very often used inconsistently (different calls for the same goal), the `Formatter` is now used everywhere
* some of the "shortcuts" in `output` where swapped with more explicit code


gvarisco:
I like the rationale behind this PR. +1 from my side.

cli/board/details.go
```diff
@@ -47,42 +48,54 @@ func runDetailsCommand(cmd *cobra.Command, args []string) {
...
+	// Required tools: arduino:avr-gcc                 5.4.0-atmel3.6.1-arduino2
+	//                 arduino:avrdude                 6.3.0-arduino14
+	//                 arduino:arduinoOTA              1.2.1
+	//
+	// Option:         Processor                       cpu
+	//                 ATmega328P                    * cpu=atmega328
+	//                 ATmega328P (Old Bootloader)     cpu=atmega328old
+	//                 ATmega168                       cpu=atmega168
+	w := tabwriter.NewWriter(feedback.OutputWriter(), 0, 0, 2, ' ', 0)
+	table := tabby.NewCustom(w)
```
cmaglie: Two critical aspects I had to fight with the many tables renderes I tried before deciding to roll out my own:

- is colored output supported?
- how does it render very big tables (or tables that may have just a single very long cell among the other)?

a good test is the output of this command:
`./arduino-cli board details esp8266:esp8266:generic`

after installing the core here:
`http://arduino.esp8266.com/stable/package_esp8266com_index.json`

cli/core/list.go
```diff
@@ -55,11 +56,13 @@ func runListCommand(cmd *cobra.Command, args []string) {
...
 	if err != nil {
-		formatter.PrintError(err, "Error listing platforms")
+		feedback.Errorf("Error listing platforms: %v", err)
 		os.Exit(errorcodes.ErrGeneric)
 	}
 
-	if output.JSONOrElse(platforms) {
+	if globals.OutputFormat == "json" {
+		feedback.PrintJSON(platforms)
+	} else {
```
cmaglie: I'll add a comment here but it's valid for all the occurrence of this snippet of course :-)

Is it better to write something like:

```go
	if feedback.PrintJSONOrElse(platforms) {
￼		outputInstalledCores(platforms)
	}
```

or

```go
	if globals.OutputFormat == "json" {
		feedback.PrintJSON(platforms)
	} else {
		outputInstalledCores(platforms)
	}
```

This is boilerplate code that is duplicated everywhere (basically on each command that may potentially output json).
Every breaking change to the internals of the `feedback`/`globals.OutputFormat`, just to say, must be propagated on all occurrences of this snippet.

In this case, IMHO, less is more.

cmaglie:
I've added two comments, BTW it's great to see the `output` module finally eradicated!
Will give a try to this branch ASAP!

cli/board/details.go
masci: no bueno, bringing back the table...

masci: `table` is back, I've moved it in a root package in case we wanted to re-use it in other projects. I've also simplified a bit the layout and the color handling. For the latter, the tradeoff is having the public API depending on `color` explicitly vs. having to redefine ourselves all the colors, my proposal is to go with the former (specially because atm we only use green).

cli/core/list.go
masci: Let me see if I can find a simple way to remove the boilerplate before defending it :smile:

masci: Here is my proposal: let's move the responsibility to evaluate the format away from the commands by introducing a ["results container"](https://github.com/arduino/arduino-cli/blob/e57ae3ca3468c09d9434a125f67861c3ec17aaa6/cli/feedback/feedback.go#L39) that's very thin but generic enough to support different formats and that would be used only when highly customized outputs are required.

PROS:
* The boilerplate is gone, cfr [this example with no if-s at all](https://github.com/arduino/arduino-cli/blob/e57ae3ca3468c09d9434a125f67861c3ec17aaa6/cli/board/details.go#L54). As a side effect, the format can be removed from the globals.
* This approach is future proof, any new format we should add would have very little to zero impact on the existing commands
* When a complex formatting logic is needed, that stays with the command that needs it
CONS:
* When the output is complex, you have to [write a struct](https://github.com/arduino/arduino-cli/blob/massi/output/cli/board/details.go#L59) to provide the custom logic
* the feedback package gets more responsibilities but somebody has to decide which output to use

cmaglie: `PrintResult` is defined as:

```
func (fb *Feedback) PrintResult(res Result) {
```

but IIUC we need a function that accepts any type and do a runtime check to see if the argument is a `Result` (unless we want to always wrap the result in a `Result` but this will multiply the boilerplate by 10x), so something like this:

```
func (fb *Feedback) PrintResult(res interface{}) {
    if r, ok := res.(Result); ok {
        ...print using JSON(r.Data()) or r.String()...
    } else {
        ...print using JSON(res) or res.String()...
    }
}
```

Anyway, besides the above, the issue is another: there isn't a generic way to print complex outputs.

One thing is a simple primitive type like `int` or `string` another is a `Library` or a `Board` or a `[]PortList`, they all have their custom print with tables and verbosity that most of the time shows a subset of the available fields, sorted in some random way, so you're basically in always in the worst case (where you need to wrap inside a `Result`).

cli/board/details.go
cmaglie: sounds good! (for both, moving to a separate repo, and depending on `color`)

cli/core/list.go
masci: No type assertion, you always get a `Result` that can provide a string or an interface of some sort that's supposed to be marshalled (today in JSON, tomorrow we'll see). The "boilerplate" you add by always providing a `Result` object is just a struct definition plus a method definition (you always have an equivalent of the String method).

>Anyway, besides the above, the issue is another: there isn't a generic way to print complex outputs.
I know, I tentatively drafted a solution to see if I could remove the pattern
```
	if globals.OutputFormat == "json" {
		feedback.PrintJSON(ports)
	} else {
		outputListResp(ports)
	}
```
but I'm fine with that. Instead I'm not fine with something like `JSONOrElse` which is non-obvious, has a vague name and doesn't solve the root problem.

cmaglie: > but I'm fine with that. Instead I'm not fine with something like JSONOrElse which is non-obvious, has a vague name and doesn't solve the root problem.

Ok, let's keep the bigger if then, there is no point in discussing this further, we could always go back if we change our mind ;-)

PR #350, opened 2019-08-16
Fix panic when board attach args are incomplete

### LINKED ISSUE #165: SIGSEGV running 'board attach' with fqbn in 0.3.6-alpha.preview
```
$ arduino-cli board attach esp8266:esp8266:generic 
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x8d10c1]

goroutine 1 [running]:
github.com/arduino/arduino-cli/commands/board.runAttachCommand(0xc000137400, 0xc000065570, 0x1, 0x1)
	$GOCODE/src/github.com/arduino/arduino-cli/commands/board/attach.go:82 +0x1e1
github.com/arduino/arduino-cli/vendor/github.com/spf13/cobra.(*Command).execute(0xc000137400, 0xc000065540, 0x1, 0x1, 0xc000137400, 0xc000065540)
	$GOCODE/src/github.com/arduino/arduino-cli/vendor/github.com/spf13/cobra/command.go:766 +0x2ae
github.com/arduino/arduino-cli/vendor/github.com/spf13/cobra.(*Command).ExecuteC(0xc000136f00, 0x93dcec, 0xc00014ff88, 0xc000034118)
	$GOCODE/src/github.com/arduino/arduino-cli/vendor/github.com/spf13/cobra/command.go:852 +0x2c0
github.com/arduino/arduino-cli/vendor/github.com/spf13/cobra.(*Command).Execute(...)
	$GOCODE/src/github.com/arduino/arduino-cli/vendor/github.com/spf13/cobra/command.go:800
main.main()
	$GOCODE/src/github.com/arduino/arduino-cli/main.go:30 +0x28
```
I replaced the GOCODE portion of the paths, obviously.

The line of code in question is the assignment in this block:
```go
if fqbn != nil {
		sketch.Metadata.CPU = sketches.BoardMetadata{
			Fqbn: fqbn.String(),
		}
```
So it found the fqbn in the args, but can't turn it into a string? This is exactly the example command line for 'board attach' with a different fqbn. Any hints will be appreciated. 

Also, is there any documentation for 'board attach'?

masci:
Fixes #165 

The command `board attach` was configured to optionally take only one argument (the FQBN) but it's actually missing any logic to fill the gap (figuring out sketch name and path) when a sketch is not passed.

This PR enforces the sketch to be passed and fixes the command documentation accordingly.

### DIFF
--- a/cli/board/attach.go
+++ b/cli/board/attach.go
@@ -32,13 +32,13 @@ import (
 
 func initAttachCommand() *cobra.Command {
 	attachCommand := &cobra.Command{
-		Use:   "attach <port>|<FQBN> [sketchPath]",
+		Use:   "attach <port>|<FQBN> <sketchPath>",
 		Short: "Attaches a sketch to a board.",
 		Long:  "Attaches a sketch to a board.",
-		Example: "  " + os.Args[0] + " board attach serial:///dev/tty/ACM0\n" +
+		Example: "  " + os.Args[0] + " board attach serial:///dev/tty/ACM0 HelloWorld\n" +
 			"  " + os.Args[0] + " board attach serial:///dev/tty/ACM0 HelloWorld\n" +
-			"  " + os.Args[0] + " board attach arduino:samd:mkr1000",
-		Args: cobra.RangeArgs(1, 2),
+			"  " + os.Args[0] + " board attach arduino:samd:mkr1000 HelloWorld",
+		Args: cobra.ExactArgs(2),
 		Run:  runAttachCommand,
 	}
 	attachCommand.Flags().StringVar(&attachFlags.searchTimeout, "timeout", "5s",
@@ -52,16 +52,15 @@ var attachFlags struct {
 
 func runAttachCommand(cmd *cobra.Command, args []string) {
 	instance := instance.CreateInstance()
-	var path string
-	if len(args) > 0 {
-		path = args[1]
-	}
+	path := args[1]
+
 	_, err := board.Attach(context.Background(), &rpc.BoardAttachReq{
 		Instance:      instance,
 		BoardUri:      args[0],
 		SketchPath:    path,
 		SearchTimeout: attachFlags.searchTimeout,
 	}, output.TaskProgress())
+
 	if err != nil {
 		formatter.PrintError(err, "attach board error")
 		os.Exit(errorcodes.ErrGeneric)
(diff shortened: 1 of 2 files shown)


cmaglie:
The intended behaviour is to use the current folder if the sketch path is not specified (and I remember it worked in the past, this is surely a regression that was introduced during the GRPC implementation).

Other command that have the same behaviour:

https://github.com/arduino/arduino-cli/blob/master/cli/compile/compile.go#L82
https://github.com/arduino/arduino-cli/blob/master/cli/upload/upload.go#L65

also... this code is duplicated, it may be worth it to create a shortcut to avoid copying it everywhere :-)

masci:
Ok for the path but what the sketch name should be? (Also one might argue why the current folder and not the sketchbook).

I didn't know there were other commands following this pattern but if I got right how we want the behaviour of `sketch new` to be, this will clash with what we're assuming here, which is "when nothing is provided, assume current folder and xxx as name". 

As a side note, from the UX perspective I think any command requiring a sketch should be provided in input with a sketch name (so I can look in the sketchbook) or sketch path (so I can look in the fs).

cmaglie:
> Ok for the path but what the sketch name should be? (Also one might argue why the current folder and not the sketchbook).

In this case the name of the sketch will be exactly the same of the parent directory (i.e. if the current directory is `~/workspace/mysketch` the sketch name will be `mysketch`. This should be also validated by checking the existence of `~/workspace/mysketch/mysketch.ino`.

> but if I got right how we want the behaviour of sketch new to be, this will clash with what we're assuming here

Yes, I expressed exactly this concern in the JIRA task.

> from the UX perspective I think any command requiring a sketch should be provided in input with a sketch name (so I can look in the sketchbook) or sketch path (so I can look in the fs).

don't underestimate how handy can be omitting the sketch path or the sketch name, instead of

```
arduino-cli compile -b arduino:avr:uno sketch
arduino-cli upload -p /dev/ttyACM0 -b arduino:avr:uno sketch
```

you can do:

```
arduino-cli compile -b arduino:avr:uno
arduino-cli upload -p /dev/ttyACM0 -b arduino:avr:uno
```

or even (after `board attach`-ing):

```
arduino-cli compile
arduino-cli upload
```

masci:
Closing as this doesn't solve the root issue.

PR #364, opened 2019-08-28
Fix logging from low level operations

### LINKED ISSUE #174: JSON format not working properly with compile command
When using `--format json` with the `compile` command, the output is not properly formatted as JSON in `version 0.3.6-alpha.preview`
### Running
```bash
./arduino-cli compile --fqbn aruino:avr:uno Arduino/Test --format json
```
### Outputs
On stdout
```
Sketch uses 444 bytes (1%) of program storage space. Maximum is 30720 bytes.
```

### Running
```bash
./arduino-cli compile --fqbn aruino:avr:uno Arduino/Test --format json
```
### Outputs
On stdout
```
Build options changed, rebuilding all
{"Message":"Compilation failed.","Cause":"exit status 1"}
```
On stderr
```
/home/renaud/Arduino/Test/Test.ino: In function 'void setup()':
/home/renaud/Arduino/Test/Test.ino:3:1: error: 'a' was not declared in this scope
 a
 ^
```

masci:
Fixes #174 

This PR provides a wrapper around logrus to be used by the legacy package. This way, the global `-v` and `--log-level` options will be actually used by the low level packages under `legacy`, along with all the logging setup we already do at the CLI level.

### DIFF
--- a/commands/legacy_logger.go
+++ b/commands/legacy_logger.go
@@ -1,0 +1,78 @@
+// This file is part of arduino-cli.
+//
+// Copyright 2019 ARDUINO SA (http://www.arduino.cc/)
+//
+// This software is released under the GNU General Public License version 3,
+// which covers the main part of arduino-cli.
+// The terms of this license can be found at:
+// https://www.gnu.org/licenses/gpl-3.0.en.html
+//
+// You can be released from the requirements of the above licenses by purchasing
+// a commercial license. Buying such a license is mandatory if you want to modify or
+// otherwise use the software for commercial activities involving the Arduino
+// software without disclosing the source code of your own applications. To purchase
+// a commercial license, send an email to license@arduino.cc.
+
+package commands
+
+import (
+	"fmt"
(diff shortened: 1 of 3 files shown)


commands/legacy_logger.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/sirupsen/logrus"
+)
+
+// LegacyLogger wraps logrus but can be used within the `legacy` package
+type LegacyLogger struct{}
+
+// Fprintln is supposed to let the caller decide the target, let's not do it
+// and just forward to Println
+func (l LegacyLogger) Fprintln(w io.Writer, level string, format string, a ...interface{}) {
+	l.Println(level, format, a...)
```
cmaglie: I think that this will create some regressions because the parameter `w` must be considered to redirect the output to the correct stream... the legacy `i18n` module is really a mess, let's have a call to plan a way to migrate to a saner logging infra.

PR #366, opened 2019-08-28
Don't scan hidden directories in arduino-builder (.git, etc.)

earlephilhower:
Port of https://github.com/arduino/arduino-builder/pull/328

When checking for updated core files, don't scan inside UNIX hidden
directories like .git.  Scanning inside a GIT tree for a large project
with many branches could take many minutes of runtime and multiple GB of
RAM.  This was seen in the Arduino core for the ESP8266, but is probably
also an issue for other core development teams.

Fixes https://github.com/arduino/arduino-builder/issues/327

Signed-off-by: Earle F. Philhower, III <earlephilhower@yahoo.com>

### DIFF
--- a/legacy/builder/builder_utils/utils.go
+++ b/legacy/builder/builder_utils/utils.go
@@ -155,6 +155,7 @@ func findAllFilesInFolder(sourcePath string, recurse bool) ([]string, error) {
 
 		for _, folder := range folders {
 			if !utils.IsSCCSOrHiddenFile(folder) {
+				// Skip SCCS directories as they do not influence the build and can be very large
 				otherSources, err := findAllFilesInFolder(filepath.Join(sourcePath, folder.Name()), recurse)
 				if err != nil {
 					return nil, i18n.WrapError(err)
--- a/legacy/builder/builder_utils/utils.go
+++ b/legacy/builder/builder_utils/utils.go
@@ -154,11 +154,13 @@ func findAllFilesInFolder(sourcePath string, recurse bool) ([]string, error) {
 		}
 
 		for _, folder := range folders {
-			otherSources, err := findAllFilesInFolder(filepath.Join(sourcePath, folder.Name()), recurse)
-			if err != nil {
-				return nil, i18n.WrapError(err)
+			if !utils.IsSCCSOrHiddenFile(folder) {
+				otherSources, err := findAllFilesInFolder(filepath.Join(sourcePath, folder.Name()), recurse)
+				if err != nil {
+					return nil, i18n.WrapError(err)
+				}
+				sources = append(sources, otherSources...)
 			}
-			sources = append(sources, otherSources...)
 		}
 	}
 


d-a-v:
Sadly,
using today's arduino-nightly IDE, arduino-builder is still scanning my `.git` directory located in my hardware repository (under `hardware/` from arduino IDE root dir).
@gvarisco @matthijskooijman @facchinm (from arduino/arduino-builder#328)

earlephilhower:
Reopening as a WIP.  I'll need to find the core scanner as well and apply the same sort of logic that's being done on the sketch directory.

masci:
Can you rebase on `master` so the tests can run? Sorry for the trouble!

earlephilhower:
Sure thing.  I'll do the rebase by EOD my time.

I also noted in the Sketch file-filtering routine that `cvs` and `rcs` dirs are also skipped, so I think a refactor of both the core and sketch file filtering to use a single common "IgnoreDir()" function will be in order.  Don't want to duplicate code, incompletely, through the codebase, so this will need a little more work.

earlephilhower:
@masci, thanks for updating the tag.  I had to get a new Go build environment before I could rebase/fix the PR.  I can confirm @d-a-v's observation that git head of both repos does result in the 4.5GB RAM and 3:00++ min runtime, and that this PR as-is fixes the issue in my testing.

The new PR simply hooks the scans that arduino-builder *already* applies when looking at source folders to be scanned when checking that the Arduino `*.txt` (platform.txt, boards.txt, mainly) files are not updated since last build.

earlephilhower:
CI failure looks like a Codecov token issue, not related the the PR (I hope!).

d-a-v:
@masci Is it possible re-trigger CI for this PR ?

masci:
git seems to be in an inconsistent state, checkout is failing with this message: https://github.com/arduino/arduino-cli/pull/366/checks?check_run_id=209807044#step:3:346

This seems to happen to others as well, see https://github.com/actions/checkout/issues/23 but I'm afraid there's nothing I can do to fix this. @earlephilhower can you try adding another commit to your branch, or pick the commit into a brand new branch? (no idea if this will change anything, just speculation)

earlephilhower:
Thanks, @masci.  I've pushed in a no-op change and am seeing the GH actions working now.  We'll see in a few minutes if we're all good.

masci:
@earlephilhower we're good now.

The remaining failure is bad copypasta, this line https://github.com/arduino/arduino-cli/blob/master/.github/workflows/test.yaml#L80 should be added here as well https://github.com/arduino/arduino-cli/blob/master/.github/workflows/test.yaml#L66

earlephilhower:
Is the copypasta fix something you want in this PR?  Seems like it's better to have it in a separate push for sanity's sake, then I would just merge upstream/master and the commit logs would be comprehensive.

masci:
@earlephilhower you can rebase on `master`, I've merged the copypasta fix and that should give us a green build

earlephilhower:
Just a ping to say the WIP label can be removed at your convenience.  With your fix it's building and running cleanly now on my system and in CI.

d-a-v:
Many thanks for the fix, and for merging it quickly. Performance increase is just tremendous.

PR #387, opened 2019-09-02
[skip changelog] fix archive name for nightly builds

masci:


masci:
Self-reviewing to fix nightlies

PR #395, opened 2019-09-05
Show core name when listing boards

### LINKED ISSUE #191: Please specify instructions better
On step 4 of the instructions it says to use the command:

arduino-cli core search mkr1000

That's all fine and dandy if you use the mkr1000 but I do not know if that is the arduino I have.

When I google the phrase mkr1000 I get arduino mkr1000 wifi. So I assume that if you had that board you would come across a page that tells you the codeword for that is simply "mkr1000".

I have the arduino uno r3.

When I run the comand:

arduino-cli core search uno

I get: 

Searching for platforms matching 'uno'                                                                                                                                                          
                                                                                                                                                                                                
ID              Version Name                                                                                                                                                                    
arduino:avr     1.6.23  Arduino AVR Boards                                                                                                                                                      
arduino:megaavr 1.6.26  Arduino megaAVR Boards 

___________

Do I install both of these? Just one? Will either one work?

Please specify in the instructions how to take the brand name of my arduino and look up the model number or whatnot I am supposed to type into the search command OR specify what to do if multiple results come up.

masci:
Fixes #191

Now that #336 has been released, the vicious cycle of not knowing the board name if the corresponding core is not available, and not knowing which core to install without the board name, is finally broken.

To help users who are not familiar with the FQBN concept to better understand which core has to be installed, a new column in the output of `board list` was added to show the name of the core that  supports the detected board.

The README was updated accordingly. 

Note: the core name wasn't added to the JSON output because that would require a non trivial amount of code - provided that the JSON output is supposed to be read by a machine and provided that the name of the core can be easily deduced from the data the CLI already provides, my advice is to avoid making the code too complex and let client applications do a little extra work when needed.

### DIFF
--- a/cli/board/list.go
+++ b/cli/board/list.go
@@ -20,6 +20,7 @@ package board
 import (
 	"os"
 	"sort"
+	"strings"
 	"time"
 
 	"github.com/arduino/arduino-cli/cli/errorcodes"
@@ -85,7 +86,7 @@ func outputListResp(ports []*rpc.DetectedPort) {
 	})
 
 	t := table.New()
-	t.SetHeader("Port", "Type", "Board Name", "FQBN")
+	t.SetHeader("Port", "Type", "Board Name", "FQBN", "Core")
 	for _, port := range ports {
 		address := port.GetProtocol() + "://" + port.GetAddress()
 		if port.GetProtocol() == "serial" {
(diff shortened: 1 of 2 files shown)


cli/board/list.go
```diff
@@ -100,15 +101,17 @@ func outputListResp(ports []*rpc.DetectedPort) {
 			for _, b := range boards {
 				board := b.GetName()
 				fqbn := b.GetFQBN()
-				t.AddRow(address, protocol, board, fqbn)
+				core := fqbn[:strings.LastIndex(fqbn, ":")]
```
cmaglie: we have some infra for FQBN parsing here https://github.com/arduino/arduino-cli/blob/master/arduino/cores/fqbn.go#L36 that would be nice using here.

Issue #59, opened 2018-10-05
`arduino-cli sketch new XXX` command should create project at current folder, not `~/Arduino`


quchunguang:

macfisherman:
Would the logic of "current folder" extend to the other commands as well? It seems "compile" does, I haven't checked the other commands yet. "sketch new" creates a subdirectory for the project as well as a sketch file, would that behaviour be retained?

PR #396, opened 2019-09-05
Command `sketch new` do not join sketchbook path to sketch name if only sketchname is provided

rsora:
This PR closes https://github.com/arduino/arduino-cli/issues/59

This PR extends the following path logic:
1. if a sketch name only is passed as parameter, assume that the sketch directory is the current directory
2. If a relative path for the sketch is passed, join it with the current directory path
3. is an absolute path for the sketch is passed, simply use it

to the `sketch new` command in addition to `compile` and `upload` commands


cli/sketch/new.go
```diff
@@ -47,17 +48,21 @@ void loop() {
...
+	}
+	if err := os.MkdirAll(sketchDir, os.FileMode(0755)); err != nil {
 		feedback.Errorf("Could not create sketch directory: %v", err)
 		os.Exit(errorcodes.ErrGeneric)
 	}
-
-	sketchFile := sketchDir.Join(args[0] + ".ino")
-	if err := sketchFile.WriteFile(emptySketch); err != nil {
+	sketchName := filepath.Base(sketchDir)
+	sketchFile := filepath.Join(sketchDir, sketchName+".ino")
```
masci: Not sure if worth it, but we might strip the `.ino` suffix if already present in `args[0]` (e.g. maybe an user might do `sketch new foo.ino`)

test/test_compile.py
```diff
@@ -14,11 +14,40 @@
 # a commercial license, send an email to license@arduino.cc.
 import json
 import os
+
 import pytest
 
 from .common import running_on_ci
 
 
+def test_sketch_new(run_command):
```
masci: we should probably create a new module `test_sketch.py` for this

cli/sketch/new.go
rsora: Good point, added a `TrimSuffix` upstream.

test/test_compile.py
rsora: Module created. Additionally, I applyied a small modification to the `run_command` fixture in order to have a temp working dir as a sandbox for the sketch creation e2e tests

test/test_new.py
```diff
@@ -0,0 +1,51 @@
+# This file is part of arduino-cli.
```
masci: can we name this `test_sketch.py` for consistency?

test/test_sketch.py
```diff
@@ -0,0 +1,58 @@
...
+# a commercial license, send an email to license@arduino.cc.
+import os
+import platform
+
+import pytest
+
+from test.common import running_on_ci
+
+
+@pytest.mark.skipif(running_on_ci() and platform.system() == "Windows",
```
rsora: The e2e testing in the Github CI windows VM is temporarly disabled for the `sketch new` feature.
We are experiencing an odd behavior for windows VM in the `test` pipeline that we are not able to reproduce on a physical windows machine. The issue seems to be related to the managing of pytest temporary directories.
The test will remain disabled until a root cause is found and the fix introduced as a separate PR.

PR #415, opened 2019-09-18
Replace default dummy version

rsora:
The default dummy version `0.3.7-alpha.preview` overlaps with a real version and this creates confusion to developers and contributors.

This PR replaces the default version with an explicit dummy version `0.0.0-git`

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -63,12 +63,10 @@ vars:
     sh: echo `go list ./... | grep -v legacy | tr '\n' ' '`
 
   # build vars
-  VERSIONSTRING: "0.0.0-git"
   COMMIT:
     sh: echo ${TRAVIS_COMMIT:-`git log -n 1 --format=%h`}
   LDFLAGS: >
-      -ldflags '-X github.com/arduino/arduino-cli/version.versionString={{.VERSIONSTRING}}
-      -X github.com/arduino/arduino-cli/version.commit={{.COMMIT}}'
+      -ldflags '-X github.com/arduino/arduino-cli/version.commit={{.COMMIT}}'
 
   # test vars
   GOFLAGS: "-timeout 10m -v -coverpkg=./... -covermode=atomic"
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -63,7 +63,7 @@ vars:
     sh: echo `go list ./... | grep -v legacy | tr '\n' ' '`
 
   # build vars
-  VERSIONSTRING: "0.3.7-alpha.preview"
+  VERSIONSTRING: "0.0.0-git"
   COMMIT:
     sh: echo ${TRAVIS_COMMIT:-`git log -n 1 --format=%h`}
   LDFLAGS: >
--- a/version/version.go
+++ b/version/version.go
@@ -22,7 +22,7 @@ import (
 )
 
 var (
-	defaultVersionString = "0.3.7-alpha.preview"
+	defaultVersionString = "0.0.0-git"
 	versionString        = ""
 	commit               = ""
 )


Taskfile.yml
```diff
@@ -63,12 +63,10 @@ vars:
     sh: echo `go list ./... | grep -v legacy | tr '\n' ' '`
 
   # build vars
-  VERSIONSTRING: "0.3.7-alpha.preview"
   COMMIT:
     sh: echo ${TRAVIS_COMMIT:-`git log -n 1 --format=%h`}
   LDFLAGS: >
-      -ldflags '-X github.com/arduino/arduino-cli/version.versionString={{.VERSIONSTRING}}
-      -X github.com/arduino/arduino-cli/version.commit={{.COMMIT}}'
+      -ldflags '-X github.com/arduino/arduino-cli/version.commit={{.COMMIT}}'
```
cmaglie: Instead of `versionString` and `commit` isn't better to just have `version` set to the output of `git describe`?
From the git manpage it says:

> The command finds the most recent tag that is reachable from a commit. If the tag points to the commit, then only the tag is shown. Otherwise, it suffixes the tag name with the number of additional commits on top of the tagged object and the abbreviated object name of the most recent commit. The result is a "human-readable" object name which can also be used to identify the commit to other git commands.

rsora: @cmaglie I'm not sure that this approach is compatible with our current git release process: 

This is the result for `git describe` from our latest release branch:
```
12:37:43 in arduino-cli on branch 0.5.0 
➜ git describe
0.5.0
```
our current release process (documented internally) states that we:
1. create a release branch
2. Tag the HEAD of the release branch
3. Push the tag to origin to trigger the corresponding build on Github.

as reported in the documentation you provided:
> If the tag points to the commit, then only the tag is shown

So basically for the releases we lose the commit info.

Additionally, we present the version info in a "modular" fashion as you can see here
```
➜ arduino-cli version --format json
{
  "Application": "arduino-cli",
  "VersionString": "0.5.0",
  "Commit": "3be2287"
}
```
To maintain this modularity we will need to implement a parser to split the `git describe` output, so we end up writing additional code. 

Does this make sense to you? :)

cmaglie: > So basically for the releases we lose the commit info.

you don't need it, when the commit matches the tag, you can just use the tag (`0.5.0`) to reference the commit instead of the hash.

```
git checkout 0.5.0
git show 0.5.0
git rebase 0.5.0
git merge 0.5.0
```

all works fine.

> Additionally, we present the version info in a "modular" fashion as you can see here
>
> ➜ arduino-cli version --format json
> {
>   "Application": "arduino-cli",
>   "VersionString": "0.5.0",
>   "Commit": "3be2287"
> }

When version and commit referes to the same git revision the information above is correct, but also redundant, as I said the tag is sufficient.

On the other hand, the more interesting case is when you build over an untagged commit, for example building the current `master` (say with the nightly), with this PR you get:

> {
>   "Application": "arduino-cli",
>   "VersionString": "0.0.0-git",
>   "Commit": "b677403"
> }

that is correct but doesn't tell you much. Using git describe you will have:

> {
>   "Application": "arduino-cli",
>   "VersionString": "0.4.0-RC1-65-gb677403"
> }

that gives you the hint that this build is 65 commit ahead of 0.4.0-RC1: this is the kind of information I expect from a `version` command. I would remove also the `Application` field and just leave:

> {
>   "Version": "0.4.0-RC1-65-gb677403"
> }

BTW now I'm noticing that it didn't catch the most recent 0.5.0 because that tagged revision has not been merged back into master: to fully support this flow the release process should be changed as well.

So for now let's merge this PR as is, since it surely improves things, and keep `git describe` for a possible future improvement.

PR #420, opened 2019-09-22
Only copy elf file if it exists.

konkers:
Some boards do not produce elf file intermediate (i.e those that use
build.preferred_export_format=axf or the like).  This changes the
logic so that it's not an error for and elf file to be missing.


masci:
Tests pass locally, merging

PR #421, opened 2019-09-22
Use a local Walk function

### LINKED ISSUE #358: Null pointer if target .ino is a folder
### Current behavior
`./arduino-cli --debug compile -b anything:something:meh ~/Arduino/Blink` crashes with null pointer dereference if `~/Arduino/Blink` contains a **folder** called `Blink.ino`

### Expected behavior

Malformed ino files should be recognized and skipped.

### Environment

- CLI version (output of `arduino-cli version`): nightly-4-gcb3c79b 
- OS and platform: linux x86_64

### LINKED ISSUE #424: null pointer with symbolic links
### Current behavior
```
$ ls -al x y z w

lrwxrwxrwx   1 user user    1 Sep 22 17:25 y -> x      <- y is a symbolic link to x

w:
-rw-r--r--   1 user user    35 Sep 24 00:48 w.ino

x:
-rw-r--r--   1 user user    35 Sep 24 00:48 y.ino

z:
drwxr-xr-x   2 user user  4096 Sep 24 00:41 z.ino      <- z.ino is a directory

$ cat y/y.ino
void setup () { }
void loop () { }

$ cat w/w.ino
void setup () { }
void loop () { }

$ arduino-cli  compile --fqbn arduino:avr:nano y
(segfaults)

$ arduino-cli  compile --fqbn arduino:avr:nano z
(segfaults)

$ arduino-cli  compile --fqbn arduino:avr:nano w
(works)
```

Symbolic link as sketch directory fails with a segmentation fault.
Directory instead of a file fails with a segmentation faults (dup #358)

### Expected behavior

Symbolic link as sketch directory should work.
Directory instead of a file may fail with an error message.

### Environment

- CLI version (output of `arduino-cli version`): `arduino-cli Version: 0.0.0-git Commit: `
`git describe --tag: 0.3.7-alpha.preview-143-gafdf609`
- OS and platform: Linux avr

### Additional context

none

d-a-v:
Because arduino-cli segfaults when sketchDir is a symbolic link,
a simple walk function is introduced with a twofold effect:
- fix the segfault
- allow to use symlinks

ref: https://www.google.com/search?q=golang+Walk+does+not+follow+symbolic+links

*edit*:
* fixes #424 
* fixes #358

### DIFF
--- a/arduino/builder/sketch.go
+++ b/arduino/builder/sketch.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/arduino/arduino-cli/arduino/globals"
 	"github.com/arduino/arduino-cli/arduino/sketch"
+	"github.com/arduino/arduino-cli/cli/errorcodes"
 	"github.com/arduino/arduino-cli/cli/feedback"
 
 	"github.com/pkg/errors"
@@ -137,7 +138,6 @@ func SketchLoad(sketchPath, buildPath string) (*sketch.Sketch, error) {
 		if err != nil {
 			feedback.Errorf("Error during sketch processing: %v", err)
 			os.Exit(errorcodes.ErrGeneric)
-			return filepath.SkipDir
 		}
 
 		// ignore hidden files and skip hidden directories
--- a/arduino/builder/sketch.go
+++ b/arduino/builder/sketch.go
@@ -135,7 +135,8 @@ func SketchLoad(sketchPath, buildPath string) (*sketch.Sketch, error) {
 	err = simpleLocalWalk(sketchFolder, maxFileSystemDepth, func(path string, info os.FileInfo, err error) error {
 
 		if err != nil {
-			feedback.Printf("\nerror: %+v\n\n", err)
+			feedback.Errorf("Error during sketch processing: %v", err)
+			os.Exit(errorcodes.ErrGeneric)
 			return filepath.SkipDir
 		}
 
--- a/arduino/builder/sketch.go
+++ b/arduino/builder/sketch.go
@@ -60,6 +60,7 @@ func SketchSaveItemCpp(item *sketch.Item, destPath string) error {
 
 // SimpleLocalWalk locally replaces filepath.Walk and/but goes through symlinks
 func SimpleLocalWalk(root string, walkFn func(path string, info os.FileInfo, err error) error) error {
+
 	info, err := os.Stat(root)
 
 	if err != nil {
@@ -115,7 +116,6 @@ func SketchLoad(sketchPath, buildPath string) (*sketch.Sketch, error) {
 	// collect all the sketch files
 	var files []string
 	err = SimpleLocalWalk(sketchFolder, func(path string, info os.FileInfo, err error) error {
-
 		// ignore hidden files and skip hidden directories
 		if strings.HasPrefix(info.Name(), ".") {
 			if info.IsDir() {
--- a/arduino/builder/sketch.go
+++ b/arduino/builder/sketch.go
@@ -62,25 +62,25 @@ func SimpleLocalWalk(root string, walkFn func(path string, info os.FileInfo, err
 	info, err := os.Stat(root)
 
 	if err != nil {
-	    return walkFn(root, nil, err)
+		return walkFn(root, nil, err)
 	}
 
 	err = walkFn(root, info, err)
 	if err == filepath.SkipDir {
-	    return nil
+		return nil
 	}
 
 	if info.IsDir() {
-	    files, err := ioutil.ReadDir(root)
-        if err == nil {
-            for _, file := range files {
-                err = SimpleLocalWalk(root + string(os.PathSeparator) + file.Name(), walkFn)
-                if err == filepath.SkipDir {
-                    return nil
-                }
-            }
-        }
-    }
+		files, err := ioutil.ReadDir(root)
+		if err == nil {
+			for _, file := range files {
+				err = SimpleLocalWalk(root + string(os.PathSeparator) + file.Name(), walkFn)
+				if err == filepath.SkipDir {
+					return nil
+				}
+			}
+		}
+	}
 
 	return nil
 }
--- a/arduino/builder/sketch.go
+++ b/arduino/builder/sketch.go
@@ -74,7 +74,7 @@ func SimpleLocalWalk(root string, walkFn func(path string, info os.FileInfo, err
 		files, err := ioutil.ReadDir(root)
 		if err == nil {
 			for _, file := range files {
-				err = SimpleLocalWalk(root + string(os.PathSeparator) + file.Name(), walkFn)
+				err = SimpleLocalWalk(root+string(os.PathSeparator)+file.Name(), walkFn)
 				if err == filepath.SkipDir {
 					return nil
 				}
--- a/arduino/builder/sketch.go
+++ b/arduino/builder/sketch.go
@@ -23,9 +23,9 @@ import (
 	"regexp"
 	"strings"
 
-	"github.com/arduino/arduino-cli/cli/feedback"
 	"github.com/arduino/arduino-cli/arduino/globals"
 	"github.com/arduino/arduino-cli/arduino/sketch"
+	"github.com/arduino/arduino-cli/cli/feedback"
 
 	"github.com/pkg/errors"
 )
--- a/arduino/builder/sketch.go
+++ b/arduino/builder/sketch.go
@@ -22,6 +22,7 @@ import (
 	"path/filepath"
 	"regexp"
 	"strings"
+	"fmt"
 
 	"github.com/arduino/arduino-cli/arduino/globals"
 	"github.com/arduino/arduino-cli/arduino/sketch"
@@ -59,7 +60,7 @@ func SketchSaveItemCpp(item *sketch.Item, destPath string) error {
 }
 
 // SimpleLocalWalk locally replaces filepath.Walk and/but goes through symlinks
-func SimpleLocalWalk(root string, walkFn func(path string, info os.FileInfo, err error) error) error {
+func SimpleLocalWalkRecursive(root string, maxDepth int, walkFn func(path string, info os.FileInfo, err error) error) error {
 
 	info, err := os.Stat(root)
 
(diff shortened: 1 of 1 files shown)


d-a-v:
Help or informational input is requested about CI complaning :]

cmaglie:
Hi @d-a-v, thank for your submission.

We already have an implementation of the "walk-folllowing-symlinks" here:
https://github.com/arduino/arduino-cli/blob/master/legacy/builder/gohasissues/go_has_issues.go#L39

BTW, I was wondering why the golang `Walk` function doesn't follow symlinks and, looking at the golang bugtracker, I found that is because it may lead to infinite loops:
https://github.com/golang/go/issues/4759

This seems to be not golang-related but also other language that allows following symlinks suffer the same problem, for example in python we have:

https://docs.python.org/3/library/os.html#os.walk

that warns:

>Note Be aware that setting followlinks to True can lead to infinite recursion if a link points to a parent directory of itself. walk() does not keep track of the directories it visited already.

Since it seems that there isn't a saner/simple way to avoid loop we want to propose the following solution:

- Follow symlink only once on `sketchPath`
- Once sketchPath is resolved, use plain golang `Walk` from there

this should cover the use cases in #424 (and I think 99.99% of the overall use cases).
What do you think?

d-a-v:
Thanks,

I'm happy to close this PR when #424 is fixed by another mean.

About the walking issue with symbolic links, we can add a decreasing counter in the recursive walk function (like the factorial recursive function we learn at school), and return an error (like `Too many levels of symbolic links`) when it reaches 0.
That way there is no need for specific code anywhere, and we can have symbolic links for sketch dir, sketch files, `utility/` and files under `utility/`.

I've also read the golang issues in their repository and I just don't understand why or how they would rewrite the unix features their own way. A symbolic link loop is on user responsibility Golang sys lib's responsibility is only to not fail (heap here), not to remove any OS functionality IMHO.

```
$ ls -l t
lrwxrwxrwx 1 user group 1 Oct  9 16:42 t -> t
$ cd t
bash: cd: t: Too many levels of symbolic links
```

rsora:
Hi  @d-a-v , 
May we ask you to add a couple of test for this PR in order to outline your `walk` function behavior?

We want to use your PR as a starting point to introduce a more robust `walk` function, following symlinks only once on `sketchPath` (as @cmaglie mentioned), in order to fix rapidly the bugs and avoid symlink related issues.

As a second step, we will make the `walk` generic, adding a symlink loop check mechanism like https://gitlab.openebs.ci/openebs/maya/blob/5bfd0e581446d188292000626cc4dd1150e4d368/vendor/golang.org/x/tools/internal/gopathwalk/walk.go#L172

Thanks for your help!

d-a-v:
I'd be glad to.

First I need to know whether you agree with the idea that a path depth shall not be more than a given constant (like 40, similar to what's done in [linux kernel](https://lwn.net/Articles/650786/) `"Linux imposes a limit of at most 40 symlinks in any one path lookup"`) ?

Shall I make this proposal over [this file](https://github.com/arduino/arduino-cli/blob/1ceab8b1490b7a1e9a1439a7364891b7e2b692d1/legacy/builder/gohasissues/go_has_issues.go#L39) (instead of mine) ? Is it intended to stay where it is ?

masci:
My 2c:
* 40 is good because it's what the Linux kernel uses
* let's not add anything to the `legacy` package, I'd keep the function where it is now in `sketch.go`, we can move it later if other modules/packages need to walk the filesystem

rsora:
@d-a-v go ahead! Please keep the `walk` in `sketch.go` as @masci  said. Thanks!

d-a-v:
Testing the error was missing in the user walk function,
Can you please validate the print I added there ?

A reference to this issue is added in a comment about the discussion on `40`.
Usually one doesn't write raw constants.
Please direct me to the golang way/place for doing better.

d-a-v:
How to test this code (with linux):
In some sketch directory, create a symlink loop, detectable by OS:
```
$ mkdir utility
$ cd utility
$ ls -s x x
$ arduino-builder ...
[...]
error: stat /home/gauchard/dev/proj/arduino/y/utility/x: too many levels of symbolic links
[...]
```
(OS error printed)

Or not detectable by OS:
```
[...]
$ mkdir utility
$ cd utility
$ ln -s .. x
$ arduino-builder ...
error: Filesystem bottom is too deep (directory recursion or filesystem really deep): /path/to/mysketchdir/utility/x/utility/x/utility/x/utility/x/utility/x/utility/x/utility/x/utility/x/utility/x/utility/x/utility/x/utility/x/utility/x/utility/x/utility/x/utility/x/utility/x/utility/x/utility/x/utility/x
github.com/arduino/arduino-cli/arduino/builder.simpleLocalWalkRecursive
        /local/users/gauchard/arduino/gh-arduino-cli/arduino/builder/sketch.go:77
github.com/arduino/arduino-cli/arduino/builder.simpleLocalWalkRecursive
        /local/users/gauchard/arduino/gh-arduino-cli/arduino/builder/sketch.go:83
github.com/arduino/arduino-cli/arduino/builder.simpleLocalWalkRecursive
        /local/users/gauchard/arduino/gh-arduino-cli/arduino/builder/sketch.go:83
github.com/arduino/arduino-cli/arduino/builder.simpleLocalWalkRecursive
[...]
```
(depth detection error)

cmaglie:
Maybe we should differentiate two cases here:
1) a symlink that has a cycle to himself -> easy
2) a symlink that points to a directory that contains another symlink that points back to (any parent of...) the first one -> hard

Here the tricky case (2):

```
:~/tmp$ mkdir A
:~/tmp$ mkdir B
:~/tmp$ cd A/
:~/tmp/A$ ln -s ../B/
:~/tmp/A$ cd ..
:~/tmp$ cd B
:~/tmp/B$ ln -s ../A/
:~/tmp/B$ cd ..
:~/tmp$ find -L .
.
./A
./A/B
find: File system loop detected; './A/B/A' is part of the same file system loop as './A'.
./B
./B/A
find: File system loop detected; './B/A/B' is part of the same file system loop as './B'.
```

it seems that `ls` has no way to list files following symlinks recursively (it follows symlinks only if they are listed on the command line as arguments).
`find` seems to have a good algorithm to detect loops, but it will require more work.

d-a-v:
These two cases are handled by this PR.
Maybe I wasn't clear enough: the examples I show above are these two cases.
The tricky case is handled by the depth counter. The second quoted part of the examples is scrollable to the right and you can count the number of subdir levels: it's around 40.
Did I misunderstand something ?

About `find -L`'s algorithm, I believe it retains every inode and stops with a symlink-loop error when the same inode is found again. We could do that but given the rarity of the issue, a depth counter like implemented here is enough and easier to handle.

arduino/builder/sketch.go
```diff
@@ -58,6 +59,44 @@ func SketchSaveItemCpp(item *sketch.Item, destPath string) error {
...
+				}
+			}
+		}
+	}
+
+	return nil
+}
+
+// SimpleLocalWalk locally replaces filepath.Walk and/but goes through symlinks
+func SimpleLocalWalk(root string, walkFn func(path string, info os.FileInfo, err error) error) error {
```
rsora: The `simpleLocalWalkRecursive()` is currently used in this module only, There's no need to wrap it and export it. Can you please remove the wrapper and use directly `simpleLocalWalkRecursive()`? 
Regarding the `maxDepth` param, can you please create a private constant in this module (named for example maxFileSystemDepth) documenting it properly writing something like: `As currently implemented on Linux, the maximum number of symbolic links that will be followed while resolving a pathname is 40`? This way you can use something more meaningful as a parameter when calling the function.
Thanks!

rsora:
Hi @d-a-v! I left a comment on your code.
Can you please add a couple of tests to check the `maxDepth` control and the symlink loop?
You can see an example [here](https://github.com/arduino/arduino-cli/blob/f99361a94fcac76d46a2aa19dfc7458ce9774a12/arduino/builder/sketch_test.go#L53) on how to do it. The test code leverages sample sketches in the `testdata` folder, you can create a sketch that contains the symlinks you need for the testing

arduino/builder/sketch.go
```diff
@@ -79,14 +118,28 @@ func SketchLoad(sketchPath, buildPath string) (*sketch.Sketch, error) {
...
 		mainSketchFile = sketchPath
 	}
 
 	// collect all the sketch files
 	var files []string
-	err = filepath.Walk(sketchFolder, func(path string, info os.FileInfo, err error) error {
+	err = SimpleLocalWalk(sketchFolder, func(path string, info os.FileInfo, err error) error {
+
+		if err != nil {
+			fmt.Printf("\nerror: %+v\n\n", err)
```
rsora: you can use the `/cli/feedback` package to print error messages like 
```
feedback.Errorf("Error ....: %v", err)
```

d-a-v:
@rsora I think I addressed your review.

(CI is failing on `task check` but it passes when I try locally)

masci:
@d-a-v can you try manually running `go fmt arduino/builder/sketch.go`?

d-a-v:
@masci 
done, thanks

I don't understand CI messages though.

d-a-v:
What about a script to run before pushing changes that would run CI tests locally ?

masci:
@d-a-v our contributing guidelines already explain how to run tests locally before pushing code: https://github.com/arduino/arduino-cli/blob/master/CONTRIBUTING.md#running-the-tests

The style check, `task check` isn't mentioned though, I'll add it to the guide.

arduino/builder/sketch.go
```diff
@@ -79,14 +117,28 @@ func SketchLoad(sketchPath, buildPath string) (*sketch.Sketch, error) {
...
 		mainSketchFile = sketchPath
 	}
 
 	// collect all the sketch files
 	var files []string
-	err = filepath.Walk(sketchFolder, func(path string, info os.FileInfo, err error) error {
+	err = simpleLocalWalk(sketchFolder, maxFileSystemDepth, func(path string, info os.FileInfo, err error) error {
+
+		if err != nil {
+			feedback.Printf("\nerror: %+v\n\n", err)
```
rsora:
```suggestion
			feedback.Errorf("Error during sketch processing: %v", err)
			os.Exit(errorcodes.ErrGeneric)
```

Can you please add the error message as presented in the comment, and force a CLI exit with error?  In case of a symlink loop or other error during the sketch porcessing, the CLI must exit presenting an error.

rsora:
Hi @d-a-v , left a comment regarding the error management.
Thanks for your time!

d-a-v:
fixes addressed, CI passed, thanks for your feedback.

PR #427, opened 2019-09-25
Add detected_board fixture

zmoog:
This PR adds a new `detected_boards` fixture with the list of boards attached to the host machine.

I've add specific `_run_session_command` fixture to collect metadata using the same `arduino-cli` executable.

I've started to use the fixture in the `test_compile.py` but I'd love to hear what you think of this approach before go any further in the implementation.


zmoog:
D'oh! I'm sorry, I've fixed few errors on the `test_upload.py` but I did't run the whole suite of tests before push.. looking into integration tests failure right now!

masci:
Superseded by #564

PR #351, opened 2019-08-20
Added library dependency support

cmaglie:
This is preliminary work to implement library dependencies.

- [x] updated semver library to handle version constraints
- [x] json parsing
- [x] added tests for library index parser
- [x] added GRPC call `LibraryResolveDependencies` and implementation
- [x] CLI commands implementation for `install` and `upgrade`

PR #431, opened 2019-09-27
Stricter checks on libs/cores arguments

cmaglie:
Previously `globals.ParseReferenceArgs` was used to parse both core reference and library reference, but this doesn't work well because:

`MyLib@1.0.0` is a valid library reference but not a valid core reference (because it's missing the architecture part like `arduino:avr@1.0.0`.

This PR adds another function `global.ParseLibraryReferenceArgs` that also provides a better error messages for the reason why the parameters are rejected:
- `invalid empty library name` (ex: `@2.3.4`)
- `invalid empty library version` (ex: `MyLib@`)

and for cores:
- `invalid empry core argument`
- `invalid empty core reference` (ex: `@1.2.3`)
- `invalid empty core version` (ex: `arduino:avr@`)
- `invalid empty core name` (ex: `:avr@1.2.3`)
- `invalid empty core architecture` (ex `arduino:@1.2.3`)


cmaglie:
Already merged as part of #351

PR #433, opened 2019-10-01
Fix library-cache-miss regression

cmaglie:
Yet another post-builder regression.

### DIFF
--- a/legacy/builder/container_find_includes.go
+++ b/legacy/builder/container_find_includes.go
@@ -108,6 +108,7 @@ package builder
 
 import (
 	"encoding/json"
+	"fmt"
 	"os"
 	"os/exec"
 	"time"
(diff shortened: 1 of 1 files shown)


matthijskooijman:
Code looks good at first glance. I haven't tracked the builder and cli changes lately, but how is this a regression exactly? From the commit messages, you're changing the cache to use pointers for performance, and change == to equals to prevent pointer comparison?

cmaglie:
The regression was in the struct comparison with `==` operator, before it was working because the struct had only strings inside.

The comparison with `==` works only if there are no pointers involved:
- you must compare directly the structs (as values) and not pointers to the structs (otherwise the pointers will be different even if the "content" of the struct is the same)
- the struct fields must all be values (non-pointers) for the same reason as above

PS I'm going to do another round of refactoring to simplify this part and to move it out of the `legacy` package, I'd like your review once done :-)

PR #434, opened 2019-10-02
[skip changelog] gRPC "client_example" refresh

rsora:
This PR:
- updates the gRPC `client_example` to align it with the sketch new latest behavior (see #396 ) 
-  applies little cosmetics to `help` printout

### DIFF
--- a/client_example/main.go
+++ b/client_example/main.go
@@ -167,7 +167,9 @@ func initInstance(client rpc.ArduinoCoreClient) *rpc.Instance {
 	// the data folder.
 	initRespStream, err := client.Init(context.Background(), &rpc.InitReq{
 		Configuration: &rpc.Configuration{
-			DataDir: dataDir,
+			DataDir:       dataDir,
+			SketchbookDir: dataDir,
+			DownloadsDir:  dataDir,
 		},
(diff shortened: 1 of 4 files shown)


client_example/main.go
```diff
@@ -167,7 +167,9 @@ func initInstance(client rpc.ArduinoCoreClient) *rpc.Instance {
 	// the data folder.
 	initRespStream, err := client.Init(context.Background(), &rpc.InitReq{
 		Configuration: &rpc.Configuration{
-			DataDir: dataDir,
+			DataDir:       dataDir,
+			SketchbookDir: dataDir,
+			DownloadsDir:  dataDir,
```
cmaglie: SketchbookDir may be something like `dataDir/sketchbook`? just to differentiate it from dataDir (like in a real use case) but still keep it inside dataDir so it get wiped out after the test.

DownloadsDir should be automatically set to `dataDir/staging` I think it's not needed here?

rsora: :+1: see https://github.com/arduino/arduino-cli/pull/434/commits/899c13a0e075930a3079ba72daccd6884a29b274

PR #439, opened 2019-10-11
Avoid calling cloud API if vid or pid are invalid

masci:
When the USB vid or pid are obviously wrong, avoid calling the API, sparing users' bandwidth and reducing useless pressure on the backend.

I moved vid/pid validation inside `apiByVidPid` to keep the `List` logic tidy and clean. A specific test case was added to cover the case vid or pid are invalid.

### DIFF
--- a/commands/board/list.go
+++ b/commands/board/list.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"strconv"
 	"sync"
 
 	"github.com/arduino/arduino-cli/cli/globals"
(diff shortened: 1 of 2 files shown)


commands/board/list.go
```diff
@@ -35,9 +36,18 @@ var (
 	// ErrNotFound is returned when the API returns 404
 	ErrNotFound = errors.New("board not found")
 	m           sync.Mutex
+	vidPidURL   = "https://builder.arduino.cc/v3/boards/byVidPid"
 )
 
-func apiByVidPid(url string) ([]*rpc.BoardListItem, error) {
+func apiByVidPid(vid, pid string) ([]*rpc.BoardListItem, error) {
+	// ensure vid and pid are valid before hitting the API
+	_, vidErr := strconv.ParseInt(vid, 0, 64)
```
cmaglie: mmmhh does the API accepts VID/PID like `2341` instead of `0x2341`?

Also I think we should use `strconv.ParseUInt(vid, 16, 16)` because:
- it forces `16` as base, since VID/PID are always hex
- if the number is outside the VID/PID range `0x0000` `0xFFFF` (unsigned 16 bit) an error is returned

rsora: http://builder.arduino.cc/docs#!/boards95v3/boards_v3_byVidPid
The api accepts only the `0x` format

masci: Good catch for the int size, I'll bring it down to 16. 

It was my understanding the prefix was always there, if it is not the case, we've a bunch of logic to add to validate/normalize the string before hitting the API.

masci: Using regex since value from boards must be a string in the form `0x1234`.

PR #443, opened 2019-10-15
Fix "lib uninstall" when library name contains spaces

### LINKED ISSUE #362: Libraries installation/uninstallation name inconcistency
## Bug Report

The name to use when installing or uninstalling libraries is inconsistent

### Current behavior

Installing libraries work as intented. ie:

```
PS C:\Users\gaspa\Downloads> .\arduino-cli.exe lib install "LiquidCrystal I2C"
Downloading LiquidCrystal I2C@1.1.2...
LiquidCrystal I2C@1.1.2 already downloaded
Installing LiquidCrystal I2C@1.1.2...
Installed LiquidCrystal I2C@1.1.2
```
But trying to remove the library using the name used when installing, does not always remove the library.
```
PS C:\Users\gaspa\Downloads> .\arduino-cli.exe lib uninstall "LiquidCrystal I2C"
Library LiquidCrystal I2C is not installed
```
In this case I had to replace the space with a `_`
```
PS C:\Users\gaspa\Downloads> .\arduino-cli.exe lib uninstall "LiquidCrystal_I2C"
Uninstalling LiquidCrystal_I2C@1.1.2...
```

### Expected behavior

The libraries should be installed and uninstalled using the same name

#### CLI version
```
PS C:\Users\gaspa\Downloads> .\arduino-cli.exe version
arduino-cli.exe Version: 0.4.0 Commit: =98b7be9
```
#### OS/Platform

Windows 10 Pro build 1903

HowJMay:
Now,  the "_" is replaced with a blanket " ", to reslove the
inconsistency problem. However, use the name of the installing
library may be a better solution.

fixes #362 

BTW, when we uninstall libraries, there are no messages to notify the user libraries are successfully uninstalled.
However, when we install libraries message like `Installed LiquidCrystal I2C@1.1.2` is printed to notify  libraries are successfully installed.

### DIFF
--- a/arduino/libraries/librariesmanager/librariesmanager.go
+++ b/arduino/libraries/librariesmanager/librariesmanager.go
@@ -24,6 +24,7 @@ import (
 	"github.com/arduino/arduino-cli/arduino/cores"
 	"github.com/arduino/arduino-cli/arduino/libraries"
 	"github.com/arduino/arduino-cli/arduino/libraries/librariesindex"
+	"github.com/arduino/arduino-cli/arduino/utils"
 	paths "github.com/arduino/go-paths-helper"
 	"github.com/pmylund/sortutil"
 	"github.com/sirupsen/logrus"
@@ -209,7 +210,8 @@ func (sc *LibrariesManager) LoadLibrariesFromDir(librariesDir *LibrariesDir) err
 // name and version or, if the version is nil, the library installed
 // in the sketchbook.
 func (sc *LibrariesManager) FindByReference(libRef *librariesindex.Reference) *libraries.Library {
-	alternatives, have := sc.Libraries[libRef.Name]
+	saneName := utils.SanitizeName(libRef.Name)
+	alternatives, have := sc.Libraries[saneName]
 	if !have {
 		return nil
 	}
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -56,7 +56,7 @@ def run_command(pytestconfig, data_dir, downloads_dir, working_dir):
     Useful reference:
         http://docs.pyinvoke.org/en/1.2/api/runners.html#invoke.runners.Result
     """
-    cli_path = os.path.join(pytestconfig.rootdir, "..", "arduino-cli")
+    cli_path = os.path.join(str(pytestconfig.rootdir), "..", "arduino-cli")
     env = {
         "ARDUINO_DATA_DIR": data_dir,
         "ARDUINO_DOWNLOADS_DIR": downloads_dir,


masci:
Thanks for the PR!
Before we start reviewing, can you fix these two tests that are failing?
https://github.com/arduino/arduino-cli/blob/master/legacy/builder/test/libraries_loader_test.go#L92
https://github.com/arduino/arduino-cli/blob/master/legacy/builder/test/libraries_loader_test.go#L200

masci:
@HowJMay we're almost there, now integration tests are failing here: https://github.com/arduino/arduino-cli/blob/master/test/test_lib.py#L61
You can find the nstructions to run integration tests locally here: https://github.com/arduino/arduino-cli/tree/master/test let me know if you need any guidance, those tests require a Python3 interpreter to run.

HowJMay:
Thanks @masci 
It seems I may meet a problem, but I am still digging it.
I will ask for help if I can't solve alone!

HowJMay:
Hi @masci Sorry for answering late. My PC was broken in the last few weeks.
I met a problem that in 
https://github.com/arduino/arduino-cli/blob/292277f0db169ab65070c04827a0b91d6024447c/arduino/utils/filenames.go#L22
Name with space " " would be replace with underline "_", but this would cause a problem if I simply transform an underline back to space. Because the underline may be transformed from other symbol.

I am kind confused that how can I solve this problem. Maybe we need a naming rule for libraries ?

zmoog:
Hi @HowJMay, unfortunately, we currently don't have a specification for the library names.

The original library name is changed here

https://github.com/arduino/arduino-cli/blob/bc6c82320189e5dd80df0faffdffb2aec67f271c/arduino/libraries/librariesmanager/install.go#L40

using the `SanitizeName()` function to be more _filesystem friendly_ and is also used internally as an identifier for the library within the `LibrariesManager` data structures.

Why don't we limit the use of the _sanitized_ version of the library name inside the `librariesmanager` package only? 

We could change this:
https://github.com/arduino/arduino-cli/blob/bc6c82320189e5dd80df0faffdffb2aec67f271c/arduino/libraries/librariesmanager/librariesmanager.go#L212
into something like this:
```
    saneName := utils.SanitizeName(libRef.Name)
    alternatives, have := sc.Libraries[saneName]
```

All the sanitize/normalization stuff would remain inside the `librariesmanager` package making it easier to change it into something else later on.

HowJMay:
@zmoog Thank you !!
I didn't be aware there is a way like this!!

HowJMay:
Hi @masci I failed in CI for some unknown issue. Do you know what happen?

masci:
@HowJMay my apologies, the CI system is currently disabled due to administrative problems but it'll be back online soon. For the time being I'm afraid you have to rely on running the tests locally, thanks for your patience!

HowJMay:
Hi @masci How is the situation of CI?
Last time I ran the test of this PR, I passed it, even though I met some problems to run tests again.

masci:
@HowJMay CI is back, you can add a commit and push (or just force-push last commit) to retrigger

HowJMay:
done

HowJMay:
Thank you @masci

PR #447, opened 2019-10-18
Do not bail out if invalid PID is detected during board listing

cmaglie:
The USB PID, in theory, should be always filled with a value, but for some
reason the serial library is not detecting it correctly.

This should fix a fatal error on MacOSX, for some devices the returned
PID is empty:

  $ lsusb
  2019-10-18 09:47:32.186 system_profiler[34964:3584353] SPUSBDevice: IOCreatePlugInInterfaceForService failed 0xe00002be
  Bus 020 Device 000: ID 05ac:8600 Apple Inc. Apple T1 Controller
  Bus 000 Device 001: ID 1d6b:ISPT Linux Foundation USB 3.0 Bus
  Bus 000 Device 001: ID 1d6b:CIAR Linux Foundation USB 3.1 Bus
  Bus 001 Device 001: ID 1d6b:CIAR Linux Foundation USB 3.1 Bus

This fix should increase robustness until the underlying bug is fixed.

### DIFF
--- a/commands/board/list.go
+++ b/commands/board/list.go
@@ -35,6 +35,9 @@ import (
 var (
 	// ErrNotFound is returned when the API returns 404
 	ErrNotFound = errors.New("board not found")
+	// ErrInvalidParams is returned when invalid VID or PID are passed to the API
+	ErrInvalidParams = errors.New("Invalid VID/PID parameters")
+
 	m           sync.Mutex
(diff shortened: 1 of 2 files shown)


cmaglie:
> This fix should increase robustness until the underlying bug is fixed.

the underlying bug has been fixed in #448, there is no more point in merging this one.

PR #463, opened 2019-10-30
Board identification: always check vid.1/pid.1

### LINKED ISSUE #456: Custom board is detected as "Unknown"
### Current behavior

````bash
./arduino-cli board list
Port         Type              Board Name FQBN Core
/dev/ttyACM0 Serial Port (USB) Unknown
````

The connected board is not recognized.

### Expected behavior

````bash
./arduino-cli board list
Port         Type              Board Name FQBN Core
/dev/ttyACM0 Serial Port (USB) ftduino:avr:ftduino
````
I'd expect the device to get recognized. 
The arduino IDE correctly recognizes it.
The pid and vid are defined in the extension here:
https://github.com/harbaum/ftduino/blob/master/ftduino/boards.txt#L3-L7
Is there something else that needs to be done to recognize a custom board?
### Environment

- CLI version (output of `arduino-cli version`): arduino-cli Version: 0.6.0 Commit: 3a08b07
- OS and platform: Linux.

cmaglie:
Even if vid.0/pid.0 is missing.

Fix #456

@intrigus

### DIFF
--- a/arduino/cores/packagemanager/identify.go
+++ b/arduino/cores/packagemanager/identify.go
@@ -55,7 +55,7 @@ func (pm *PackageManager) IdentifyBoard(idProps *properties.Map) []*cores.Board 
 			if found {
 				foundBoards = append(foundBoards, board)
 			}
-			if !again {
+			if !again && id > 0 { // Always check id 0 and 1 (https://github.com/arduino/arduino-cli/issues/456)
 				break
 			}
 			id++
--- a/arduino/cores/packagemanager/identify.go
+++ b/arduino/cores/packagemanager/identify.go
@@ -30,7 +30,7 @@ func (pm *PackageManager) IdentifyBoard(idProps *properties.Map) []*cores.Board 
 		return []*cores.Board{}
 	}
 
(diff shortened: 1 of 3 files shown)


intrigus:
Thank you for your swift reply!

How does the arduino ide do the identification?
(Since the arduino ide correctly identifies the board)
So it might be beneficial to copy the ide's code.

This PR will probably fix the problem but it feels a little bit "hacky".
It is absolutely no problem for me to change the `boards.txt` file.
I.e. I technically don't "need" this PR, because I have full control over the `boards.txt` file.

cmaglie:
With this patch the cli will accept vid/pid starting from 0 or 1 but not from 2.
The Java IDE will accept any index, even starting from 2.

When I first looked at your issue it wasn't immediately clear the problem, I discovered it by comparing with another `boards.txt`, I think that the off-by-one may be a very common mistake and it's worth adding this corner case and relaxing a bit the specification.

Of course, if you can easily change your board.txt to start from 0, I would do it anyway :-)

cmaglie:
> That might make the code slightly more readable.

Right I've renamed the variables, I'll merge as soon as the CI checks the PR

PR #467, opened 2019-11-01
Fix handling of referenced cores

matthijskooijman:
A core is referenced by specifying `build.core=package:core` in board properties. However, the code that handled resolving these, had two problems:
 - It used the second part (`core`) rather than the first part (`package`) to look up the package to reference. The most commonly used core reference is `arduino:arduino` which works because both parts are identical. However, when referencing cores where the core name does not equal the package name (e.g. the Adafruit or STM32 cores), there is no way to actually reference such a core properly.
 - It also failed to handle a missing architecture, resulting in a segfault.

This PR fixes both problems, and additionally adds a testcase for `ResolveFQBN` that exposes these problems (and tests some other cases as well).


cmaglie:
Thanks!

PR #479, opened 2019-11-11
Auto-disable (deep) caching of core builds if needed

### LINKED ISSUE #464: [Windows] Sketch compilation and core cache cannot live in different partitions
If a certain buildpath is configured by adding "build.path=X:\directory_of_choice" to preferences.txt it may lead to a problem for the compiler:
In SDK versions before 1.8.10 it was possible to select "Aggressively cache compiled core" via a checkbox on File/Preferences. When enabled then core was stored in local temp directory which normally is placed on C:. In this case the configurable setting for build.path has to be on C: too. In any other case we get a compiler error like:
"can't make C:\Users\username\AppData\Local\Temp\arduino_cache_37519\core\core_bf266dfb68e28495ab12fdd4a1409781.a relative to X:\ArduinoBuild"
If caching core is disabled or build.path is on the same drive as default temp then everything works fine.
Since v1.8.10 the checkbox is not longer available. The feature can be enabled/disabled via compiler.cache_core=true/false in preferences,txt.

=> It seems that core is always build in default path while scetch is build in configurable path. Would it be possible to use configured build.path for core compilation too? Or is there any reason for using the present method?

cmaglie:
When build-path and cache-path are on different partitions it will not be possible (on windows) to reference them with a relative path. This commit allows this by automatically disabling the aggressive core caching.

Previously an error was emitted and the build interrupted.

Should fix #464

### DIFF
--- a/legacy/builder/phases/core_builder.go
+++ b/legacy/builder/phases/core_builder.go
@@ -53,7 +53,13 @@ func (s *CoreBuilder) Run(ctx *types.Context) error {
 	}
 
 	if coreBuildCachePath != nil {
-		if err := coreBuildCachePath.MkdirAll(); err != nil {
+		if _, err := coreBuildCachePath.RelTo(ctx.BuildPath); err != nil {
+			logger := ctx.GetLogger()
+			logger.Println(constants.LOG_LEVEL_INFO, "Couldn't deeply cache core build: {0}", err)
+			logger.Println(constants.LOG_LEVEL_INFO, "Running normal build of the core...")
+			coreBuildCachePath = nil
+			ctx.CoreBuildCachePath = nil
+		} else if err := coreBuildCachePath.MkdirAll(); err != nil {
 			return i18n.WrapError(err)
 		}
 	}


matthijskooijman:
Do we need to refer to them with a relative path? I guess absolute paths add command length, so that is not ideal, but maybe that is ok for just the core? Or are there other problems with absolute paths?

facchinm:
@matthijskooijman It was just about commandline length, so using the absolute path should solve the issue if `coreBuildCachePath.RelTo(ctx.BuildPath)` fails

cmaglie:
The problem is that platform.txt uses (example taken from samd core):

```
 recipe.c.combine.pattern="{compiler.path}{compiler.c.elf.cmd}"  "-L{build.path}" ...
   -o "{build.path}/{build.project_name}.elf" {object_files} ...
   -lm "{build.path}/{archive_file}" -Wl,--end-group
```

so we need `"{build.path}/{archive_file}"` with `archive_file` relative to `build.path`

BTW on other parts of platform.txt there is also:

```
# archive_file_path is needed for backwards compatibility with IDE 1.6.5 or older, IDE 1.6.6 or newer overrides this value
archive_file_path={build.path}/{archive_file}
recipe.ar.pattern="{compiler.path}{compiler.ar.cmd}" {compiler.ar.flags} {compiler.ar.extra_flags} "{archive_file_path}" "{object_file}"
```

so in theory we should only use `archive_file_path` instead of `{build.path}/{archive_file}`, but this is not true even for our cores.
So a full fix should check for `{archive_file}` usage in `combine` recipe and do the right action consequently.

Anyway this is a lot more work for a very edge case.

PR #487, opened 2019-11-18
Introduce Viper to manage the configuration file

masci:
This PR introduces https://github.com/spf13/viper to manage the configuration file in order to:

* consistently manage config bindings to env vars and command line switches
* get a more generic, key-value API to store/retrieve options
* easily get structured config options (see example below)

An example of configuration file would look like this:
```
board_manager:
  additional_urls:
  - https://raw.githubusercontent.com/ultimachine/ArduinoAddons/master/package_ultimachine_index.json

directories:
  data: /home/massi/.arduino15
  downloads: /home/massi/.arduino15/staging
  libraries: /home/massi/Arduino/libraries
  packages: /home/massi/.arduino15/packages
  sketchbook: /home/massi/Arduino

logging:
  file: ""
  format: text
  level: info
```

A new package, `configuration`, has been added to:
* keep default values definition in one single module
* provide an init mechanism, useful for testing and to allow user customization
* wrap common paths definitions

I've left inline comments to ease the review process.


arduino/cores/packagemanager/loader.go
```diff
@@ -24,25 +24,20 @@ import (
...
-	dirs, err := config.HardwareDirectories()
-	if err != nil {
-		return fmt.Errorf("getting hardware directory: %s", err)
-	}
+func (pm *PackageManager) LoadHardware() error {
+	dirs := configuration.HardwareDirectories()
 	if err := pm.LoadHardwareFromDirectories(dirs); err != nil {
 		return err
 	}
-	dirs, err = config.BundleToolsDirectories()
```
masci: the error was removed because these functions always returned `nil`

arduino/cores/packagemanager/loader.go
```diff
@@ -24,25 +24,20 @@ import (
...
-	if err != nil {
-		return fmt.Errorf("getting hardware directory: %s", err)
-	}
+
+	dirs = configuration.BundleToolsDirectories()
```
masci: the error was removed because these functions always returned `nil`

cli/cli.go
```diff
@@ -82,12 +80,16 @@ func createCliCommandTree(cmd *cobra.Command) {
 	cmd.AddCommand(version.NewCommand())
 
 	cmd.PersistentFlags().BoolVarP(&verbose, "verbose", "v", false, "Print the logs on the standard output.")
-	cmd.PersistentFlags().StringVar(&globals.LogLevel, "log-level", defaultLogLevel, "Messages with this level and above will be logged.")
-	cmd.PersistentFlags().StringVar(&logFile, "log-file", "", "Path to the file where logs will be written.")
-	cmd.PersistentFlags().StringVar(&logFormat, "log-format", "text", "The output format for the logs, can be [text|json].")
+	cmd.PersistentFlags().String("log-level", "", "Messages with this level and above will be logged.")
```
masci: We don't need a var anymore since the flag value is stored in the settings objects directly (this applies to following similar cases)

cli/cli.go
```diff
@@ -138,20 +161,9 @@ func preRun(cmd *cobra.Command, args []string) {
...
-			ForceColors: true,
-		})
-	} else {
-		// Discard logrus output if no writer was set
-		logrus.SetOutput(ioutil.Discard)
-	}
-
 	// configure logging filter
-	if lvl, found := toLogLevel(globals.LogLevel); !found {
-		fmt.Printf("Invalid option for --log-level: %s", globals.LogLevel)
```
masci: unrelated but we should use `feedback` instead of `fmt` across the codebase

cli/config/config.go
```diff
@@ -1,19 +1,17 @@
...
- * which covers the main part of arduino-cli.
- * The terms of this license can be found at:
- * https://www.gnu.org/licenses/gpl-3.0.en.html
- *
- * You can be released from the requirements of the above licenses by purchasing
- * a commercial license. Buying such a license is mandatory if you want to modify or
- * otherwise use the software for commercial activities involving the Arduino
- * software without disclosing the source code of your own applications. To purchase
- * a commercial license, send an email to license@arduino.cc.
- */
```
masci: unrelated but we should uniform license format when we find inconsistencies

configuration/configuration.go
```diff
@@ -0,0 +1,175 @@
...
+			// the test folder doesn't exist or is not accessible
+			return viper.GetBool("IDE.Bundled")
+		}
+
+		if test == "portable" {
+			viper.Set("IDE.Portable", true)
+		}
+	}
+
+	viper.Set("IDE.Directory", ideDir)
```
cmaglie: I think you forget to set `IDE.Bundled` to `true` here?

configuration/configuration.go
```diff
@@ -0,0 +1,175 @@
...
+	// We check an arbitrary number of folders that are part of the IDE
+	// install tree
+	tests := []string{
+		"tools-builder",
+		"examples/01.Basics/Blink",
+		"portable",
+	}
+
+	for _, test := range tests {
+		if _, err := os.Stat(filepath.Join(ideDir, test)); err != nil {
```
cmaglie: If the check for `portable` directory fails then the IDE is set as not-bundled, but in this case the cli may still be bundled with a IDE. I think that the check for "portable" should be done separately after this one.

configs/navigate.go
```diff
@@ -1,37 +0,0 @@
...
-
-package configs
-
-import (
-	paths "github.com/arduino/go-paths-helper"
-	"github.com/sirupsen/logrus"
-)
-
-// Navigate FIXMEDOC
-func (c *Configuration) Navigate(pwd *paths.Path) {
```
cmaglie: Does viper provides a `Navigate` feature already?

commands/compile/compile.go
```diff
@@ -88,20 +89,11 @@ func Compile(ctx context.Context, req *rpc.CompileReq, outStream, errStream io.W
...
-		builderCtx.BuiltInToolsDirs = toolsDir
-	} else {
-		return nil, fmt.Errorf("cannot get bundled tools directories: %s", err)
-	}
+	builderCtx.HardwareDirs = configuration.HardwareDirectories()
+	builderCtx.BuiltInToolsDirs = configuration.BundleToolsDirectories()
 
 	builderCtx.OtherLibrariesDirs = paths.NewPathList()
-	builderCtx.OtherLibrariesDirs.Add(config.LibrariesDir())
+	builderCtx.OtherLibrariesDirs.Add(paths.New(viper.GetString("directories.Libraries")))
```
cmaglie: Why not put `viper.GetString("directories.Libraries")` into an helper function? like as you did with `configuration.HardwareDirectories()` for example?

At this point I'd write an helper function for each key we put in viper, doing so will enforce a more strict compile-time checks.

configs/preferences_txt_serializer.go
```diff
@@ -1,147 +0,0 @@
...
-		config.IsPortable = true
-	}
-
-	config.ArduinoIDEDirectory = ideDir
-	res = true
-	return true
-}
-
-// LoadFromDesktopIDEPreferences loads the config from the Desktop IDE preferences.txt file
-func (config *Configuration) LoadFromDesktopIDEPreferences() error {
```
cmaglie: Are we dropping loading/overlaying options from the IDE `preferences.txt`? (just asking I may be OK with that :-))

configuration/configuration.go
masci: Fixed

configuration/configuration.go
masci: Fixed

configs/navigate.go
masci: You can call `viper.AddConfigPath()` several times passing the path where Viper should look for a config file

configs/preferences_txt_serializer.go
masci: At first I didn't fully get what this was for - I'd say let's get rid of it. 
If needed, I'd add a specific command like `arduino-cli config import /path/to/preferences.txt` so the operation is explicit.

commands/compile/compile.go
masci: One of the reason to switch to Viper was having a generic k/v store that can arbitrarily grow without the need to change the API. I understand the issue of losing the check on the configuration keys at compile time (if you write `viper.GetString("directories.Libs")` the program will happily compile despite not working as expected) but I think the tradeoff is worth it. We can catch errors on configuration keys at unit test time.

PR #488, opened 2019-11-19
Terminate daemon command when parent process exits; added "--daemonize" flag to keep old behaviour

cmaglie:
When running `arduino-cli` as a backend for the IDE, sometimes it happens that the IDE fails to terminate the process. This leaves a "zombie" `arduino-cli` process taking system resources and waiting for commands forever.

This patch adds the `--autoclose` flag to the cli daemon mode. When this option is flagged the cli monitors the 'os.Stdin` pipe and, if the parent process closes it, the cli terminates itself. Usually when the stdin is closed it means that the parent process is dead.

### DIFF
--- a/cli/daemon/daemon.go
+++ b/cli/daemon/daemon.go
@@ -50,27 +50,28 @@ func NewCommand() *cobra.Command {
 }
 
 func runDaemonCommand(cmd *cobra.Command, args []string) {
-	lis, err := net.Listen("tcp", port)
-	if err != nil {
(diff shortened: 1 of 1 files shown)


masci:
Should we do this by default, without an option? I couldn't think of a use case where you would want a 
zombie-like process...

cmaglie:
Maybe someone wants to keep it running as a real daemon... :-)

BTW I can see that this is a rare use case, let me invert the default: I'm changing it from `--autoclose` to `--daemonize`

PR #514, opened 2019-12-12
fix filter sketch source

umbynos:
Hopefully this will fix this:

> Go to Libraries
    Search for 'up' . Find "UP SQUARED GROVE IOT DEV KIT"
    Expand "Examples"
    Choose Blink sample.
    Choose board or port as " Intel X86_64 / IoT gateways
    Click Verify.
    Click "..." and "Export Cmake File"
    Wait for zip to download, unzip the file. Open the Blink.ino.cpp.
    The first two lines are random lines :
    '
    #line 1 "/tmp/115304058/Blink/Blink.ino"
    #line 1 "/tmp/115304058/Blink/Blink.ino"
    '
    There is also
    `
    #line 26 "/tmp/115304058/Blink/Blink.ino"
    void setup();
    #line 34 "/tmp/115304058/Blink/Blink.ino"
    void loop();
    #line 26 "/tmp/115304058/Blink/Blink.ino"
    '
>Surprisingly, the same issue does not exist for other samples in the same library.


legacy/builder/create_cmake_rule.go
```diff
@@ -125,22 +127,36 @@ func (s *ExportProjectCMake) Run(ctx *types.Context) error {
 		fmt.Println(err)
 	}
 
-	// Use old ctags method to generate export file
-	commands := []types.Command{
-		//&ContainerMergeCopySketchFiles{},
-		&ContainerAddPrototypes{},
```
cmaglie: `ContainerAddPrototypes` is still required, I think?

umbynos: I think it's not necessary because the only goal of `create_cmake_rule.go` is to generate the cmake and all the folder/files structure in order to make the compiling process possible. So the sketch I'm using it's already preprocessed and it's only copied from the sketch folder.

cmaglie: are you sure? have you tried to build something like:

```
void setup() {
   forwardFunction();
}
void loop() {
}
void forwardFunction() {
}
```

If it builds OK, then we can go ahead and merge.

umbynos: Yeah, it's not working with the code you provided:

cmaglie: I see the prototype of `forwardFunction`, what's the error exactly?

umbynos: There is no error, I probably misunderstood the expected behaviour

cmaglie: ok then, if the sketch compiles, the PR looks good to me.

PR #515, opened 2019-12-13
[skip changelog] Sliglthly optimize searchConfigTree

masci:
This PR optimize `searchConfigTree` to speed up performance on Windows, where it was particularly slow.

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -21,7 +21,6 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
-	"path"
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


cli/cli.go
```diff
@@ -124,15 +123,19 @@ func parseFormatString(arg string) (feedback.OutputFormat, bool) {
...
 	// go back up to root and search for the config file
 	for {
-		if _, err := os.Stat(path.Join(cwd, "arduino-cli.yaml")); os.IsNotExist(err) {
+		if _, err := os.Stat(filepath.Join(cwd, "arduino-cli.yaml")); err == nil {
+			// config file found
+			return cwd
+		} else if os.IsNotExist(err) {
 			// no config file found
-			next := path.Join(cwd, "..")
-			if filepath.Clean(next) == filepath.Clean(cwd) {
```
rsora: After a quick profiling session it seems that on Windows  the `filepath.Clean()` was responsible for the performance degradation. Removing this check fixed the CLI slow response (even with the `version` command) in case of missing `arduino-cli.yaml` configuration file on the machine.

Issue #503, opened 2019-11-30
Compile still unable to compile sketch with version 0.6.0


NormanDunbar:
## Bug Report
`Sketch new <name>` no longer creates the new sketch within the sketchbook location in the config. `Compile` does now find the sketches located in sketchbook.

Similar to #406, but the other way around!

### Current behavior
````
# Get config
arduino-cli config dump

proxy_type: auto
sketchbook_path: /home/norman/Arduino
arduino_data: /home/norman/.arduino15
board_manager: {}

# Where am I?
pwd
/home/norman

# Create sketch - outside of sketchbook location
arduino-cli sketch new testingAgain
Sketch created in: /home/norman/testingAgain

# Compile sketch
arduino-cli compile -b arduino:avr:mega testingAgain
Sketch uses 656 bytes (0%) of program storage space. Maximum is 253952 bytes.
Global variables use 9 bytes (0%) of dynamic memory, leaving 8183 bytes for local variables. Maximum is 8192 bytes.

# Move sketch to sketchbook
mv testingAgain/ /home/norman/Arduino/
cd /home/norman/Arduino
ls
... testingAgain ...

# Re-compile sketch
arduino-cli compile -b arduino:avr:mega testingAgain
Sketch uses 656 bytes (0%) of program storage space. Maximum is 253952 bytes.
Global variables use 9 bytes (0%) of dynamic memory, leaving 8183 bytes for local variables. Maximum is 8192 bytes.
````

So, #406 appears to have been fixed, but where the sketch is created appears to have regressed. Previous versions to 0.6.0, created new sketches within the sketchbook.

Sketches are no longer created within the sketchbook location according to the config. However, they can now be compiled if they are within the current folder or within the sketchbook

### Expected behavior
I expect the `sketch new` to continue to create sketches within the sketchbook area as defined in the config.

### Environment

- CLI version (output of `arduino-cli version`): arduino-cli Version: 0.6.0 Commit: 3a08b07
- OS and platform: Linux Mint 19.2 64 bit.

### Additional context
#406 was raised as the sketches were created in sketchbook, but compiled from the current directory. Now it's the other way around!

Thanks for all the hard work you do.

Cheers,
Norm.

masci:
Hi @NormanDunbar and thanks for the detailed feedback.

[...]

PR #516, opened 2019-12-13
Remove Sketchbook concept, introduce User data folder

masci:
## Rationale

After some QA iterations we decided to fade out the Sketchbook concept from the CLI. Please note the Sketchbook will continue existing on the IDEs, just it will be implemented at application level, not at the CLI level.

`arduino-cli sketch new` and other commands already stopped relying on the Sketchbook, assuming the current path will be used when no destination path is provided but the "Sketchbook" keyword continued to be present in some places in the CLI, most notably the configuration file, causing confusion among users.

### Implications

The Sketchbook isn't only used to collect sketch files, it also contains custom libraries and custom hardware under the paths `<sketchbook_path>/libraries` and `<sketchbook_path>/hardware` respectively - this functionality will be preserved by introducing a new concept at the CLI level: the **User** folder.

This PR removes the Sketchbook dir from the settings and introduces a new concept, the "User" folder. User folder will have the same layout as the existing "Data" folder and will contain "custom stuff" like libraries and hardware. Users can keep using it to store sketch files, just the CLI won't care nor assume anything about those.

### Backward compatibility

In order to keep backward compatibility, the environment var `ARDUINO_SKETCHBOOK_DIR` will keep existing and point to the User folder. The old `sketchbook_path` value will be removed from the settings instead and user will use this new configuration value in `arduino-cli.yaml`:

```
directories:
  data: /home/massi/.arduino15
  user: /home/massi/Arduino
```

### gRPC interface

At this moment the gRPC won't be touched and will keep showing references to the Sketchbook. That part will be refactored to improve the way gRPC clients can pass configuration values without having to rely on a config file.


cmaglie:
Everything looks good to me except:

> The old sketchbook_path value will be removed from the settings instead and user will use this new configuration value in arduino-cli.yaml

I'd keep this option so users that have an old config file will not suffer breaking changes.
I'm ok to remove it from `config dump` so new configurations will use the new values.

rsora:
@cmaglie keeping the `sketchbook_path` options unfortunately will continue to create issues as https://github.com/arduino/arduino-cli/issues/503 in which users are not able to figure out why the sketchbook concept does not work properly in the CLI.

I suggest to go ahead and merge, we will explain precisely in the changelog the effects of this change and how to migrate the `arduino-cli.yaml` to be used under the next release.

masci:
@pirropirro FYI we're merging this

PR #517, opened 2019-12-13
[skip changelog] bump packages to support Python 3.8

masci:
Also use pip-tools to express dependencies.


rsora:
Do we need also to update the pinned python version in the `/.github/wokflows/test` pipeline? https://github.com/arduino/arduino-cli/blob/cf39fbe69df7f9655121d0d8225b2c119c48e3d6/.github/workflows/test.yaml#L70

masci:
> Do we need also to update the pinned python version in the `/.github/wokflows/test` pipeline?
> 
We don't have to but let's do it :+1:

Issue #529, opened 2019-12-24
.arduino15/packages is not a directory


SConaway:
## Bug Report
On a fresh install, any command using the package manager fails. 
### Current behavior

```
# arduino-cli core update-index
Error initializing package manager: cannot initialize package manager: loading hardware packages: loading hardware from /root/.arduino15/packages: /root/.arduino15/packages is not a directory
# 
```
### Expected behavior
```
# arduino-cli core update-index
Downloading missing tool builtin:ctags@5.8-arduino11...
builtin:ctags@5.8-arduino11 downloaded                                                                                                                                                                             
Installing builtin:ctags@5.8-arduino11...
builtin:ctags@5.8-arduino11 installed
Downloading missing tool builtin:serial-discovery@1.0.0...
builtin:serial-discovery@1.0.0 downloaded                                                                                                                                                                          
Installing builtin:serial-discovery@1.0.0...
builtin:serial-discovery@1.0.0 installed
Updating index: library_index.json downloaded                                                                                                                                                                      
Updating index: package_index.json downloaded
# 
```

### Environment

- CLI version (output of `arduino-cli version`):
```
# arduino-cli version
arduino-cli Version: 0.7.0 Commit: 3809fc3
```
- OS and platform:
```
Any seems to cause this issue. I've noticed it on: 
1. Ubuntu 18.04 x64 in Docker
2. Ubuntu 19.10 x64
3. Fedora 31 x64
```

PR #533, opened 2019-12-28
[skip changelog] check if PackagesDir exists when reporting it in HardwareDirectories()

phd:
Fix error on `arduino-cli core update-index` with nonexistent ".arduino15" dir:
"Error initializing package manager: cannot initialize package manager: loading hardware packages: loading hardware from /.../.arduino15/packages: /.../.arduino15/packages is not a directory"

### DIFF
--- a/configuration/directories.go
+++ b/configuration/directories.go
@@ -33,7 +33,10 @@ func HardwareDirectories() paths.PathList {
 	}
 
 	if viper.IsSet("directories.Data") {
-		res.Add(PackagesDir())
+		packagesDir := PackagesDir()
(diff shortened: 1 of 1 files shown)


phd:
Fixes #529

masci:
Hi, thanks for the PR but there's already #530 fixing the same issue. That PR also contains some refactoring I'd like to ship as soon as I can.

phd:
Sure.
But I think this could still go in as it makes HardwareDirectories() function consistent.
With #530 ".arduino15" dir should now always exist, but who knows.

PR #536, opened 2020-01-02
Improve error message on build failure

### LINKED ISSUE #527: Incomplete error message when sketch uses too much memory
### Current behavior

When compiling a sketch that uses too much memory, the compile command output ends with something like this:

```
Sketch uses 22312 bytes (69%) of program storage space. Maximum is 32256 bytes.
Global variables use 3263 bytes (159%) of dynamic memory, leaving -1215 bytes for local variables. Maximum is 2048 bytes.
Not enough memory; see http://www.arduino.cc/en/Guide/Troubleshooting#size for tips on reducing your footprint.
Error during build: build failed: 
```

### Expected behavior

Last line should not end with colon (:)

### Environment

- CLI version (output of `arduino-cli version`): arduino-cli Version: 0.6.0 Commit: 3a08b07d458ef65fdefe88a7d71056227f626f5a
- OS and platform: macOS 10.15.2

masci:
Fixes #527 

At some point, deep into the `legacy` package, some function returns an error with an empty message, causing the command to fail in suspense with `"Error during build: build failed: `.

This is a workaround, the real fix would be going on a quest in search of the command(s) returning the empty error.

### DIFF
--- a/cli/compile/compile.go
+++ b/cli/compile/compile.go
@@ -19,6 +19,7 @@ package compile
 
 import (
 	"context"
+	"fmt"
 	"os"
 
 	"github.com/arduino/arduino-cli/cli/feedback"
@@ -88,7 +89,7 @@ func NewCommand() *cobra.Command {
 func run(cmd *cobra.Command, args []string) {
 	inst, err := instance.CreateInstance()
 	if err != nil {
-		feedback.Errorf("Error during build: %v", err)
+		feedback.Errorf("Error creating instance: %v", err)
 		os.Exit(errorcodes.ErrGeneric)
 	}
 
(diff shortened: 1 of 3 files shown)


federicobond:
Change LGTM. The errors with missing messages can be found here: https://github.com/arduino/arduino-cli/blob/7e547b0472bde76d1a676ec8c5323569be367224/legacy/builder/phases/sizer.go#L106-L114

If you want, I can submit a pull request adding proper messages along the lines of "text/data section exceeds available space in board" .

masci:
@federicobond thanks a ton for pointing me to the error, it was faster for me to just fix it in this PR but I've used your suggestion regarding the message. I've also removed the workaround `if err.Error() == ""`.

PR #523, opened 2019-12-18
Improve error message when port or fqbn flags are not set

federicobond:
This changeset marks the fqbn and port CLI flags as required in the compile and upload commands. The error message is thus changed from the generic:
```
Error during Upload: no upload port provided
```
To the more directly actionable:
```
Error: required flag(s) "fqbn", "port" not set
```
Which is followed by the CLI help text.

PR #545, opened 2020-01-08
Fix regression on compile and upload with attached sketches

masci:
In attempt to improve the error messages we merged #523 but this broke the `compile` and `attach` commands, complaining for missing FQBN even for sketches that have been previously attached.

Side note, since we merged #536, the error messages with this PR will be ok even with #523 partially reverted.


federicobond:
Hi @masci, I had opened another PR which produces a nice CLI message and handles attached sketches correctly. Let me know if you are still interested in it.

masci:
@federicobond go for it, I needed a quick fix to address the regression but I'll be happy to merge a more comprehensive solution!

PR #548, opened 2020-01-09
[skip changelog] [skip ci] Clarify additional-urls option usage

rsora:
This PR propose the usage of the config file option as best mode for 3rd party core usage.

The `--additional-urls` flag is now presented as an alternative, adding a complete usage explanation.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -293,21 +293,10 @@ Great! Now we are ready to compile and upload the sketch.
 Adding 3rd party cores
 ^^^^^^^^^^^^^^^^^^^^^^
 
-If your board requires 3rd party core packages to work, you can pass a link to
-the the additional package index file with the ``--additional-urls`` option to
-any command that require a platform core to work:
+If your board requires 3rd party core packages to work, you can list the URLs
+to additional package indexes in the Arduino CLI configuration file.
 
-.. code:: console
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -330,6 +319,19 @@ additional URL from the configuration file:
    ID              Version Name
    esp8266:esp8266 2.5.2   esp8266
 
+Alternatively you can pass a link to the the additional package index file with the ``--additional-urls`` option to
+every command that require a platform core to work:
```
masci: s/require/requires

README.rst
```diff
@@ -330,6 +319,19 @@ additional URL from the configuration file:
    ID              Version Name
    esp8266:esp8266 2.5.2   esp8266
 
+Alternatively you can pass a link to the the additional package index file with the ``--additional-urls`` option to
```
masci: we could stress the fact that you always have to pass the option if you decide to not use the config file, something on these lines:

> Alternatively, you can pass a link to the the additional package index file with the `--additional-urls` option, but this has to be done every time and for every command that operates on a platform core, for example:

rsora: :+1: I have rephrased it a bit, please check if it makes sense.

PR #558, opened 2020-01-16
Remove ignored Configuration GRPC API

cmaglie:
At the moment we have the `Configuration` field in `InitReq` that is marked as deprecated but it's actually not supported anymore.

IMHO if we want to deprecate it we should not silently ignore it otherwise, from the client perspective, it seems that the command has been accepted, leading to a hard to diagnose errors.

Otherwise, if we don't want to support it anymore it's better to merge this PR and remove it altogether so we have a clear indicator that the old API is no longer available.


rsora:
@cmaglie rebased on master to include latest changes and re-generated proto files!

go.mod
```diff
@@ -47,6 +47,7 @@ require (
 	go.bug.st/serial.v1 v0.0.0-20180827123349-5f7892a7bb45 // indirect
 	golang.org/x/net v0.0.0-20190311183353-d8887717615a
 	golang.org/x/text v0.3.0
+	google.golang.org/appengine v1.4.0 // indirect
```
cmaglie: Why this appengine keeps coming back? Are we using it?

cmaglie:
I can't approve my own PR, anyway the rebase seems ok, we can merge IMHO.

masci:
Ok to remove because we're 0.x, but the reason why I deprecated it instead of killing with fire was to avoid breaking clients that were not using the Config message (it was the case of the PRO ide).

PR #560, opened 2020-01-18
Add cache clean command

HowJMay:
Add command `cache clean` to clean the cache files
depending on the loaction of the caching files
under different OS.

Relates #285

### DIFF
--- a/cli/cache/clean.go
+++ b/cli/cache/clean.go
@@ -1,0 +1,75 @@
+// This file is part of arduino-cli.
+//
+// Copyright 2020 ARDUINO SA (http://www.arduino.cc/)
+//
+// This software is released under the GNU General Public License version 3,
+// which covers the main part of arduino-cli.
+// The terms of this license can be found at:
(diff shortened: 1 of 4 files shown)


HowJMay:
I should appending test case in this PR right?

cli/cache/clean.go
```diff
@@ -0,0 +1,75 @@
...
+		Args:    cobra.NoArgs,
+		Run:     runCleanCommand,
+	}
+	return cleanCommand
+}
+
+func runCleanCommand(cmd *cobra.Command, args []string) {
+	logrus.Info("Executing `arduino cache clean`")
+
+	cachePath := getDefaultArduinoDataDir() + "/staging"
```
masci: The configuration system already computes the path to `staging` at startup (see https://github.com/arduino/arduino-cli/blob/master/cli/cli.go#L163), all you have to do here is:
```
cachePath := viper.GetString("directories.Downloads")
```
the rest looks good! :+1:

HowJMay:
Fixed it

PR #573, opened 2020-01-31
Export Prometheus telemetry in daemon mode

rsora:
This PR implements the basis for a telemetry feature for the CLI in `daemon` mode.

**Rationale:**
Enhance the observability of the CLI internals at runtime, as we are proceeding internally to integrate the CLI as a gRPC server in all our services.

**Module `telemetry`**
Leverages https://github.com/segmentio/stats library to configure and expose a Prometheus endpoint when activated (in `cli/daemon/daemon.go`). It also provides helpers for transformation of `commands` input data in `stats.Tags` objects, using the installation id and secret generated by the `inventory` module.

The telemetry activation and endpoint config are available via CLI configuration.

**Module**  ~`repertory` (feel free to ask for a name change :smile:)~ **`inventory`**
This module leverages an additional Viper instance, in order to create a configuration repository that contains values that may create confusion in the user-facing `arduino-cli.yaml` file, polluting it.

This is the case for the `installation` values stored in the `inventory.yaml` file, that are used to generate server specific UUIDs that help the monitoring tools to perform statistics aggregation, when monitoring multiple CLI gRPC server instances.

The inventory file is created at the CLI first launch and it is stored in the `directories.Data` folder.

**Instrumentation**
Code instrumentation is done via standard `stats` apis for each command (only for the `compile` and `board list` in this PR):
1. Create the set of tags for the command
``` go
tags := map[string]string{
		"fqbn":            req.Fqbn,
		"sketchPath":      telemetry.SanitizeSketchPath(req.SketchPath),
		"showProperties":  strconv.FormatBool(req.ShowProperties),
		"preprocess":      strconv.FormatBool(req.Preprocess),
		"buildProperties": strings.Join(req.BuildProperties, ","),
...
                "success":          "true",
```

2. Modify the `success` tag according to each case
```go 
tags["success"] = "true"
```
3. Push the data to the `stats` engine 
```go
stats.Incr("compile", stats.M(tags)...)
```
**Tests**
- `test_inventory_creation`: python integ test
- `test_inventory_prometheus_endpoint`: python integ test

**Notes**
- Upgraded dependencies (Viper, pyinvoke as https://github.com/pyinvoke/invoke/issues/660 was closed)
- Updated README.rst adding details about telemetry endpoint and config


repertory/repertory.go
```diff
@@ -0,0 +1,81 @@
...
+)
+
+// Store is the Read Only config storage
+var Store = viper.New()
+
+var (
+	// Type is the repertory file type
+	Type = "yaml"
+	// Name is the repertory file Name with Type as extension
+	Name = "repertory" + "." + Type
```
cmaglie: I tried the deamon and the `repertory.yaml` file is overwritten at every run (so we lose the previously generated installation id each time the deamon starts). After adding a print of the error I found that viper is unable to find the config file even if the file is there. Since we do recreate the config in case of `viper.ConfigFileNotFoundError` the file is overwritten.

Looking at the viper docs I see that `Name` should be without extension https://github.com/spf13/viper#reading-config-files and in fact by removin the extension from `Name` it works. (I glanced into viper internals and the config file handling seems a bit overengineered, but... anyway...)

commands/compile/compile.go
```diff
@@ -33,26 +34,48 @@ import (
...
+		"fqbn":            req.Fqbn,
+		"sketchPath":      telemetry.SanitizeSketchPath(req.SketchPath),
+		"showProperties":  strconv.FormatBool(req.ShowProperties),
+		"preprocess":      strconv.FormatBool(req.Preprocess),
+		"buildProperties": strings.Join(req.BuildProperties, ","),
+		"warnings":        req.Warnings,
+		"verbose":         strconv.FormatBool(req.Verbose),
+		"quiet":           strconv.FormatBool(req.Quiet),
+		"vidPid":          req.VidPid,
+		"exportFile":      req.ExportFile,
```
cmaglie: This one should be sanitized as well?

commands/compile/compile.go
```diff
@@ -33,26 +34,48 @@ import (
...
+		"jobs":            strconv.FormatInt(int64(req.Jobs), 10),
+		"libraries":       strings.Join(req.Libraries, ","),
+		"success":         "false",
+	}
+
 	pm := commands.GetPackageManager(req.GetInstance().GetId())
 	if pm == nil {
+		stats.Incr("compile", stats.M(tags)...)
```
cmaglie: Can we `defer` this at the begginig? it's more future-proof IMHO since it's very easy to forget to call `stats.Incr("compile", stats.M(tags)...)` before returning.

cli/daemon/daemon.go
```diff
@@ -88,6 +103,9 @@ func runDaemonCommand(cmd *cobra.Command, args []string) {
 		go func() {
 			// Stdin is closed when the controlling parent process ends
 			_, _ = io.Copy(ioutil.Discard, os.Stdin)
+			if viper.GetBool("telemetry.enabled") {
```
masci: do we need this `if`? If telemetry wasn't activated Flush should be a no-op, correct?

commands/compile/compile.go
```diff
@@ -61,10 +84,12 @@ func Compile(ctx context.Context, req *rpc.CompileReq, outStream, errStream io.W
 		fqbnIn = sketch.Metadata.CPU.Fqbn
 	}
 	if fqbnIn == "" {
+		stats.Incr("compile", stats.M(tags)...)
 		return nil, fmt.Errorf("no FQBN provided")
 	}
 	fqbn, err := cores.ParseFQBN(fqbnIn)
 	if err != nil {
+		stats.Incr("compile", stats.M(tags)...)
```
masci: I find a bit hard to read that `tags` here contains `success: false`, can we tag success status explicitly?

test/conftest.py
```diff
@@ -78,6 +79,35 @@ def _run(cmd_string):
...
+    os.makedirs(os.path.join(data_dir, "packages"))
+    run_context = Context()
+    run_context.cd(working_dir)
+    # Local Class is the implementation of a Runner abstract class
+    runner = Local(run_context)
+    runner.run(
+        cli_full_line, echo=False, hide=True, warn=True, env=env, asynchronous=True
+    )
+
+    return runner
```
masci: If you `yield` instead of `return`, you can tear down the process in this function instead of having the function tests doing so. The pattern would be:
```
def my_fixture():
  ...
  runner.run(...)
  yield runner  # we block here until the test function using this fixture has returned
  kill(runner)
```

repertory/repertory.go
```diff
@@ -0,0 +1,81 @@
...
+// The terms of this license can be found at:
+// https://www.gnu.org/licenses/gpl-3.0.en.html
+//
+// You can be released from the requirements of the above licenses by purchasing
+// a commercial license. Buying such a license is mandatory if you want to
+// modify or otherwise use the software for commercial activities involving the
+// Arduino software without disclosing the source code of your own applications.
+// To purchase a commercial license, send an email to license@arduino.cc.
+
+package repertory
```
masci: I think `inventory` would sound more familiar for something that'll store metadata about the running environment.

telemetry/telemetry.go
```diff
@@ -0,0 +1,66 @@
...
+	go func() {
+		http.Handle(serverPattern, ph)
+		logrus.Error(http.ListenAndServe(serverAddr, nil))
+	}()
+
+}
+
+// SanitizeSketchPath uses config generated UUID (installation.secret) as an HMAC secret to sanitize and anonymize
+// the sketch name maintaining it distinguishable from a different sketch from the same Installation
+func SanitizeSketchPath(sketchPath string) string {
```
masci: I think this should be more generic, like just `Sanitize(s string)`: if `s` is a path, it'll be caller's responsibility to get the file name with `filepath.Base` and pass it to this function.
If sanitizing a sketch name is recurring and calling `filepath.Base` gets annoying, we might add a `SanitizeFileName(path string)` helper that gets the file name out of a path and calls `Sanitize`.

configuration/defaults.go
```diff
@@ -35,4 +35,10 @@ func setDefaults(dataDir, userDir string) {
 
 	// daemon settings
 	viper.SetDefault("daemon.port", "50051")
+
+	//telemetry settings
+	viper.SetDefault("telemetry.enabled", true)
+	viper.SetDefault("telemetry.addr", ":9090")
+	viper.SetDefault("telemetry.pattern", "/metrics")
```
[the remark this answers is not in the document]
rsora: Do you mean we should remove these 2 settings?
```
viper.SetDefault("telemetry.addr", ":9090")
viper.SetDefault("telemetry.pattern", "/metrics")
```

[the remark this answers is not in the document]
masci: only `telemetry.pattern`, port can be useful

repertory/repertory.go
rsora: This bug is caused by an issue contained in viper 1.3.2, upgrading to 1.6.2 (the version that I had during my tests) solved your issue. 
Maybe something went wrong while rebasing the PR

telemetry/telemetry.go
rsora: You're right, this also makes me think that we should pass to the `Sanitize` function the complete path, as users may use the same name for sketches organized in different sub-folders (or projects) that can have different content based on the project.

test/conftest.py
```diff
@@ -78,6 +81,42 @@ def _run(cmd_string):
...
+    run_context = Context()
+    run_context.cd(working_dir)
+    # Local Class is the implementation of a Runner abstract class
+    runner = Local(run_context)
+    runner.run(
+        cli_full_line, echo=False, hide=True, warn=True, env=env, asynchronous=True
+    )
+
+    # we block here until the test function using this fixture has returned
+    yield runner
```
masci: :cake:

commands/board/list.go
```diff
@@ -105,15 +105,19 @@ func List(instanceID int32) ([]*rpc.DetectedPort, error) {
...
 	defer m.Unlock()
 
+	tags := map[string]string{}
+	// Use defer func() to evaluate tags map when function returns
+	defer func() { stats.Incr("board.list", stats.M(tags)...) }()
+
 	pm := commands.GetPackageManager(instanceID)
 	if pm == nil {
-		stats.Incr("board.list", stats.T("success", "false"))
+		tags["success"] = "false"
```
cmaglie: I think we can push this further with named return params.

rsora: Can you give me an example of that?

PR #580, opened 2020-02-12
Force notarization process to take place configuring Gon zip output

rsora:
In order to force the notarization process to take place, we need to ask `gon` to produce a `.zip` file as output.

As explained here https://github.com/arduino/arduino-cli/pull/578 / https://github.com/arduino/arduino-cli/pull/578#discussion_r377794210 the `gon.config.hcl` "zip" section is not mandatory di per se', but not  having an "output" section at all will not start the notarization submission process for the binary.

For our release process the generated `.zip` file has no use, as we repackage the signed binary that `gon` replaces in-place. The `.zip` is functional for the notary service submission only

Thanks to @zmoog for patiently explaining the whole process to me countless times :bowing_man:

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -61,7 +61,7 @@ jobs:
         # 1. Download keychain from GH secrets and decode it from base64
         # 2. Add the keychain to the system keychains and unlock it
         # 3. Call Gon to start notarization process (using AC_USERNAME and AC_PASSWORD)
-        # 4. Repackage the signed binary replaced in place by Gon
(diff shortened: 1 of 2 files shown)


zmoog:
Yeah, it looks like the `altool`, that gon is orchestrating behind the scenes, require the binary to be packaged in a zip file, as described in this [Customizing the Notarization Workflow](https://developer.apple.com/documentation/xcode/notarizing_macos_software_before_distribution/customizing_the_notarization_workflow#3087734) article.

So, LGTM.

PR #582, opened 2020-02-14
Move board rpc commands in their own package

matteosuppo:
Keeping with the process of isolating features, the board commands
have been moved in their own package.

Nothing should change from the user perspective

Signed-off-by: Matteo Suppo <matteo.suppo@gmail.com>


cmaglie:
A related [comment](https://github.com/arduino/arduino-cli/issues/553#issuecomment-591896109) from @kittaakos: 

> > Hi @andmeyee you're right, at this moment settings are global,
> 
> @masci, before the introduction of the new Settings API, two services were available: the ArduinoCore and the Monitor. I thought the clear separation of these two services was based on the fact that the Monitor service is entirely independent of the ArduinoCore since one could tune the datadir, sketchbookdir, downloadsdir, and the boardmanageradditionalurlsList in the InitRequest's Configuration. So it made sense to have one single gRPC monitor client and several Arduino core clients, as they were initialized differently.
> 
> My questions would be the followings:
> 
> Knowing that the settings are global, would you still recommend downstream consumers of the CLI to maintain multiple ArduinoCore gRPC clients, or is it safe to have one and share it?
> If it is safe to share one ArduinoCore gRPC client, what is the benefit of having three different types of clients (Settings, ArduinoCore, and Monitor)?  **As a consumer of the CLI, I have to maintain the lifecycle of three different gRPC clients, although they're connected to the same server.**
> Any kind of feedback is welcome on this topic; I am trying to figure out whether we can simplify the architecture of the Pro IDE. Thank you!

PR #585, opened 2020-02-18
Use enumeration for LibraryLocation and LibraryLayout fields in grpc API

cmaglie:
The enumeration is also reused inside the `arduino` package to reduce code duplication.

This kind of change IMHO improves:
* code duplication, because the enumeration is defined in a single place
* the old `string` field in grpc definition, because now is an enumeraion and not a free string: t's more clear which value it can assume
* language bindings. In particular in the Java bindings the generated API is better, for example see https://github.com/arduino/Arduino/pull/9643/commits/9f58bae901bf6133b437d00ce581ef7ff3a9ce5c

As a side effect, after this change, the rpc definitions starts to spill inside the `arduino` package.
This means that the `rpc` definitions inside the .proto files became a sort of tables-of-truth, at least for this part of the API.


cmaglie:
Superseded by #613

Issue #567, opened 2020-01-24
Add debug command


spoenemann:
## Feature Request

We are implementing a debugger for the Arduino Pro IDE. The debugger needs specific command line tools for the selected board. Currently we have hard-coded the paths to these tools for the Arduino Zero board, but of course we'd like to generalize this (https://github.com/arduino/arduino-pro-ide/issues/185).

Could the CLI provide paths to the debugging tools when we query board details?

This is the list of required tools, along with the path we currently use for Arduino Zero:

 * gdb
   `$Arduino15/packages/arduino/tools/arm-none-eabi-gcc/7-2017q4/bin/arm-none-eabi-gdb`
 * objdump
   `$Arduino15/packages/arduino/tools/arm-none-eabi-gcc/7-2017q4/bin/arm-none-eabi-objdump`
 * openocd
   `$Arduino15/packages/arduino/tools/openocd/0.10.0-arduino7/bin/openocd`
 * openocd scripts
   `$Arduino15/packages/arduino/tools/openocd/0.10.0-arduino7/share/openocd/scripts`
 * openocd configuration
   `$Arduino15/packages/arduino/hardware/samd/1.8.4/variants/arduino_zero/openocd_scripts/arduino_zero.cfg`

spoenemann:
An even better alternative to printing the debugging tool locations as described above would be to add a `debug` command that takes care of spawning `gdb` and `openocd` with the correct arguments.

The current proof-of-concept of the debugger does the following:

 - `openocd` is spawned with the command
```
~/Arduino15/packages/arduino/tools/openocd/0.10.0-arduino7/bin/openocd -s ~/Arduino15/packages/arduino/tools/openocd/0.10.0-arduino7/share/openocd/scripts --file ~/Arduino15/packages/arduino/hardware/samd/1.8.4/variants/arduino_zero/openocd_scripts/arduino_zero.cfg --file /tmp/.../gdb.cfg
```
   where gdb.cfg is a temporary file with the content
```
gdb_port 50000
telnet_port 4444
echo "GDB server started"
```
 - gdb is spawned with the command
```
~/Arduino15/packages/arduino/tools/arm-none-eabi-gcc/7-2017q4/bin/arm-none-eabi-gdb --interpreter=mi2 -q ~/Arduino/Blink/Blink.arduino.samd.arduino_zero_edbg.elf
```
- Afterwards we send the following commands to the gdb process via std I/O
```
target-async on
-target-select extended-remote localhost:50000
[...]

PR #590, opened 2020-02-20
Implement debug feature

rsora:
This PR Implements the `debug` functionality, in gRPC and "Command Line Interface" flavor (see  https://github.com/arduino/arduino-cli/issues/567).

**IMPORTANT NOTES:** 
1. The `debug` functionality expects that a core is configured properly to expose a `debug` recipe. ATM the only Arduino core that will support the `debug` feature is the `arduino:samd` core (see https://github.com/arduino/ArduinoCore-samd). Other Arduino cores will follow.

1. In most cases you will need a debugger hardware tool to use the `debug` feature (i.e. to debug a  `mkr1000` you can use an Atmel ICE tool). Another option is to use boards like the Arduino Zero that has an additional debugging port.

### gdb Command Generation
This PR implements a `commands/debug/debug.go` module that wraps an interactive `gdb` session, this makes the Arduino CLI to act as a `StdIn` and `StdOut` proxy both in gRPC and "Command Line Interface" mode. The CLI is also responsible to manage the `gdb` process life-cycle gracefully.

The creation of the board specific `gdb` command with all its options is made with the same "recipe building" approach as the `compile` and `upload` command, resolving groups of config parameters stored in the `boards.txt` and `platform.txt` core files

The command creation is tested in the `commands/debug/debug_test.go` file in a unit-test fashion. Proper e2e testing for both the gRPC and CLI interfaces will be implemented in a separate PR.

### gRPC Interface
A new `rpc/debug` module is created to implement a gRPC "debug service" in streaming mode, and the `client_example` code is refreshed to implement a sample of use of the debug gRPC interface (please note that you will need a debugger tool to properly use the example)
 
### CLI Interface
The CLI interface is basically a stdIn/Out proxy for the `gdb` command

### Additional Details
`arduino/utils/stream.go` contains Pipe helpers used to implement StdIn and StdOut forwarding for shelled-out commands


commands/debug/debug_test.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	goldCommand := fmt.Sprintf("%s/arduino-test/tools/arm-none-eabi-gcc/7-2017q4/bin//arm-none-eabi-gdb", dataDir) +
+		" -ex target extended-remote |" +
+		fmt.Sprintf(" %s/arduino-test/tools/openocd/0.10.0-arduino7/bin/openocd", dataDir) +
+		fmt.Sprintf(" -s \"%s/arduino-test/tools/openocd/0.10.0-arduino7/share/openocd/scripts/\"", dataDir) +
+		fmt.Sprintf(" --file \"%s/arduino-test/samd/variants/arduino_zero/openocd_scripts/arduino_zero.cfg\"", customHardware) +
+		fmt.Sprintf(" -c \"gdb_port pipe\" -c \"telnet_port 0\" -c init -c halt %s/hello.arduino-test.samd.arduino_zero_edbg.elf", sketchPath)
+
+	if runtime.GOOS == "windows" {
```
masci: You could build all the paths with `/` and then call https://golang.org/pkg/path/filepath/#FromSlash which will provide the right separators for each platform

commands/debug/debug.go
```diff
@@ -0,0 +1,222 @@
...
+		return &dbg.DebugResp{Error: err.Error()}, nil
+	}
+
+	go func() {
+		// Copy data from passed inStream into command stdIn
+		io.Copy(in, inStream)
+		// In any case, try process termination after a second to avoid leaving
+		// zombie process.
+		time.Sleep(time.Second)
+		cmd.Process.Kill()
```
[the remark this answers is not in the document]
cmaglie: Yes, it happens when the pipe is not gracefully terminated (for example if we kill the client with CTRL-C), in this case we do not receive the EOF and the process is left zombie.
We may catch this error and kill the process only in that case, but in the end we opted to kill the process anyway (after a second of delay) to avoid unforeseen cases like this one.
If the process terminates normally, the `Kill` will just fail silently.

commands/debug/debug_test.go
cmaglie: Windows is able to handle paths with forward slash `/`, even mixed with `\`, there is no need to use `FromSlash` or `ToSlash` (we used it in the past, before discovering this nice fact :-))

commands/debug/debug.go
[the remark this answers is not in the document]
rsora: Basically the only case of graceful `gdb` process close is when a `quit` command is sent. 
Other cases has to be managed by the cli.

commands/debug/debug_test.go
rsora: I have used `FromSlash` to sanitize the golden path and the generated path, this way the test is simpler, Anyway a "windows" switch is necessary, because of `.exe` extension for tools binaries

commands/debug/debug.go
```diff
@@ -0,0 +1,221 @@
...
+func getCommandLine(req *dbg.DebugConfigReq, pm *packagemanager.PackageManager) ([]string, error) {
+	// TODO: make a generic function to extract sketch from request
+	// and remove duplication in commands/compile.go
+	if req.GetSketchPath() == "" {
+		return nil, fmt.Errorf("missing sketchPath")
+	}
+	sketchPath := paths.New(req.GetSketchPath())
+	sketch, err := sketches.NewSketchFromPath(sketchPath)
+	if err != nil {
+		return nil, fmt.Errorf("opening sketch: %s", err)
```
masci:
```suggestion
		return nil, errors.Wrap(err, "opening sketch")
```

commands/debug/debug.go
```diff
@@ -0,0 +1,221 @@
...
+	}
+
+	toolProperties.SetPath("build.path", importPath)
+	toolProperties.Set("build.project_name", importFile)
+	uploadFile := importPath.Join(importFile + ext)
+	if _, err := uploadFile.Stat(); err != nil {
+		if os.IsNotExist(err) {
+			return nil, fmt.Errorf("compiled sketch %s not found", uploadFile.String())
+		}
+		return nil, fmt.Errorf("cannot open sketch: %s", err)
```
masci:
```suggestion
		return nil, errors.Wrap(err, "cannot open sketch")
```

commands/debug/debug.go
```diff
@@ -0,0 +1,221 @@
...
+	fqbnIn := req.GetFqbn()
+	if fqbnIn == "" && sketch != nil && sketch.Metadata != nil {
+		fqbnIn = sketch.Metadata.CPU.Fqbn
+	}
+	if fqbnIn == "" {
+		return nil, fmt.Errorf("no Fully Qualified Board Name provided")
+	}
+	fqbn, err := cores.ParseFQBN(fqbnIn)
+	if err != nil {
+		return nil, fmt.Errorf("incorrect FQBN: %s", err)
```
masci:
```suggestion
		return nil, errors.Wrap(err, "error parsing FQBN")
```

commands/debug/debug.go
```diff
@@ -0,0 +1,221 @@
...
+	}
+
+	// Find target board and board properties
+	_, _, board, boardProperties, _, err := pm.ResolveFQBN(fqbn)
+	if err != nil {
+		return nil, fmt.Errorf("incorrect FQBN: %s", err)
```
masci:
```suggestion
		return nil, errors.Wrap(err, "error resolving FQBN")
```

commands/debug/debug.go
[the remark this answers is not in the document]
matthijskooijman: I'm not sure I understand this entirely. If the process is terminated ungracefully, why kill it (it will be already dead). Also, zombie processes are processes which are actually dead, but are still waiting for their parent to reap them (e.g. read their exit status). This should be done using a "wait" call, which I suppose kill might do internally, but which is also done a few lines below, so I wonder why this kill is really needed, then.

cli/debug/debug.go
```diff
@@ -0,0 +1,96 @@
...
+		Short:   "Debug Arduino sketches.",
+		Long:    "Debug Arduino sketches. (this command opens an interactive gdb session)",
+		Example: "  " + os.Args[0] + " debug -b arduino:samd:mkr1000  /home/user/Arduino/MySketch",
+		Args:    cobra.MaximumNArgs(1),
+		Run:     run,
+	}
+
+	debugCommand.Flags().StringVarP(&fqbn, "fqbn", "b", "", "Fully Qualified Board Name, e.g.: arduino:avr:uno")
+	debugCommand.Flags().StringVarP(&port, "port", "p", "", "Upload port, e.g.: COM10 or /dev/ttyACM0")
+	debugCommand.Flags().StringVarP(&importFile, "input", "i", "", "Input file to be uploaded for debug.")
```
matthijskooijman: This says "to be uploaded", but I did not see any code that actually uploads? Instead, I *think* it assumes that this .elf file is already uploaded?

commands/debug/debug.go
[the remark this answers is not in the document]
cmaglie: The process is not zombie, is just "hanging" (this is a mistake in the comment).
In other words if `gdb` doesn't receive the `quit` command it just sits there: it's not dead, it's not zombie... it's just running waiting for input.

If something goes wrong on the IDE side or on the OpenOCD side, for any reason, gdb will not receive `quit` and will not close itself gracefully, the only way out is to kill it. But we have no way to understand if everything went smooth unless we inspect the stdin/out traffic...

That's the reason why we kill it anyway. We could check if the process is still running and kill it only if still running, but it will just complicate things without a real benefit.

> This should be done using a "wait" call, which I suppose kill might do internally

No kill doesn't wait, the `wait` syscall is made here https://github.com/arduino/arduino-cli/pull/590/files#diff-ec656dd0bf60c33c4e4144e7d5e9508aR86 outside of the go-routine, basically it waits for the subprocess to terminate to finish the external debug call.

commands/debug/debug_test.go
cmaglie: Turned out that to successfully use this pattern we had to turn everything to **forward** slashes `/`...
https://github.com/arduino/arduino-cli/pull/597/files#diff-ec656dd0bf60c33c4e4144e7d5e9508aR57

more on that is coming...

cmaglie:
Hey @matthijskooijman,

yes this is a prototype version, we will surely iterate on this one. BTW it has been very helpful to find critical issues and to see if the whole thing works... :-)

> The handling of the tmp_file property and the importpath seems a bit fragile. What seems to be needed is the name of the .elf file produced by the build, so maybe that should just be exposed (by platform.txt) in a new property somehow for cli to read and re-exposed (by cli to platform.txt) or possibly replaced by the specified importPath? Or alternatively, some import_file property can be set and used by platform.txt (setting it to the .elf file generated by the build) for when no importFile was specified on the commandline, and this property can then be overwritten by cli when an importFile is specified on the commandline?

The filename of the `.elf` is already (almost) exposed in platform.txt for example samd has:
```
build.preferred_out_format=bin
## Save hex
recipe.output.tmp_file={build.project_name}.{build.preferred_out_format}
recipe.output.save_file={build.project_name}.{build.variant}.{build.preferred_out_format}
```
We assume that there is a .elf equally named as the corresponding .bin.
The problem is that this is produced in the temp folder after the build, so the cli copies it inside the sketch when compiled. The remainder of the importpath handling is to retrieve the correct build artifact in all cli use-cases, this piece has been copied from the upload command and is duplicated code, it must be factored into a general purpose function.

> Related, I'm not sure I understand the name "importFile". I can imagine that using elfFile is too arch-specific, but I'm not sure what "import" means in this sense?

I think it comes from the Java IDE where you "Export .hex file", this being the opposite we used import. Nothing stop up from changing it BTW.

> Currently, the board explicitly selects the debug tool to use with it, which works for e.g. the Arduino Zero that has a debug tool built in. However, when using an external debugger (e.g. the atmel ICE, or JTAGICE3, etc.), the board definition cannot know which tool is selected, so some external selection must be used (I think this is analogous to the "Upload" vs "Upload using programmer" distinction. Any thoughts on how to handle this yet?

This is something to prepare for the next iteration together with the "Upload using programmer" feature, those are strictly correlated. This could be the chance also to clarify a bit the programmers.txt definitions.

> Selecting the board to debug might also need some additional thought. I think (mostly based on the java IDE, maybe arduino-cli can do more) that board selection currently happens exclusively based on serial ports (the java IDE can also do network ports, but I'm not quite sure how that works internaly). This works well for serial uploads, and could work well enough as long as the board to use exposes a serial port. However, some boards (or programmers/debuggers) do not expose a serial port (or, the tool that uploads does not know about serial ports). I actually suspect that this is already the case for the Arduino zero, that it just uploads to any cmsis-dap USB device it finds, completely ignoring the selected serial port. For this to work, there should be some way to select USB devices (probably based on vidpid or other filters) and expose them to tools (which might need multiple and platform-specific identifiers, e.g. Linux can specify as "bus 1, device 23", but also a port-path like "1-3.2:1.3"). This is probably an issue completely separate from debugging, though.

I think we just need to provide a port parameter (port to the programmer of course, not the board), or none the programmer is autodetected. About the different "kinds" of ports the pluggable discovery mechanism will allow to discover different types of ports with their properties attached to them. The pluggable discovery is another thing planned for the release 1.0 of the arduino-cli...

> The gdb command might need an extra --nx to prevent reading the .gdbinit file (or files maybe). Maybe also --quiet to prevent printing startup messages?

`--nx` may be added for sure, `--quiet` I don't know, maybe it could be helpful to have the startup message in some cases.

> In my original prototype, I also used --interpreter=mi2 to switch gdb from its normal "human readable" command mode into a slightly more machine-readable format (which also has better defined semantics for connecting commands and replies, IIRC). This is probably not suitable for direct commandline usage, but I can imagine that mi2 might be easier to work with for IDEs that will run on top of this. Maybe both would need to be supported? Or maybe the IDE can switch to mi2 mode using a normal mode command? Or is normal mode good enough in practice (i.e. does the IDE the Pro IDE is based on use it by default)?

we added `--interpreter=mi2` to the recipe in the platform.txt, this works for the IDE but not so much if you run the debug interactively by command-line. Maybe we can add a command line switch, or better, try to switch mode from the IDE by sending command in normal mode as you suggested. Again, room for improvements on the next iterations.

Thanks again for your comments @matthijskooijman, now that I re-read it there is a lot going on, it may be better to open separate issues for each part...

PR #609, opened 2020-03-06
Added configuration option for proxy usage

cmaglie:
This PR makes a bunch of changes:

- ~~use `downloader.SetDefaultConfig()` to set globally the user-agent: this allows to avoid passing headers back and forth~~
- ~~previously in the `core update-index` command user agent headers wasn't set when downloading: with the global `downloader.SetDefaultConfig(...)` this is no more a problem.~~
- added support for network proxy configuration
- refactored the function that generates the user-agent to add a subcomponent part (we use it in the `daemon` mode).
- that showed we make an http API call in `board list` without using `downloader` library, so it ignores the global network settings. I've added a TODO for missing proxy usage.

### DIFF
--- a/arduino/cores/packagemanager/download.go
+++ b/arduino/cores/packagemanager/download.go
@@ -17,7 +17,6 @@ package packagemanager
 
 import (
 	"fmt"
-	"net/http"
 
 	"github.com/arduino/arduino-cli/arduino/cores"
 	"go.bug.st/downloader"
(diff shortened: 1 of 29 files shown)


cmaglie:
In the meantime I've changed the implementaion.

- we do not use downloader.SetDefaultConfig() anymore
- now we have a `command.GetDownloaderConfig()` method that returns a `downloader.Config` based on the current configuration stored in `viper`. This way any change on the configuration is immediately reflected on subsequent downloads

PR #512, opened 2019-12-12
Improve precompiled libraries handling

facchinm:
General change:
 - library compilation bails out if the precompiled object is found.
   This allows mixed libraries that fallback using sources if no suitable precompiled object is found

ARM float specification change:
 - Cortex M4+ allows specifying different flags for floating point ABI
   This patch introduces a second level of subfolder that MUST be used if -mfpu or -mfloat-abi are specified on the commandline
   Since there's no clear specification (unlike build.mcu), the values are extracted from c++ recipe

   For example, for a target which commandline contains `-mfloat-abi=hard -mfpu=fpv4-sp-d16` , the precompiled search path will be
   `$libfolder/cortex-m4/fpv4-sp-d16-hard`
   If that folder doesn't exist the library will be compiled from sources

- Open questions:
  While reviewing this I found a code path just to add static libraries with non standard name (not starting with "lib" ). Should we keep it or not? @cmaglie 
  I don't like the name of the subfolder but it's super precise. We should be able to get away with `-float-abi=xxx` only but I'm not 100% sure. Experts' advise are welcome :slightly_smiling_face: 

Fixes https://github.com/arduino/arduino-builder/issues/256

PR #611, opened 2020-03-10
Fix mixed code precompiled libraries

facchinm:
This patch restores some functionalities broken by https://github.com/arduino/arduino-cli/pull/512 .
It introduces a BREAKING CHANGE to "precompiled" field

If merged, "precompiled" means that the library ONLY contains a precompiled library and its headers.
No file containing implementation will be compiled if the library is specified as such.

Fixes https://github.com/arduino/arduino-builder/issues/353

Tested with
* https://github.com/arduino/arduino-cli/pull/512#issuecomment-565070453 (Nano33BLE and generic MKR board) -> same code and precompiled
* https://github.com/BoschSensortec/BSEC-Arduino-library (works after removing precompiled=true directive in library.properties) -> mixed (different code and precompiled)
* https://github.com/vidor-libraries/USBBlaster (MKRVidor) -> only precompiled


legacy/builder/phases/libraries_builder.go
```diff
@@ -134,8 +134,14 @@ func fixLDFLAG(ctx *types.Context, libs libraries.List) error {
 			}
 		}
 
-		currLDFlags := ctx.BuildProperties.Get(constants.BUILD_PROPERTIES_COMPILER_LIBRARIES_LDFLAGS)
-		ctx.BuildProperties.Set(constants.BUILD_PROPERTIES_COMPILER_LIBRARIES_LDFLAGS, currLDFlags+"\"-L"+path.String()+"\" "+libsCmd+" ")
+		key := "compiler.libraries.ldflags"
```
facchinm: Since the linking order is crucial here, I would avoid the fallback if it **can** fail

cmaglie: I've removed the extra commit, it's worth keeping the PR clean anyway IMHO

PR #632, opened 2020-03-23
update compile --build-path to support both relative and absolute flags

### LINKED ISSUE #630: "compile --build-path" does not work for relative paths
I cannot specify a relative path for `compile --build-path`.

In other words, this does not work:
```
arduino-cli compile -b adafruit:avr:protrinket3 . -v --build-path ./_build
[... snip ...]
Linking everything together...
Error during build: Rel: can't make /home/ktdreyer/dev/arduino/_build/core/core.a relative to _build
```

It took me a while to understand this error message.

It does work when I specify an absolute path:
```
arduino-cli compile -b adafruit:avr:protrinket3 . -v --build-path $PWD/_build
```

It would be great to make the following improvements:
1) (Easy) Update the help text in `compile --help` to indicate that `--build-path` must be an absolute path.
2) (Harder) Support relative paths for `--build-path`.

vinay-lanka:
Small change in compile.go to allow people to use either relative or absolute paths
A person would use relative paths assuming he wants it in the same directory as that of the sketch, allowing us to use the sketch path.

Can fix #630

### DIFF
--- a/commands/compile/compile.go
+++ b/commands/compile/compile.go
@@ -123,7 +123,12 @@ func Compile(ctx context.Context, req *rpc.CompileReq, outStream, errStream io.W
 	builderCtx.OtherLibrariesDirs.Add(configuration.LibrariesDir())
 
 	if req.GetBuildPath() != "" {
-		builderCtx.BuildPath = paths.New(req.GetBuildPath())
+		if req.GetBuildPath() != "/" {
+			builderCtx.BuildPath = paths.New(req.GetSketchPath() + strings.TrimPrefix(req.GetBuildPath(), "."))
+		} else {
+			builderCtx.BuildPath = paths.New(req.GetBuildPath())
+		}
+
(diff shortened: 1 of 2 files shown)


rsora:
@vinay-lanka thanks for submitting this PR!
I think that this fix should be placed at a lower level( `commands/compile/compile.go`). 

Can you please also add a couple of tests  (including the one in the issue https://github.com/arduino/arduino-cli/issues/630) in order to ease the reviewprocess?

As a side note, can you please rebase your code on master?

vinay-lanka:
Hey,
I placed the change at a lower level in `command/compile/compile go` and wrote tests according to #630 
Any idea why the windows do test is failing? My local integration tests ran fine.

(Also sorry for the messy intermediate commit history, it's cleared now)

vinay-lanka:
Should I skip the windows CI test on this? I do not understand why the assertion error is like that.

PR #633, opened 2020-03-24
Add macro during lib discovery phase

facchinm:
This information can be used by the core to turn off certain heavyweight headers (that don't need to be discovered anyway)

### DIFF
--- a/legacy/builder/gcc_preproc_runner.go
+++ b/legacy/builder/gcc_preproc_runner.go
@@ -77,6 +77,8 @@ func prepareGCCPreprocRecipeProperties(ctx *types.Context, sourceFilePath *paths
 	// to create a /dev/null.d dependency file, which won't work.
 	cmd.Args = utils.Filter(cmd.Args, func(a string) bool { return a != "-MMD" })
 
+	cmd.Args = append(cmd.Args, "-DARDUINO_LIB_DISCOVERY_PHASE")
+
 	return cmd, nil
 }
 


fpistm:
> This information can be used by the core to turn off certain heavyweight headers (that don't need to be discovered anyway)

Hi @facchinm 

when you said it can be used, how ?

In fact we've met an issue since this has been added. It add an extra definition at the end of the command line but the compiler used doesn't understand it (https://github.com/stm32duino/Arduino_Core_STM8/issues/21):
` unknown file type -DARDUINO_LIB_DISCOVERY_PHASE_258740\sketch\master_reader.pde.cpp:`

as for the specific cosmic compiler used for the STM8 core "-D" is replaced by "-d"

Is it possible to avoid to add extra arguments?

Thanks in advance

cmaglie:
hi @fpistm 

I didn't know that the STM8 compiler use `-D` instead of `-d`, never heard of any other gcc-compatible compiler doing so, I guess it's kind of unique.

The definition `-DARDUINO_LIB_DISCOVERY_PHASE` has been added to allow exclusion of some header files during the library discovery, since during library discovery the compiler is launched many times this allows huge speedups if big headers are included from the sketch (that was the case for the `mbed` core for example).

BTW let's fix this, I'll open a new issue with a proposal.

fpistm:
Hi @cmaglie 
thanks for the quick feedback. 
And right this is special 😕

PR #634, opened 2020-03-25
Bump required go version to 1.14

matteosuppo:
Fixed faulty json because of new behaviour of encoding/json:

> encoding/json
> Number no longer accepts invalid numbers, to follow the documented behavior more closely.

Signed-off-by: Matteo Suppo <matteo.suppo@gmail.com>


arduino/cores/packageindex/testdata/package_Zoubworld_index.json
```diff
@@ -11,61 +11,61 @@
       },
       "platforms": [
         
-{
```
rsora: Why this section is changed? :thinking:

matteosuppo: It probably converted spaces into tabs or viceversa

PR #656, opened 2020-04-14
Implementation of a Board-based flow (WIP)

cmaglie:
This PR tries to implement a board-centric flow like the following:

```
$ arduino-cli board install zero
....
$ arduino-cli upload -b zero
...
```

In the examples above `zero` is a board Alias. Each board alias corresponds to a FQBN (for example `zero` corresponds to `arduino:samd:arduino_zero_edbg`). There should be a table that maps Alias and FQBN 1:1 but how this is obtained is still TBD.

- A board alias can be used anywhere a FQBN was previously expected. The old flag `--fqbn` is still accepted but has been deprecated in favor of `--board/-b` that can accept both alias and FQBN.
- FQBNs may carry also configuration options for a board. To allow configuration using aliases an extra flag has been added `--board-config/-c` that can accept a config option. The flag may be used multiple times. For example:
   `arduino-cli board details arduino:avr:nano:cpu=atmega168`
can now be rewritten as:
   `arduino-cli board details nano -c cpu=atmega168`
or:
   `arduino-cli upload -b arduino:avr:nano:cpu=atmega168`
can now be rewritten as:
   `arduino-cli upload -b nano -c cpu=atmega168`


cmaglie:
Too much time passed, and I'm closing this one until the board-based flow will regain traction.

PR #672, opened 2020-04-23
add cli httpclient with support for proxy configuration

hdiniz:

A default HTTP client for the cli which handles proxy configuration and default HTTP headers (User-Agent)

HTTP calls in the cli that don't use the downloader are not using the proxy configuration or the user-agent header

All HTTP calls in the cli are proxy-enabled and submit the cli User-Agent

No

* **Other information**:

---

### DIFF
--- a/httpclient/httpclient_config.go
+++ b/httpclient/httpclient_config.go
@@ -1,3 +1,18 @@
+// This file is part of arduino-cli.
+//
+// Copyright 2020 ARDUINO SA (http://www.arduino.cc/)
+//
+// This software is released under the GNU General Public License version 3,
+// which covers the main part of arduino-cli.
+// The terms of this license can be found at:
+// https://www.gnu.org/licenses/gpl-3.0.en.html
+//
+// You can be released from the requirements of the above licenses by purchasing
+// a commercial license. Buying such a license is mandatory if you want to
+// modify or otherwise use the software for commercial activities involving the
+// Arduino software without disclosing the source code of your own applications.
+// To purchase a commercial license, send an email to license@arduino.cc.
+
 package httpclient
 
 import (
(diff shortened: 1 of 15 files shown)


httpclient/httpclient_config.go
```diff
@@ -0,0 +1,64 @@
...
+	"fmt"
+	"net/url"
+	"runtime"
+
+	"github.com/arduino/arduino-cli/cli/globals"
+	"github.com/spf13/viper"
+)
+
+// Config is the configuration of the http client
+type Config struct {
```
cmaglie, line 29: This is almost identical to `downloader` config:
https://github.com/bugst/go-downloader/blob/deb605b33c14bd1a66cbd2d1d1cb634af9dd7b2f/downloader_config.go#L14-L22

Also adding an `httpclient` module seems like a lot of duplication because even if we now have `httpclient` and `downloader` we still need to implement a download loop:
https://github.com/arduino/arduino-cli/blob/83d7f6e33d556b272825f9f699873f8399bb826d/commands/board/list.go#L59-L69

Can't you add a pair of functions in the `downloader` so we can reuse the `downloader.Config`? something like:

`downloader.DownloadInBufferWithConfig(buff bytes.Buffer, reqURL string, config Config, options ...DownloadOptions) (*Downloader, error)`

 (disclaimer: we own the library we can change as we wish) We already have a function to create a `downloader.Config` here: https://github.com/arduino/arduino-cli/blob/master/commands/download.go#L32

cmaglie: Now that I'm thinking of it, `bytes.Buffer` implements the `io.Writer` interface so it would be even more general to do something like:
`downloader.StreamWithConfig(out io.Writer, reqURL string, config Config, options ...DownloadOptions) (*Downloader, error)`

PR #717, opened 2020-05-21
[skip changelog] Configure git push from CI

masci:
- [ ] The PR has no duplicates (please search among the [Pull Requests](https://github.com/arduino/arduino-cli/pulls) before creating one)
- [ ] The PR follows [our contributing guidelines](https://arduino.github.io/arduino-cli/CONTRIBUTING/#pull-requests)

This PR fixes the generation pipeline for versioned docs.

Most of the feature is implemented in master but docs are never pushed to the `gh-pages` branch. Also there's a bug in the version selector component that breaks it when docs are not at the root path `/`.

This PR implements several bugfixes:
 * versioned docs are now pushed to `gh-pages`
 * the docs pipeline is now triggered when changes are made to the GH workflow itself
 * The command line interface of the `build.py` script has been expanded to simplify the code needed from within the workflow yaml definition
 * The javascript component implementing the version selector has been fixed

No changes to the CLI in this PR

---

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -16,7 +16,6 @@ on:
   push:
     branches:
       - master
-      - massi/publish
       # release branches have names like 0.8.x, 0.9.x, ...
       - '[0-9]+.[0-9]+.x'
     # At this day, GitHub doesn't support YAML anchors, d'oh!
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -14,6 +14,7 @@ on:
   push:
     branches:
       - master
+      - massi/publish
       # release branches have names like 0.8.x, 0.9.x, ...
       - '[0-9]+.[0-9]+.x'
     # At this day, GitHub doesn't support YAML anchors, d'oh!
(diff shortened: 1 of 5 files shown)


matthijskooijman:
I hate to be a nag, but if the pull request template is not filled in, I would suggest removing it (or at least the parts that are apparently not relevant) and maybe providing at least a brief description of the pullrequest if the title is not enough. Doing so makes the pullrequests easier to view, and also makes the notification e-mails I get a lot easier.

While I'm here: I would also suggest adding a bit more info in commit messages. Instead of just documenting **what** changed, also document **why** it needs to change or why it is helpful, which makes it easier to review and easier to trace the origin of code later. It does seem that short commit messages are more commonly used in this repository, so this is probably more of a general remark, not specific to this PR.

One remark that **is** specific to this PR: It seems there is a test commit in here which probably serves to test the workflow and should be removed before merging. In such cases, I would suggest marking the PR as "draft" to prevent it from being accidentally merged.

masci:
@matthijskooijman this should be a draft, my apologies.

> I hate to be a nag

Then please don't be. I'm the author of the contributing guidelines for this repo and I know I should lead by example but I'm also going above and beyond to deliver this feature while swamped. I think I've done enough for this project to deserve more than a 20 minutes buffer before getting so much criticism about my work.

matthijskooijman:
> but I'm also going above and beyond to deliver this feature while swamped.

My comment was meant as friendly and constructive feedback. I can understand that time pressure or other circumstances sometimes force suboptimal choices to be made (and also make feedback look like harsher criticism than it was intended), but I could not tell that from here, of course. Apologies if my comment offended you, that was genuinely not the intention.

Anyway, good luck with your workload, and thanks for all the work you've put in already!

PR #727, opened 2020-05-27
Reduce codecov threshold to 30%

cmaglie:
We are currently hitting ~31.8% of test coverage, keeping the threshold to 35% will mark all the commits as failing, for example int [the latest merge](https://github.com/arduino/arduino-cli/commit/f5b41257d7e0ba721562fabf57ad91163eeea9d6):

image

but this is the same for all the recent merge commits.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -7,7 +7,7 @@ coverage:
   status:
     project:
       default:
-        target: 35%
+        target: 30%
         threshold: null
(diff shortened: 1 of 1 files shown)


rsora:
I state publicly here before merge that we will plan a coverage increase activity in the following months, to raise the coverage bar again :mag_right: :bug:

PR #744, opened 2020-06-08
Fixed skethbook+bootloader hex merger.

cmaglie:
Previous hex merger:
- doesn't take into account .bin bootloaders
- it may have problems if .hex data is not sorted by address.

This PR use a proper .hex parser and fix the above problems.

This PR should not introduce regressions, but testing is required.

Fix:
https://github.com/arduino/arduino-builder/pull/295
https://github.com/arduino/arduino-builder/issues/286


legacy/builder/merge_sketch_with_bootloader.go
```diff
@@ -66,68 +69,86 @@ func (s *MergeSketchWithBootloader) Run(ctx *types.Context) error {
...
-		sketch = append(sketch, row)
+	if mergeFile, err := mergedSketchPath.Create(); err == nil {
+		defer mergeFile.Close()
+		memMerged.DumpIntelHex(mergeFile, 16)
+	} else {
+		return err
 	}
 
-	return mergedSketchPath.WriteFile([]byte(strings.Join(sketch, "\n")))
+	// size := lastAddress - initialAddress
```
facchinm, line 145: I would restore this part, the merged bin is super useful

Issue #639, opened 2020-04-02
Builder: Phase out `--relax` option adding


matthijskooijman:
When building a sketch, if the MCU is set to atmega2560, `,--relax` is added to the linker commandline:

https://github.com/arduino/arduino-cli/blob/322bc6a59248e63e3ff331e8eda6e72d384a1d6e/legacy/builder/phases/linker.go#L80-L85

This is something that really should be taken care of in the `platform.txt` file, not in the builder, since the current code assumes to much about the platform, such as that you're actually using gcc, and that the last option on the link commandline is actually a `-Wl` option (otherwise [things break](https://github.com/arduino/Arduino/commit/4b4f185239eae3cfe7968f313ee65dbbc28d7e7b)).

This seems to stem from way back when the Arduino IDE still generated commandlines explicitly: https://github.com/arduino/Arduino/commit/fa4ab4f6ab07117dd95ad284b2b9afff5c81c376

However, it seems that this option can be just be added to the commandline unconditionally, it is potentially helpful for any board with > 8K flash. Maybe this was different for the compiler versions used in 2011 when this was added, but I actually suspect this option was just misunderstood.

I just tried adding the option on a Uno board (which saves a few bytes on the empty sketch) and an Arduino NG with atmega8, which also accepts the option but as expected it does not make a difference.

I'll provide a PR for the AVR core in a minute.

Once that is done, I think this code should be removed from arduino-cli. How shall we approach this? I can imagine adding a check to see if the option is present in platform.txt and if not show a warning to notify core authors to update their core, except that we do not really have any machinery to toggle such warnings meaningfully (so it might just confuse users).

Since it is just an optimization (without it, some calls might end up 2 bytes longer), we could also consider just removing this?

matthijskooijman:
Pullrequest for ArduinoCore-AVR is here: https://github.com/arduino/ArduinoCore-avr/pull/327

facchinm:
Any concern in merging this? It would also incidentally solve https://github.com/arduino/ArduinoCore-avr/issues/339#issuecomment-625602068

matthijskooijman:
It would indeed "solve" that, at least until https://github.com/arduino/ArduinoCore-avr/pull/327 is merged.

facchinm:
Correct, but before merging arduino/ArduinoCore-avr#327 we should check if it's still an hard requirement with latest toolchains

cmaglie:
[...]

PR #749, opened 2020-06-11
Removed automatic '--relax' hack for compiles on Atmega2560 MCU

cmaglie:
As discussed in: https://github.com/arduino/arduino-cli/issues/639

This change has no effects on all board with MCU != Atmega2560.
AFAIU the `--relax` option should be no more needed with the latest available avr compilers.

### DIFF
--- a/legacy/builder/phases/linker.go
+++ b/legacy/builder/phases/linker.go
@@ -57,13 +57,11 @@ func (s *Linker) Run(ctx *types.Context) error {
 }
 
 func link(ctx *types.Context, objectFiles paths.PathList, coreDotARelPath *paths.Path, coreArchiveFilePath *paths.Path, buildProperties *properties.Map) error {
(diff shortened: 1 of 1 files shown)


matthijskooijman:
>  AFAIU the --relax option should be no more needed with the latest available avr compilers.

Do you mean the gcc7 we use? Or later versions? Since with gcc7, removing `--relax` seems to prevent a miscompile, so it does actually change code (though maybe LTO somehow applies a double correction here were LTO relaxes and then the linker erronously relaxes again, then `--relax` would indeed not be needed).

matthijskooijman:
Hm, did you intentionally not include my suggested addition to the commit message? Should I have just pushed it to the branch directly (I think I could have)?

cmaglie:
I added it in the squash

matthijskooijman:
Ah, should have checked the master history. I had somehow assumed that this would have looked differently for a squash commit, but apparently not. Sorry for that, and thanks! :-)

PR #763, opened 2020-06-25
Daemon listen to localhost

cmaglie:
Previously the daemon would listen to all interfaces, now it will listen only to localhost.

/cc @luigigubello


kittaakos:
@rsora, @cmaglie This was a breaking change for the Pro IDE, please give us a heads-up if you change the daemon messages, we are listening on to them. 😊 Thanks!

PR #766, opened 2020-06-26
Added support for version pinning in install.sh

cmaglie:
Usage:

to install the latest version:
```
./install.sh
```

to install a specific version:
```
./install.sh 0.9.0
```

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -4,6 +4,16 @@
 # See https://github.com/Masterminds/glide/blob/master/LICENSE for more details
 # and copyright notice.
 
+#
+# Usage:
+#
+# To install the latest version of the CLI:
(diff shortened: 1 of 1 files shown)


rsora:
Tested like this to install the `0.9.0`version and it works
```sh
curl -fsSL https://raw.githubusercontent.com/arduino/arduino-cli/043134cba7c09c1af4b587ef4a3dfffca338a217/install.sh | BINDIR=~/.local/bin sh -s 0.9.0
```
I'll open a separate PR to update the documentation.
cc @janjongboom

janjongboom:
@rsora Thanks, we've updated our CI scripts!

PR #791, opened 2020-07-06
Implemented package_index.json signature verification

cmaglie:
- a `.sig` file is downloaded together with package file on `downloads.arduino.cc` domain
- the signature is verified against the Arduino public GPG keys
- if the signature is valid both files, index and signature, are copied over older ones in data directory


rsora:
(BTW let's keep a close look on GeertJohan/go.rice#159 wink )

Issue #861, opened 2020-07-23
[gRPC] Check whether CompileReq#programmer is required


kittaakos:
I have noticed, the `CompileReq` has the `programmer` field:

https://github.com/arduino/arduino-cli/blob/3d74a8f64462afe0caef4b8857e7c3fe2bfdc631/rpc/commands/compile.proto#L43

One can trigger an automatic upload with the `compile` command from the command line by appending the `-u` (`--upload`) flag, but it is not possible to trigger an upload with the `CompileReq` via gRPC.

Long story short: the `programmer` seems superfluous. Please check whether it is needed or not.

Thank you!

PR #862, opened 2020-07-23
Removed useless Compile.programmer parameter from GRPC interface

cmaglie:
It's never used, I guess it's a leftover from previous experiments.


rsora:
Closes https://github.com/arduino/arduino-cli/issues/861

PR #909, opened 2020-08-19
Add "core add-index" command to add additional urls

silvanocerza:

      before creating one)
- [x] The PR follows

A new feature.

Adding a new url to `additional_urls` list in user's settings is a manual process.

Adding a new url to `additional_urls` list is done with the `core add-index` command.

Nope.

* **Other information**:
None.

---


cmaglie:
Now that I'm looking at the implementation I'm not sure that this should be a GRPC service, if it is going to update the `arduino-cli.yml` configuration file then this is something that is related to the "CLI" side only, GRPC consumers may handle the config by themself.

Since we are here I would add also a commands to list and remove, something like:

```
$ arduino-cli core index add URL...
$ arduino-cli core index remove URL...

# and to list
$ arduino-cli core index list
# or even better just
$ arduino-cli core index
```

silvanocerza:
Yeah, I was doubtful about it, I'll change it so that it's only on the CLI side. 
The gRPC interface is useless at that point, guess I'll remove it.

I like the suggestion about the other commands but I'd like to hear what @ubidefeo thinks about it too.

ubidefeo:
@cmaglie @silvanocerza 

>`$ arduino-cli core index add {URL}`
`$ arduino-cli core index remove {URL}`
and to list
`$ arduino-cli core index list`

I very much like the idea of having a command for `index` with subcommands for `add`, `remove`, and `list`.
Without going into the interactive prompt paradigm (sorry... it's a dream), can we think of a way to `edit` a url from the index list?

on the last option
>or even better just
`$ arduino-cli core index` 

I believe it breaks behaviour.
If `index` is thought to be a command then by itself it should just display the contextual help like it happens for `board`, `core`, `lib` etc

On a final note, I'm a bit confused about this
> this is something that is related to the "CLI" side only, GRPC consumers may handle the config by themself.

One of the consumers is the Pro IDE, which needs a way to add custom URLs to a list of providers in its own `arudino-cli.yaml`

silvanocerza:
> One of the consumers is the Pro IDE, which needs a way to add custom URLs to a list of providers in its own `arudino-cli.yaml`

I think @cmaglie is refering to the fact that the gRPC interface also manages the `arduino-cli.yaml` file instead of letting the consumer do it, the Pro IDE in our case.

Either way it's fine for me, it would makes sense to remove the gRPC interface if the Pro IDE actually doesn't use the `arduino-cli.yaml` to store settings.

silvanocerza:
@cmaglie @ubidefeo I think we close this PR in favor of the addition of the `add`, `edit` and other subcommands to the `config` command, what do you say?

ubidefeo:
@silvanocerza 
go on :)

silvanocerza:
Closing this in favor of a different solution that will be implemented in the future.

PR #911, opened 2020-08-20
Allow post_install for untrusted platforms

cmaglie:
Follows https://github.com/arduino/arduino-cli/pull/893#issuecomment-673556652

> Does it really add anything to add an isTrusted flag here? AFAICS, an "untrusted" platform can still provide custom compiler binaries and compiler commands in platform.txt, so this trusted flag would only prevent running untrusted code at platform install time, as soon as you actually use the platform to compile a sketch, untrusted code will be ran in any case?

/cc @luigigubello @rsora

### DIFF
--- a/commands/core/install.go
+++ b/commands/core/install.go
@@ -140,7 +140,7 @@ func installPlatform(pm *packagemanager.PackageManager,
 	}
 
 	// Perform post install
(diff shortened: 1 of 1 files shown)


rsora (approved):
LGTM, we will address the concept of "Trusted Platform" in the near future.

Issue #759, opened 2020-06-24
arduino-cli board details crashes with external FQBNs


a7hybnj2:
### Current behavior

```
arduino-cli board details ATTinyCore:avr:attinyx4
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x46b324f]

goroutine 1 [running]:
github.com/arduino/arduino-cli/commands/board.Details(0x4b080a0, 0xc000126008, 0xc000547d60, 0x46bebc7, 0xc000190200, 0xc000589d70)
        /private/tmp/arduino-cli-20200624-55394-jn0e1u/commands/board/details.go:65 +0x4df
github.com/arduino/arduino-cli/cli/board.runDetailsCommand(0xc000168b00, 0xc000119ba0, 0x1, 0x1)
        /private/tmp/arduino-cli-20200624-55394-jn0e1u/cli/board/details.go:57 +0x145
github.com/spf13/cobra.(*Command).execute(0xc000168b00, 0xc000119b80, 0x1, 0x1, 0xc000168b00, 0xc000119b80)
        /Users/a7hybnj2/Library/Caches/Homebrew/go_mod_cache/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:846 +0x29d
github.com/spf13/cobra.(*Command).ExecuteC(0xc000168000, 0x0, 0x0, 0xc00010e058)
        /Users/a7hybnj2/Library/Caches/Homebrew/go_mod_cache/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:950 +0x349
github.com/spf13/cobra.(*Command).Execute(...)
        /Users/a7hybnj2/Library/Caches/Homebrew/go_mod_cache/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:887
main.main()
        /private/tmp/arduino-cli-20200624-55394-jn0e1u/main.go:31 +0x36
```

### Expected behavior
I would like for the command to show me all the available options for the board I am specifying.

### Environment
- CLI version (output of `arduino-cli version`): `arduino-cli Version: HEAD-7541c80 Commit: 7541c80d180106f4370731de45a3ff4cc6709657`
- OS and platform: `macOS 10.15.5`

### Additional Info
```
arduino-cli board listall
Board Name                    FQBN                         
ATtiny1634 (No bootloader)    ATTinyCore:avr:attiny1634    
[...]

PR #924, opened 2020-08-26
Fix NPE for 'board install' command

cmaglie:
The NPE is triggered if Platform.Resource is nil (this may happen forplatforms installed manually in the sketchbook/hardware directory for example).

Related to #759

@matthijskooijman @fpistm @a7hybnj2


matthijskooijman (approved):
Code looks good, just tested this for `adafruit:samd:adafruit_feather_m0` and `STM32:stm32:Nucleo_64` and works like expected. Thanks!

Issue #849, opened 2020-07-22
Feature Request:  Export compile_commands.json


sslupsky:
A useful enhancement to the arduino cli would be the option to generate a compile_commands.json file.  This can be used by Microsoft's Visual Studio Code (vscode) application to configure the its' IntelliSense code completion, parameter info, quick info, and member lists capabilities.  I find using arduino-cli and vscode is particularly productive for embedded development.

sudobash1:
I think this is a cool idea, and I have wanted this feature. I think that it would be at least slightly complicated by how Arduino preprocesses the `.ino` file. It doesn't just feed it directly to the compiler. You can learn about that [here](https://github.com/arduino/arduino-preprocessor#about-the-arduino-sketch-ino-preprocessing).

Without this preprocessing, IntelliSense, `clangd` and `ccls` might think they are looking at invalid C++ (which they would be).

cheers

matthijskooijman:
I'm also interested in this feature, since my current manual compiler-flags configuration for code completion with `clangd` is a bit fragile (and does not have a complete list of files in the project, preventing refactoring from working).

As for the .ino file, it will indeed be pretty much impossible to let clangd see the preprocessed version, especially when you're editing it (unless the Arduino IDE talks to clangd and does the preprocessing on the fly, but that's not the subject of this issue). But I think it would actually be fine if all .ino files would end up in the compilation database with just the flags used to compile the preprocessed version. This means that *if* you have just a single .ino file (or multiple that do not depend on each other without forward declarations), and you do not depend on the autogenerated forward declarations (by defining functions in order and/or adding forward declarations manually), clangd should be able to figure things out most of the time.

One additional thing is that preprocessing adds an `#include <Arduino.h>` line, but I think this could be emulated by adding `-include Arduino.h` to the compiler flags as well (this could even have been used for the actual compilation, except that would break compatibility now).

So, outputting a compilation database would simply be a matter of dumping all compilation commands ran, and some special casing for the .ino files, which should be fairly easy to implement.

Where should this file be generated? I suspect that the sketch directory is the obvious place, since that's also where editors / clangd will look for it, right?

[...]

PR #944, opened 2020-09-11
Generate compile_commands.json compilation database

matthijskooijman:

      before creating one)
- [X] The PR follows

New feature: Allow generating `compile_commands.json` file as a side effect of compilation, for consumption by editors and other tooling, as requested in #849.

All compile commands (not linking and other steps) ran are output to a compilation database, a file called `compile_commands.json` (see https://clang.llvm.org/docs/JSONCompilationDatabase.html for the format). This file can be used by editors (often using `clang` or `clangd`) to provide code completion, navigation and diagnostics.

* **Other information**:
This is just a rough draft, that still needs more work, some decisions and some feedback. I wanted to quickly throw something together, which ended up a bit more work than I thought. The basics work, though this is probably more of a proof of concept that might need to be significantly restructured. Open points are:
 - How to log:  introduced a `CompilationDatabase` class outside of the `legacy/builder` directory (seemed good to not add more legacy), but now I'm unsure how to approach logging. It seems that the cli and builder code have distinct logging facilities, neither of which seems easily accessible. For now I just added "fmt.Print". Suggestions are welcome.
 - Where to write the file: I now write a `compile_commands.json` to the sketch directory, since that's where editors would look for it. I later realized that it would probably be good to *also* write a version to the temp build directory (note: I mean the `/tmp/arduino-sketch-XYZ` directory, not the `build` subdirectory of the sketch dir where the compilation result is exported), for any tooling that wants to inspect the copied and preprocessed sketch files.
 - Updating sketch paths: Currently, the paths are written as they passed to the compiler, so pointing at the copy of the sketch in the temp build dir. This is probably good for the `compile_commands.json` to be written to the temp build dir, but the sketch dir version should really point to the original sketch files. This means that the `file` element in the json should be updated, but maybe also any compiler arguments that resolve to sketch files should have their paths updated. This should probably be applied generically (i.e. search replacing back to the original sketch dir) to all arguments, since we can't be sure what a `platform.txt` does exactly. Alternatively, updating just the `file` field might be sufficient and can be done by passing the original filename around (does require significant refactoring, I think).
 - Handling .ino files: Currently, only one entry is generated for .cpp file resulting from the combined and preprocessed .ino file(s). This is again fine for the temp build version, but for the sketch dir version, I think this should have a separate entry for each .ino file, with an `-include=Arduino.h` flag added (as suggested in https://github.com/arduino/arduino-cli/issues/849#issuecomment-690635998). This is imperfect, but probably as close as we can get.
 - Making this optional: I believe that writing a `compile_commands.json` to the temp build dir can always be done, but writing to the sketch directory should almost certainly be opt-in (through commandline or config), to prevent user surprise and because the IDE and arduino-cli seem to try hard to prevent touching the sketch dir now. For arduino-cli, it might make sense to always generate the file in the `build` subdirectory, along with the compiled program, and provide an option to also, or instead of, generate in the sketch directory itself (where it will be mostly automatically picked up by editors or `clangd`). For the IDE, some extra care should probably be taken to not generate this file in the sketch directory for unmodified examples, even if the user enabled it in the config.
 - Partial compilation: Currently, the file is written from scratch at the end of every compilation. So if compilation fails, or files are not changed and thus not recompiled, files will be missing from the `compile_commands.json` file. Ideally, any existing file should be read and merged, taking care to remove any extra entries (e.g. for deleted files), but only when compilation was successful (or at least complete). Alternatively, a complete list of files to be compiled could also be used maybe (even when compilation is aborted halfway).

### DIFF
--- a/arduino/builder/compilation_database.go
+++ b/arduino/builder/compilation_database.go
@@ -1,0 +1,73 @@
+package builder
+
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/arduino/go-paths-helper"
+	"io/ioutil"
+	"os"
+	"os/exec"
+)
+
+type compilationCommand struct {
+	Directory string   `json:"directory"`
+	Arguments []string `json:"arguments"`
+	File      string   `json:"file"`
+}
+
+type CompilationDatabase struct {
+	contents []compilationCommand
+	filename *paths.Path
+}
+
+func NewCompilationDatabase(filename *paths.Path) *CompilationDatabase {
+	return &CompilationDatabase{
+		filename: filename,
+	}
+}
+
+func (db *CompilationDatabase) UpdateFile(complete bool) {
+	// TODO: Read any existing file and use its contents for any
+	// kept files, or any files not in db.contents if !complete.
+	fmt.Printf("Writing compilation database to \"%s\"...\n", db.filename.String())
+
+	contents := db.contents
+	jsonContents, err := json.MarshalIndent(contents, "", " ")
+	if err != nil {
+		fmt.Printf("Error serializing compilation database: %s", err)
+		return
+	}
+	err = ioutil.WriteFile(db.filename.String(), jsonContents, 0644)
+	if err != nil {
+		fmt.Printf("Error writing compilation database: %s", err)
+	}
+}
+
+func (db *CompilationDatabase) dirForCommand(command *exec.Cmd) string {
+	// This mimics what Cmd.Run also does: Use Dir if specified,
+	// current directory otherwise
+	if command.Dir != "" {
+		return command.Dir
+	} else {
+		dir, err := os.Getwd()
+		if err != nil {
+			fmt.Printf("Error getting current directory for compilation database: %s", err)
+			return ""
+		}
+		return dir
+	}
+}
+
+func (db *CompilationDatabase) ReplaceEntry(filename *paths.Path, command *exec.Cmd) {
+	entry := compilationCommand{
+		Directory: db.dirForCommand(command),
+		Arguments: command.Args,
+		File:      filename.String(),
+	}
+
+	db.contents = append(db.contents, entry)
+}
+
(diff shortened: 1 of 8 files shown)


cmaglie:
> Handling .ino files: Currently, only one entry is generated for .cpp file resulting from the combined and preprocessed .ino file(s). This is again fine for the temp build version, but for the sketch dir version, I think this should have a separate entry for each .ino file, with an -include=Arduino.h flag added (as suggested in #849 (comment)).

does the added `#line` directive as suggested in https://github.com/arduino/arduino-cli/pull/707 (as we do in the preprocessed *.ino -> .cpp) may solve this?

matthijskooijman:
> does the added #line directive as suggested in #707 (as we do in the preprocessed *.ino -> .cpp) may solve this?

I *think* it does not, since I think editors will just request an analysis of the file that is being edited, which does not have these directives. For those line directives to be helpful, a language server such as `clangd` would have to analyze *all* files in the compilation database (which it already does, I think), then see the `#line` directives and then somehow conclude that the flags of the preprocessed copy apply to the .ino files (which I don't think it does, or could reliably do, since #line directives can very well be the result of an `#include` too). Even more, to *really* fix the issues coming from preprocessing, clangd should be told to analyze the preprocessed copy (which I think requires changes in the editor and cannot be done through the compilation database), and then generate diagnostics for the original (rather than the actually being analyzed) file based on the `#line` directives (just tested this quickly, this is not what clangd does now).

So I'm afraid that this is just the hole we dug with this preprocessing. Anyone who is serious about using the compilation database with a third-party editor should probably use the "empty .ino file"-approach to work around this.

vivi90:
I need this `compile_commands.json` file for the code check and completion by the `ide-cpp` plugin of my Atom editor.
Without it i get the message `Unable to handle compilation, expected exactly one compiler job in ''` by clangd.

matthijskooijman:
This is superseded by the already merged #1081.

PR #949, opened 2020-09-15
Temporary: fix builds on dependent projects

cmaglie:
Golang projects that uses the CLI as-a-library will fail to build because the `replace` directive in `go.mod` is valid only for the local project. This should fix the build of `arduino-builder` until https://github.com/GeertJohan/go.rice/pull/159 is merged.

Once the upstream patch is merged this commit could be reverted and the fork removed.


arduino/security/rice-box.go
```diff
@@ -1,9 +1,10 @@
+// Code generated by rice embed-go; DO NOT EDIT.
 package security
 
 import (
 	"time"
 
-	"github.com/GeertJohan/go.rice/embedded"
+	"github.com/cmaglie/go.rice/embedded"
```
rsora, line 7: @cmaglie did you edit this file manually or is it generated?

cmaglie: generated

rsora (approved):

matthijskooijman (commented):
I looked quickly, seems ok. Didn't test yet.

PR #953, opened 2020-09-17
Add --run-post-install and --skip-post-install flags to upgrade command

silvanocerza:

      before creating one)
- [x] The PR follows

Adds two flags to an existing command.

`arduino-cli upgrade` doesn't handle post install scripts.

User can now specify  `--run-post-install` or `--skip-post-install` when running `arduino-cli upgrade` .

Nope.

* **Other information**:
None.

---


cli/upgrade/upgrade.go
```diff
@@ -24,11 +24,39 @@ import (
...
+	}
+
+	if !configuration.IsInteractive {
+		logrus.Info("Not running from console, will skip post-install by default")
+		return true
+	}
+	logrus.Info("Running from console, will run post-install by default")
+	return false
+}
+
```
cmaglie, line 59: This piece is duplicated from here: https://github.com/arduino/arduino-cli/blob/be39aa9e70e28715392d7d78b9e0d7ff757bd117/cli/core/install.go#L50-L80

Ideally, we should avoid duplicating because it's easy to forget to update both copies if we change something.
What about making the functions in `cli/core/install.go` public and use them? I mean the two functions:

`func addPostInstallFlagsToCommand(cmd *cobra.Command) {`
and
`func detectSkipPostInstallValue() bool {`

hopefully it doesn't create an import loop...

commands/instances.go
```diff
@@ -558,13 +561,27 @@ func Upgrade(ctx context.Context, req *rpc.UpgradeReq, downloadCB DownloadProgre
...
 				// In case uninstall fails tries to rollback
 				if err != nil {
+					logrus.WithError(err).Error("Error updating platform.")
 					taskCB(&rpc.TaskProgress{Message: "Error upgrading platform: " + err.Error()})
 
 					// Rollback
 					if err := pm.UninstallPlatform(latest); err != nil {
+						logrus.WithError(err).Error("Error rolling-back changes.")
 						taskCB(&rpc.TaskProgress{Message: "Error rolling-back changes: " + err.Error()})
 					}
```
cmaglie, line 571: I think we need to return here in case of rollback and not perform the postinstall

cmaglie (changes requested):

cmaglie (approved):

PR #1019, opened 2020-10-08
Added --clean flag in 'compile' command

### LINKED ISSUE #969: Allow forcing a clean build and/or cleaning build dir
### Current behavior

Currently, if an existing build dir is present, it is always used by `arduino-cli compile`

### Expected behavior

It would be nice if there was a `arduino-cli compile --clean` option that would force a clean build (similar to what happens when the compilation options change, but this should probably also prevent using a precompiled core).

Additionally, an option to clean up the build directory would be nice. At first glance `arduino-cli cache clean` might do this, but the --help output does not specify what it cleans exactly, `cache clean -v` does not show what is removed either, and a quick test suggests that it does not clean the build directory.

### Environment

- CLI version (output of `arduino-cli version`): Current git master be39aa9e70e28715392d7d78b9e0d7ff757bd117 (`arduino-cli version` did not output anything meaningful)
- OS and platform: Linux

cmaglie:
This PR adds the `--clean` flag to the `compile` command. This flag has two effects:

* force a cleanup of the build folder, so sketch and libraries and variants are recompiled
* force a rebuild of the core, even if a previously cached `core.a` is found

I've extended the `WipeoutBuildPathIfBuildOptionsChanged` action, that seemed the best place to implement this.
I've reused and refactored the already existing clean-up function.

Fix #969


matthijskooijman (approved):
LGTM, thanks!

I wonder if there should also be a clean command, so you can clean without also building (e.g. to clean up after a build and not litter temporary directories)?

cmaglie:
> I wonder if there should also be a clean command, so you can clean without also building (e.g. to clean up after a build and not litter temporary directories)?

Maybe, but I have no time to work on it right now. For the moment I'll merge this one as is and we'll see...

matthijskooijman:
> For the moment I'll merge this one as is and we'll see...

Sounds good, thanks!

PR #1028, opened 2020-10-14
[skip changelog] Split publish docs workflow

silvanocerza:

      before creating one)
- [x] The PR follows

Splits the `publish-docs.yaml` workflow in two different files to handle different branches correctly.

Pushing to `master` triggers the workflow only if commited files match the following paths:
```
      - "docs/**"
      - "docsgen/**"
      - "cli/**"
      - "rpc/**"
      - ".github/workflows/publish-docs.yaml"
```

The same happens when pushing to a branch that matches the `[0-9]+.[0-9]+.x` regex like `0.12.x`, `0.13.x`, etc.

The behavior when pushing to `master` is unchanged.
Pushing to a branch that matches the above regex always trigger the workflow.

No.

* **Other information**:
None.

---


per1234 (changes requested):
If I understand correctly, the problem is that the docs are not published when the release branch is created, but only when a push is made to the release branch that modifies a file matching the `on.push.paths` filter. So what we really need is to trigger the workflow on the release branch creation. After that, it's fine to continue with the current behavior of only publishing when pushes are made to the release branch that modify a file matching the `on.push.paths` filter

I have an alternative proposal:
https://github.com/per1234/arduino-cli/commit/6f7ecd120f15f98a5a3f2bd34ce48bb679bdc356

It uses the [`create` event](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#create) to trigger the workflow on every branch or tag creation. Unfortunately, the [`branches` filter](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestbranchestags) is not supported for this event. Also, filter patterns aren't supported in the `jobs.<job_id>.if` key. So the job will run on every branch or tag creation. The solution is to use a step with some bash code to do the filter:
```bash
BRANCH_CREATION_REGEX="refs/heads/[0-9]+.[0-9]+.x"
if [[ "${{ github.event_name }}" == "push" || ( "${{ github.event_name }}" == "create" && "${{ github.ref }}" =~ $BRANCH_CREATION_REGEX ) ]]; then
```
then use the output of that job in the conditional of the `publish` job.

It's a little messy, but I think it's slightly better than maintaining duplicate workflows.

Let me know what you think.

silvanocerza:
Uuuh yes, your solution is certainly better! Didn't think about using the `create` event for this.
Feel free to push the changes directly to this branch, or if you want I can cherry pick the change from your repo. Let me know.

Issue #532, opened 2019-12-27
using multiple linking flags when overriding a build property


artynet:
As described in the object I'd like to override the build properties of my custom **boards.txt** file leveraging all the benefits of the FreeRTOS samd21 library. For this reason, I have defined an empty build variable :
```
build.freertos.memory_wrapping_flags=
```
which is called at runtime with :

```
arduino-cli compile --fqbn <myboard> --build-properties build.freertos.memory_wrapping_flags="-Wl,--wrap=malloc -Wl,--wrap=free -Wl,--wrap=calloc -Wl,--wrap=realloc" mysketch.ino -vv
```
but just `-Wl` is set as its value so all which is located after the first comma is removed, altough enclosed within double quotes. I can see that because if I launch the command with the `--show-properties` option the output is just :
```
memory_wrapping_flags=-Wl
```
any hints about that ?

### Environment

- CLI version (output of `arduino-cli version`): 0.7.0 Commit: 3809fc3
- OS and platform: Linux x64

artynet:
@per1234 @masci 

this quick fix https://github.com/artynet/arduino-cli/commit/e8920fc2895134f9608f0c838ee45a14bb4c1523 seems to solve my problem. Would you mind taking a look at that so, in case, I can process a pull request ?

Thanks in advance

masci:
@artynet I think `StringArray` is the way to go here, so +1 from me.
This will slightly change (for the better) the way `--build-properties` works, so we should also add an example in the README file showing how to pass multiple properties at once.

artynet:
At this very point we should think of writing something like that. Given two sets of custom building flags :

- build.platform_extra_flags="flag1 flag2 flag3"
- build.platform_generic_flags"flag4 flag5 flag6"

we can override the main build properties this way :

```
--build-properties build.platform_generic_flags"flag4 flag5 flag6 flag1 flag2 flag3"
```
or
```
--build-properties build.platform_generic_flags"flag4 flag5 flag6" --build-properties build.platform_extra_flags="flag1 flag2 flag3"
```
the first being a general override while the second a separated one for single set of flags

artynet:
Hello @masci, any update on this issue ?

masci:
Ok for me to move on with a PR containing the comment linked above. Not sure I got the final example in your last comment: is that ok from a user perspective? Or is it a counter-example against changing that flag to `StringArray`? 

[...]

PR #1044, opened 2020-10-22
Deprecate compile --build-properties flag in favor of new featureful --build-property flag 

### LINKED ISSUE #1037: Can't include quoted strings in --build-properties args
I'm clearly doing something wrong, but I'm having a heck of a time figuring out how arduino-cli wants me to escape strings containing " marks in values passed to --build-properties.

A slightly cut down example:

$ export LOCAL_CFLAGS='"-I/tmp/x" "-I/tmp/y"'

$ arduino-cli compile --fqbn my:valid:fqbn --build-properties "compiler.cpp.extra_flags=$LOCAL_CFLAGS" Sketch.ino 

Gets me:

Error: invalid argument "compiler.cpp.extra_flags= \"-I/tmp/x\"  \"-I/tmp/y\"" for "--build-properties" flag: parse error on line 1, column 26: bare " in non-quoted-field

I've been through just about every permutation I can think of, but can't figure out the correct escaping to be able to pass a string containing a double-quote in as a value for --build-properties.

A search of github issues for arduino-cli doesn't seem to turn up anyone else who's run into this, which leads me to suggest that I'm just missing something really obvious.  

Currently testing against arduino-cli v0.13 and v0.11 on macOS and Linux.

I initially reported this as https://forum.arduino.cc/index.php?topic=709766.msg4769079#msg4769079

silvanocerza:

      before creating one)
- [x] The PR follows

Deprecate `--build-properties` flag in `compile` command and add a new `--build-property` flag 

The following commands would all fail because of quotes (`"`) in them.
```
export COMPILE_OPTIONS='-DMY_DEFINE="hello world"'
arduino-cli compile -b arduino:avr:uno --build-properties "build.extra_flags='$COMPILE_OPTIONS'" ./my-sketch

arduino-cli compile -b arduino:avr:uno --build-properties="build.extra_flags=\"-DMY_DEFINE=\"hello world\"\"" ./my-sketch

arduino-cli compile -b arduino:avr:uno --build-properties="build.extra_flags=\"-DMY_DEFINE=\"hello world\"\"" ./my-sketch

arduino-cli compile -b arduino:avr:uno --build-properties='compiler.cpp.extra_flags="-DPIN=2" "-DSSID=\"This is a String\"' ./my-sketch
```

The `--build-property` is introduced to substitute in the future the `--build-properties` flag, it supports the use of quotes but has a different syntax.
Different build flags can't be specified using the same `--build-property` flag, nor you can specify the same build flag on separate `--build-property` otherwise they'll get overridden.

These are some examples.
```
  ./arduino-cli compile -b arduino:avr:uno /home/user/Arduino/MySketch
  ./arduino-cli compile -b arduino:avr:uno --build-property="build.extra_flags=\"-DMY_DEFINE=\"hello world\"\"" /home/user/Arduino/MySketch
  ./arduino-cli compile -b arduino:avr:uno --build-property="build.extra_flags=-DPIN=2 \"-DMY_DEFINE=\"hello world\"\"" /home/user/Arduino/MySketch
  ./arduino-cli compile -b arduino:avr:uno --build-property=build.extra_flags=-DPIN=2 --build-property="compiler.cpp.extra_flags=\"-DMY_DEFINE=\"hello world\"\"" /home/user/Arduino/MySketch
```

No.

* **Other information**:
Fixes #1037, #210 and probably also #532 and #846.

---

### DIFF
--- a/cli/compile/compile.go
+++ b/cli/compile/compile.go
@@ -62,9 +62,9 @@ func NewCommand() *cobra.Command {
 		Long:  "Compiles Arduino sketches.",
 		Example: "" +
 			"  " + os.Args[0] + " compile -b arduino:avr:uno /home/user/Arduino/MySketch\n" +
-			"  " + os.Args[0] + " compile -b arduino:avr:uno --build-property='build.extra_flags=\"-DMY_DEFINE=\"hello world\"\"' /home/user/Arduino/MySketch\n" +
-			"  " + os.Args[0] + " compile -b arduino:avr:uno --build-property='build.extra_flags=-DPIN=2 \"-DMY_DEFINE=\"hello world\"\"' /home/user/Arduino/MySketch\n" +
-			"  " + os.Args[0] + " compile -b arduino:avr:uno --build-property=build.extra_flags=-DPIN=2 --build-property='compiler.cpp.extra_flags=\"-DSSID=\"hello world\"\"'-DMY_DEFINE=\"hello world\"' /home/user/Arduino/MySketch\n",
+			"  " + os.Args[0] + " compile -b arduino:avr:uno --build-property=\"build.extra_flags=\\\"-DMY_DEFINE=\\\"hello world\\\"\\\"\" /home/user/Arduino/MySketch\n" +
+			"  " + os.Args[0] + " compile -b arduino:avr:uno --build-property=\"build.extra_flags=-DPIN=2 \\\"-DMY_DEFINE=\\\"hello world\\\"\\\"\" /home/user/Arduino/MySketch\n" +
+			"  " + os.Args[0] + " compile -b arduino:avr:uno --build-property=build.extra_flags=-DPIN=2 --build-property=\"compiler.cpp.extra_flags=\\\"-DSSID=\\\"hello world\\\"\\\"\"-DMY_DEFINE=\"hello world\"' /home/user/Arduino/MySketch\n",
 		Args: cobra.MaximumNArgs(1),
 		Run:  run,
 	}
--- a/cli/compile/compile.go
+++ b/cli/compile/compile.go
@@ -62,9 +62,9 @@ func NewCommand() *cobra.Command {
 		Long:  "Compiles Arduino sketches.",
 		Example: "" +
 			"  " + os.Args[0] + " compile -b arduino:avr:uno /home/user/Arduino/MySketch\n" +
-			"  " + os.Args[0] + " compile -b arduino:avr:uno --build-property=\"build.extra_flags=\\\"-DMY_DEFINE=\\\"hello world\\\"\\\"\" /home/user/Arduino/MySketch\n" +
-			"  " + os.Args[0] + " compile -b arduino:avr:uno --build-property=\"build.extra_flags=-DPIN=2 \\\"-DMY_DEFINE=\\\"hello world\\\"\\\"\" /home/user/Arduino/MySketch\n" +
-			"  " + os.Args[0] + " compile -b arduino:avr:uno --build-property=build.extra_flags=-DPIN=2 --build-property=\"compiler.cpp.extra_flags=\\\"-DSSID=\\\"hello world\\\"\\\"\"-DMY_DEFINE=\"hello world\"' /home/user/Arduino/MySketch\n",
+			"  " + os.Args[0] + ` compile -b arduino:avr:uno --build-property "build.extra_flags=\"-DMY_DEFINE=\"hello world\"\"" /home/user/Arduino/MySketch` + "\n" +
+			"  " + os.Args[0] + ` compile -b arduino:avr:uno --build-property "build.extra_flags=-DPIN=2 \"-DMY_DEFINE=\"hello world\"\"" /home/user/Arduino/MySketch` + "\n" +
+			"  " + os.Args[0] + ` compile -b arduino:avr:uno --build-property build.extra_flags=-DPIN=2 --build-property "compiler.cpp.extra_flags=\"-DSSID=\"hello world\"\"" /home/user/Arduino/MySketch` + "\n",
 		Args: cobra.MaximumNArgs(1),
 		Run:  run,
 	}
--- a/cli/compile/compile.go
+++ b/cli/compile/compile.go
@@ -57,12 +57,16 @@ var (
 // NewCommand created a new `compile` command
 func NewCommand() *cobra.Command {
 	command := &cobra.Command{
-		Use:     "compile",
-		Short:   "Compiles Arduino sketches.",
-		Long:    "Compiles Arduino sketches.",
-		Example: "  " + os.Args[0] + " compile -b arduino:avr:uno /home/user/Arduino/MySketch",
-		Args:    cobra.MaximumNArgs(1),
-		Run:     run,
+		Use:   "compile",
+		Short: "Compiles Arduino sketches.",
+		Long:  "Compiles Arduino sketches.",
+		Example: "" +
+			"  " + os.Args[0] + " compile -b arduino:avr:uno /home/user/Arduino/MySketch\n" +
+			"  " + os.Args[0] + " compile -b arduino:avr:uno --build-property='build.extra_flags=\"-DMY_DEFINE=\"hello world\"\"' /home/user/Arduino/MySketch\n" +
+			"  " + os.Args[0] + " compile -b arduino:avr:uno --build-property='build.extra_flags=-DPIN=2 \"-DMY_DEFINE=\"hello world\"\"' /home/user/Arduino/MySketch\n" +
+			"  " + os.Args[0] + " compile -b arduino:avr:uno --build-property=build.extra_flags=-DPIN=2 --build-property='compiler.cpp.extra_flags=\"-DSSID=\"hello world\"\"'-DMY_DEFINE=\"hello world\"' /home/user/Arduino/MySketch\n",
+		Args: cobra.MaximumNArgs(1),
+		Run:  run,
 	}
 
 	command.Flags().StringVarP(&fqbn, "fqbn", "b", "", "Fully Qualified Board Name, e.g.: arduino:avr:uno")
@@ -72,8 +76,10 @@ func NewCommand() *cobra.Command {
 	command.Flags().StringVarP(&exportDir, "output-dir", "", "", "Save build artifacts in this directory.")
 	command.Flags().StringVar(&buildPath, "build-path", "",
 		"Path where to save compiled files. If omitted, a directory will be created in the default temporary path of your OS.")
-	command.Flags().StringArrayVar(&buildProperties, "build-properties", []string{},
-		"List of custom build properties separated by spaces. Or can be used multiple times for multiple properties.")
+	command.Flags().StringSliceVar(&buildProperties, "build-properties", []string{},
+		"List of custom build properties separated by commas. Or can be used multiple times for multiple properties.")
+	command.Flags().StringArrayVar(&buildProperties, "build-property", []string{},
+		"Override a build property with a custom value. Can be used multiple times for multiple properties.")
 	command.Flags().StringVar(&warnings, "warnings", "none",
 		`Optional, can be "none", "default", "more" and "all". Defaults to "none". Used to tell gcc which warning level to use (-W flag).`)
 	command.Flags().BoolVarP(&verbose, "verbose", "v", false, "Optional, turns on verbose mode.")
@@ -94,6 +100,8 @@ func NewCommand() *cobra.Command {
 	command.Flags().BoolP("export-binaries", "e", false, "If set built binaries will be exported to the sketch folder.")
 
 	configuration.Settings.BindPFlag("sketch.always_export_binaries", command.Flags().Lookup("export-binaries"))
+
+	command.Flags().MarkDeprecated("build-properties", "please use --build-property instead.")
 
 	return command
 }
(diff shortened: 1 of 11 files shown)


legacy/builder/types/context.go
```diff
@@ -187,22 +187,6 @@ func (ctx *Context) ExtractBuildOptions() *properties.Map {
...
-	fqbn, err := cores.ParseFQBN(opts.Get("fqbn"))
-	if err != nil {
-		i18n.ErrorfWithLogger(ctx.GetLogger(), "Error in FQBN: %s", err)
-	}
-	ctx.FQBN = fqbn
-	ctx.ArduinoAPIVersion = opts.Get("runtime.ide.version")
-	ctx.CustomBuildProperties = strings.Split(opts.Get("customBuildProperties"), ",")
-	ctx.OptimizationFlags = opts.Get("compiler.optimization_flags")
-}
-
```
cmaglie, line 205: Why did you remove this one? Is this function unused?

silvanocerza: Yep, completely unused. I was digging around trying to understand how to fix the issue and stumbled upon it, noticed it wasn't used and decided to delete since I was there.

cli/compile/compile.go
```diff
@@ -73,8 +73,8 @@ func NewCommand() *cobra.Command {
 	command.Flags().BoolVarP(&dryRun, "dry-run", "n", false, "Perform the build but do not copy the compile output file.")
 	command.Flags().StringVar(&buildPath, "build-path", "",
 		"Path where to save compiled files. If omitted, a directory will be created in the default temporary path of your OS.")
-	command.Flags().StringSliceVar(&buildProperties, "build-properties", []string{},
-		"List of custom build properties separated by commas. Or can be used multiple times for multiple properties.")
+	command.Flags().StringArrayVar(&buildProperties, "build-properties", []string{},
+		"List of custom build properties separated by spaces. Or can be used multiple times for multiple properties.")
```
cmaglie, line 77: Since if you want to change multiple properties you now need to repeat the flag multiple times I think that the correct description is something along the line of:

```suggestion
		"Override a build property with a custom value. Can be used multiple times for multiple properties.")
```

legacy/builder/types/context.go
cmaglie: I checked, it's used in the `arduino-builder`, so we need it: https://github.com/arduino/arduino-builder/blob/642e2cd708dd024e658067d383387691f7e79abd/main.go#L244

silvanocerza:
I've updated the PR with some changes, I marked the `--build-properties` flag as deprecated in favor of a new `--build-property` flag that works a bit differently but supports quotes in defines and similar things.

ubidefeo:
This took a while to test, had to look hard for a way to break it but haven't found it yet.
it works as far as I'm concerned

silvanocerza:
> This took a while to test, had to look hard for a way to break it but haven't found it yet.
> it works as far as I'm concerned

This is a first people! Marking it on the calendar! :calendar: 

I'll be waiting for a review by @cmaglie anyway before merging. :smile:

silvanocerza:
Also would love some feedback from @obra, @bxparks and @artynet. 
You can find a testable build on here: https://github.com/arduino/arduino-cli/pull/1044/checks?check_run_id=1310376266
There's an `Artifacts` dropdown menu on the top right with download links for different platforms.

obra:
I’ll do my best to give it a workout today. Does arduino-cli have a written policy about what “deprecation” means?

Some projects, for example, have a policy of deprecations warning for two releases before failing with a message that the facility was removed for another n releases. Some projects explicitly state that every .0 is fair game for breakage.

silvanocerza:
> I’ll do my best to give it a workout today. Does arduino-cli have a written policy about what “deprecation” means?

Thanks, appreciate it. :)

> Some projects, for example, have a policy of deprecations warning for two releases before failing with a message that the facility was removed for another n releases. Some projects explicitly state that every .0 is fair game for breakage.

We're working on it, we're exploring different alternatives.

obra:
> > Some projects, for example, have a policy of deprecations warning for two releases before failing with a message that the facility was removed for another n releases. Some projects explicitly state that every .0 is fair game for breakage.
> 
> We're working on it, we're exploring different alternatives.

I can respect that answer :) My background comes from managing Perl's backwards-compatibility policy, which is, roughly "do not break on a user unless it is unavoidable." Being that focused on not hurting users comes with a huge maintenance burden that can calcify a project if such a policy is applied to early.

Is there a place that it would be useful for me to comment about what I as a consumer of the tool (and as someone who's been responsible for writing such policies) might want? It's 100% ok if the answer is "feedback is not useful right now"

rsora:
@obra 

> I can respect that answer :) My background comes from managing Perl's backwards-compatibility policy, which is, roughly "do not break on a user unless it is unavoidable." Being that focused on not hurting users comes with a huge maintenance burden that can calcify a project if such a policy is applied to early.

Exactly this :) Being the CLI still in alpha and in very heavy development, we are running fast to stabilize the APIs and implement all the features we think are needed. We are trying to avoid to enforce a structured Deprecation policy ATM to not slow down the development.

However we recognize that the project gained traction, and we have users that rely on the CLI in their projects. We don't want to let our users down :)  so I started to modify slightly our contribution guidelines here https://github.com/arduino/arduino-cli/pull/1039 and our internal release documents in order to start managing the project breaking changes in the easiest way possible.

It would be great to get your feedback on the PR :+1:

rsora:
Let me clarify that having a structured deprecation policy is a milestone in our roadmap for the `1.0.0` stable release, so it is something we are working on. 

Maybe my previous message was a bit misleading on this specific point. :smile:

alranel:
@obra, I'm so happy to see you again in this repository :) 
Your feedback is _very_ welcome!

obra:
@rsora Is the internal release document something that can't be made public?  

Poking at 0.13.. there's nowhere I see in the help or the online docs that arduino-cli is an alpha release. Ditto https://arduino.github.io/arduino-cli/latest

Now, most advanced users can probably figure out that 0.13 isn't a "release product" version number. But it looks like older alphas had 'alpha' in the filename or version.

In terms of something "lightweight" for deprecation policy before 1.0, maybe something like this would be enough?

"arduino-cli is a powerful tool that many people depend on, but it is still pre-release software. We are still hard at work making sure that arduino-cli 1.0 will be a tool that you can trust long into the future. Because of this, we may make backwards-incompatible changes at any time, but promise to do so only after careful deliberation."

cli/compile/compile.go
```diff
@@ -57,12 +57,16 @@ var (
...
-		Args:    cobra.MaximumNArgs(1),
-		Run:     run,
+		Use:   "compile",
+		Short: "Compiles Arduino sketches.",
+		Long:  "Compiles Arduino sketches.",
+		Example: "" +
+			"  " + os.Args[0] + " compile -b arduino:avr:uno /home/user/Arduino/MySketch\n" +
+			"  " + os.Args[0] + " compile -b arduino:avr:uno --build-property=\"build.extra_flags=\\\"-DMY_DEFINE=\\\"hello world\\\"\\\"\" /home/user/Arduino/MySketch\n" +
+			"  " + os.Args[0] + " compile -b arduino:avr:uno --build-property=\"build.extra_flags=-DPIN=2 \\\"-DMY_DEFINE=\\\"hello world\\\"\\\"\" /home/user/Arduino/MySketch\n" +
+			"  " + os.Args[0] + " compile -b arduino:avr:uno --build-property=build.extra_flags=-DPIN=2 --build-property=\"compiler.cpp.extra_flags=\\\"-DSSID=\\\"hello world\\\"\\\"\"-DMY_DEFINE=\"hello world\"' /home/user/Arduino/MySketch\n",
```
cmaglie, line 67: are you sure that this last example is correct? on the help is displayed as:

```
  ./arduino-cli compile -b arduino:avr:uno --build-property=build.extra_flags=-DPIN=2 --build-property="compiler.cpp.extra_flags=\"-DSSID=\"hello world\"\""-DMY_DEFINE="hello world"' /home/user/Arduino/MySketch
```

and it doesn't look right to me...
To simplify your life you may use the golang verbatim string operator (backtick) so the help strings will not require triple quoting:

```
  `--build-property="build.extra_flags=\"-DMY_DEFINE=\"hello world\"\"" /home/user/Arduino/MySketch`
```

another small improvement may be to drop the (optional) `=` sign after `--build-property` so instead of:

  `--build-property="build.extra_flags=\"-DMY_DEFINE=\"hello world\"\""`

we may write

  `--build-property "build.extra_flags=\"-DMY_DEFINE=\"hello world\"\""`

that is slightly clearer IMHO.

cmaglie (changes requested):
Added a note to check the examples on the help

For the rest LGTM

cli/compile/compile.go
silvanocerza: Right! The last example is plainly wrong, I'll fix it.

Somehow I always forget about the backtick. :facepalm:

bxparks:
Hello, thank you for working on this issue. This is the main blocker that prevents me from using the ArduinoCLI.

Is there a reference or documentation that explains the tokenization and parsing rules for the expression on the right side of the `compiler.cpp.extra_flags=`? I would have assumed that the entire string to the right of the `=` would be passed without any processing to the compiler binary. But after some trial and error, there seems something is doing some processing of double-quotes and white spaces, but I cannot deduce the rules even after many attempts of trial and error.

1) The original 4th example in the `compile --help` output contained 2 macros, `SSID` and `MY_DEFINE`. There were at least 2 problems with it. One, an unmatched single-quote, which you fixed. Two, a missing space before the second `-D`.  If I attempt to get the original example to work, the following does **not** work (I'm using the `bash(1)` where single-quoted strings are passed unprocessed to the `arduino-cli` binary):

```
$ arduino-cli compile --warnings all --verbose -b arduino:avr:uno \
    --build-property 'compiler.cpp.extra_flags=-D SSID="hello world" -D MY_DEFINE="hello world"'
[...]
.../avr-g++ [...] -D "SSID=\"hello" "world\"" -D "MY_DEFINE=\"hello" "world\"" [...]
avr-g++: error: world": No such file or directory
avr-g++: error: world": No such file or directory
```

2) It looks like the the parser is tokenizing the whitespace **within** the quoted `"hello world"`. Because if we remove the white space within the quoted string, the parsing is correct:

```
$ arduino-cli compile --warnings all --verbose -b arduino:avr:uno \
    --build-property 'compiler.cpp.extra_flags=-D SSID="helloworld" -D MY_DEFINE="helloworld"'
[...]
.../avr-g++ [...] -D "SSID=\"helloworld\"" -D "MY_DEFINE=\"helloworld\"" [...]
```

3) The only way I have found to get around this tokenization problem is unescaped, nested, double-quote (which should **not** work, but it does):

```
$ arduino-cli compile --warnings all --verbose -b arduino:avr:uno \
    --build-property 'compiler.cpp.extra_flags="-D SSID="hello world"" "-D MY_DEFINE="hello world""'
[...]
.../avr-g++ [...] "-D SSID=\"hello world\"" "-D MY_DEFINE=\"hello world\"" [...]
```
4) If I attempt to double-quote the entire string on the right side of the `compiler.cpp.extra_flags=`, the parsing fails:

```
$ arduino-cli compile --warnings all --verbose -b arduino:avr:uno \
    --build-property 'compiler.cpp.extra_flags="-D SSID="hello world" -D MY_DEFINE="hello world""'
[...]
.../avr-g++ [...] "-D SSID=\"hello world" -D "MY_DEFINE=\"hello" "world\"\"" [...]
avr-g++: error: world"": No such file or directory
```

It seems like nested-double-quoting only works around a single `-D` flag.

5) However, the nested double-quote escape breaks down when I try to embed a naked double-quote character inside the `SSID`, because the parser does not seem to know to skip over the backslashed escaped double-quote. Let's say I want the SSID to be the raw string `hello" world`. The following does not work:

```
$ arduino-cli compile --warnings all --verbose -b arduino:avr:uno \
    --build-property 'compiler.cpp.extra_flags="-D SSID="hello\" world"" "-D MY_DEFINE="hello\" world""'
[...]
.../avr-g++ [...] "-D SSID=\"hello\\" "world\"\"" "-D MY_DEFINE=\"hello\\" "world\"\"" [...]
avr-g++: error: world"": No such file or directory
avr-g++: error: world"": No such file or directory
```

**Remarks**:

The `--build-property` flag is definitely better than the previous `--build-properties` flag. But it seems like the tokenizing and parsing of the `-D` flags is still not fully accurate. In particular, it does not seem to support string macros with embedded spaces or embedded double-quotes. I think a lot of these problems are caused by multiple levels of tokenizing, and double-quote parsing. If you can clarify and document what those parsing rules are, I think that would help a lot.

I am still of the opinion that exposing the `-D` flag directly from the arduino-cli go binary is the best way to implement this feature. This would remove a number of confusing layers of parsing, tokenizing and quote and backslash escaping.

Another comment: Some of the `compile --help` examples show the `build.extra_flags=` property being overridden. But isn't the `build.extra_flags` property a dangerous one to override? All of the various 3rd party cores that I've looked at defines a non-empty value for the `build.extra_flags=` property.  If the end-user overrode this, it is not clear that things would work properly. The `compiler.cpp.extra_flags=` seems to be the property that is unused by various 3rd party cores, so can be safely overriden by end-users.

bxparks:
By the way, I used the following `Test.ino` file to test my `arduino-cli` commands above:

```c++
const char STRING[3] = SSID; // will trigger a compiler warning if SSID is > 2 characters

#ifdef MY_DEFINE
  #warning MY_DEFINE
#endif

void setup() { }

void loop() { }
```

silvanocerza:
Force pushed to resolve conflicts, merging after checks are sucessful.

PR #1048, opened 2020-10-28
Add gRPC interface for board list watcher

silvanocerza:

      before creating one)
- [x] The PR follows

Add a gRPC interface for the board list watch command.

gRPC consumers must poll the daemon to know which boards are connected or disconnected.

gRPC consumers now receive an event whenever a board is connected or disconnected.

No.

* **Other information**:

None.

---


rpc/commands/board.proto
```diff
@@ -206,6 +206,22 @@ message BoardListAllResp {
...
+    bool interrupt = 2;
+}
+
+message BoardListWatchResp {
+    // Event type as received from the serial discovery tool
+    string event_type = 1;
+    // Information about the port
+    DetectedPort port = 2;
+    // Eventual errors when detecting connected boards
+    string error = 3;
```
kittaakos, line 222: I am not a `proto` expert and I hope @cmaglie can help, but I would have expected to have something like:
```proto
message BoardListWatchResp {
  oneof content {
    DetectedChange change = 1;
    string error = 2;
  }
}

message DetectedChange {
  enum ChangeType { ADDED = 0; REMOVED = 0; }
  ChangeType type = 1;
  DetectedPort port = 2;
}
```

kittaakos (commented):
I made a few proposals. No hurry with the change, I continue with the integration based on what we have now as it also works.

kittaakos:
I have a couple of questions regarding the behavior;
 1. When I attach a board, I receive a response with `event-type` `add`. This response has the port (with address and protocol) and an array of boards (can be empty, for sure) but it is usually one. However, when I attach my Portenta board, I see one port with three boards. What does it mean domain-wise?
```json
{
  "type": "add",
  "address": "/dev/cu.usbmodem14101",
  "protocol": "serial",
  "protocol_label": "Serial Port (USB)",
  "boards": [
    {
      "name": "Arduino Portenta H7 (M7 core)",
      "FQBN": "arduino-beta-development:mbed:envie_m7"
    },
    {
      "name": "Arduino Portenta H7 (ThreadDebug)",
      "FQBN": "arduino-beta-development:mbed:envie_m7_thread_debug"
    },
    {
      "name": "Arduino Portenta H7 (M7 core)",
      "FQBN": "arduino-beta:mbed:envie_m7"
    }
  ]
}
```

 2. When I detach a board, I receive a response with `even_type` `remove` and it has only the address of the port.
```json
{
  "type": "remove",
  "address": "/dev/cu.usbmodem14101"
}
```
Does it mean that consumers of the gRPC API should move away from a board-driven approach, and track ports instead? In other words, from now on boards are always kind of an additional "meta" information of the detected port, and the ports are the first-class citizens of the CLI.

Thank you!

silvanocerza:
You see 3 boards because by searching in the locally installed platforms we find mutiple matches for the VID/PID combination probably.

If it's better for you I could change the message sent so that the list of possible connected boards to that port is included, I'd have to keep track of them in the CLI though.

rpc/commands/board.proto
```diff
@@ -206,6 +206,22 @@ message BoardListAllResp {
...
+message BoardListWatchReq {
+    // Arduino Core Service instance from the `Init` response.
+    Instance instance = 1;
+    // Set this to true to stop the discovery process
+    bool interrupt = 2;
+}
+
+message BoardListWatchResp {
+    // Event type as received from the serial discovery tool
+    string event_type = 1;
```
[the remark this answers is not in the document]
silvanocerza: Yeah, didn't think about it.

rpc/commands/board.proto
silvanocerza: Neither am I and that's why didn't about this approach, I like it though.

cmaglie: I'm positive to move the pair (ChangeType, DetectedPort) into their own separate message.
I would change the naming, since we are now talking about `Ports` IMHO it should be:

```
message BoardListWatchResp {
  oneof content {
    DetectedPortEvent event = 1;
    string error = 2;
  }
}

message DetectedPortEvent {
  enum Event { ADDED = 0; REMOVED = 0; }
  Event action = 1;
  DetectedPort port = 2;
}
```
(and at this point also we should probably go all-in and call it Port instead of DetectedPort, this is something that we must face once Pluggable Discovery is merged...)

I'm not sure about the `oneof`, there are other places where it may be useful to make the API more explicit, anyway when I tried it, IIRC, the generated API turned out to be very messy especially in languages without sum types like go, java, c++ where accessors (boilerplate) methods are added to handle it correctly. Let's evaluate if the pro (a cleaner `proto` API) is more valuable than the cons (a bigger "cognitive" load in the implementation).

kittaakos: Thank you for your recommendations, @cmaglie. As I said, I am fine with this API now. It's your call whether you want to make the changes now or later.

silvanocerza:
Suggested changes to the gRPC interface will be made in a future PR, this is good enough for now.

PR #1064, opened 2020-11-12
Some small refactoring on legacy package

cmaglie:
These are smaller/medium changes that are preparatory for other bigger PR.

I found myself juggling those small patches here and there in other branches, so I've decided to bite the bullet and put them in a smaller PR and merge them once for all.

Each commit has an self-contained change, they should be fairly easy to review.

I've another batch of changes that sits on top of this PR.


legacy/builder/utils/utils.go
```diff
@@ -189,22 +148,11 @@ func TrimSpace(value string) string {
 }
 
 func PrepareCommand(pattern string) (*exec.Cmd, error) {
-	parts, err := ParseCommandLine(pattern)
+	parts, err := properties.SplitQuotedString(pattern, `"'`, false)
```
silvanocerza, line 151: Would this change the behaviour of the `compile` command `--build-property` and `--build-properties` flags?

legacy/builder/create_cmake_rule.go
```diff
@@ -241,7 +241,7 @@ func canExportCmakeProject(ctx *types.Context) bool {
 }
 
 func extractCompileFlags(ctx *types.Context, receipe string, defines, dynamicLibs, linkerflags, linkDirectories *[]string, logger i18n.Logger) {
-	command, _ := builder_utils.PrepareCommandForRecipe(ctx, ctx.BuildProperties, receipe, true)
+	command, _ := builder_utils.PrepareCommandForRecipe(ctx.BuildProperties, receipe, true)
```
silvanocerza, line 244: There's an extra `e` in `recipe`, might as well change it since we're at it. :ok_person:

legacy/builder/utils/utils.go
cmaglie: No it should work exactly the same. This is code duplication that has never been removed.
(BTW a quick test would not be bad :-))

legacy/builder/create_cmake_rule.go
cmaglie: done

silvanocerza (approved):

matthijskooijman (commented):
A bit late, but I had a look over this PR and the changes look good to me. I'll also look at part 2 next :-)

PR #1075, opened 2020-11-16
lib install --git-url and --zip-file must now be esplicitly enabled

silvanocerza:

      before creating one)
- [x] The PR follows

Changes an existing command flags.

`--git-url` and `--zip-file` flags can be freely used in `lib install`.

`--git-url` and `--zip-file` flags are now hidden by default and must be explicitly enabled by setting `library.enable_unsafe_install` to `true` in `arduino-cli.yaml` or by using the env var `ARDUINO_ENABLE_UNSAFE_LIBRARY_INSTALL`. 

Yes, but I suppose it won't be an issue since the feature has been added recently and released only on nightlies.

* **Other information**:

None.

---

### DIFF
--- a/cli/lib/install.go
+++ b/cli/lib/install.go
@@ -24,6 +24,7 @@ import (
 	"github.com/arduino/arduino-cli/cli/instance"
 	"github.com/arduino/arduino-cli/cli/output"
 	"github.com/arduino/arduino-cli/commands/lib"
+	"github.com/arduino/arduino-cli/configuration"
 	rpc "github.com/arduino/arduino-cli/rpc/commands"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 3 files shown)


per1234 (approved):
I gave it a try and it works just as I expect it to.

matthijskooijman:
Hm, I'm curious about this change, how are these options unsafe, really? They can cause random stuff to be downloaded, but the IDE does not actually *execute* anything from a library (unlike platforms), so the possibility for unsafety seems limited?

Also, this implies that regular library installs from the library manager are somehow more safe, but is this really the case? There is more potential for verifying the safety of libraries in the manager, but in practice I think there is only some limited manual inspection of library metadata correctness, but I doubt that @cmaglie fully inspects the contents of each library when adding it to the library, and I expect no manual inspection happens on future updates to the library.

Or is there another kind of safety meant here?

ubidefeo:
@matthijskooijman 
we're just taking a bit of distance from what people install as libraries that might end up in their secure IoT applications.
In the future as we work towards a centralised library repo with different levels of machine/human vetting we'd like to be able to recommend libraries that have a higher ranking based on solid information provided, maintainers being physical entities who actually take responsibility for their code, provide verification of examples based on architecture compatibility and so on.
It's a long-term plan, but we'll have to start somewhere :)

This option allows users to have the flexibility to install whatever they wish, and it'll be at some point transferred to the new IDE as a UI configuration option.
Right now we're addressing the need to install things in zip files and git repos, which IMO is really useful, and adding this configuration property makes sure that you can't just point to a URL and download whatever unless you know what you're doing.
"Overzealous" might be the word replacing "safe" for now :D

matthijskooijman:
Thanks, sounds like a reasonable approach!

PR #1079, opened 2020-11-18
Some small refactoring on legacy package (part 2)

cmaglie:
This is another refactoring of the `legacy` package.

As for #1064:

> These are smaller/medium changes that are preparatory for other bigger PR.
> 
> I found myself juggling those small patches here and there in other branches, so I've decided to bite the bullet and put them in a smaller PR and merge them once for all.
>
> Each commit has a self-contained change, they should be fairly easy to review.


silvanocerza (approved):

matthijskooijman (commented):
Looks good!

PR #1081, opened 2020-11-18
Add support for compile_command.json output

cmaglie:
This implementation is a tentative draft based on @matthijskooijman's #944

I've added a flag `--only-compilation-database` flag in `compile` command. When this flag is specified the build run will perform  only:
- library discovery
- sketch preprocessing (the preprocessed `sketch.ino.cpp` is saved in the usual `${build_path}/sketch/...` subdir)
- generation of `compile_command.json` (saved in the build-path). In this case the sketch/core/libraries are not compiled but the command line is saved for `compile_command.json`


silvanocerza (approved):

matthijskooijman:
I finally gotten around to giving this code a whirl, seems to work nicely overall. There's still some points unsolved from #944, but let's further discuss them in #849.

Why did you removing the "Writing compilation database..." message? IMHO it helps to both let people know that this feature is supported in the first place, and helps figure out where the file was written to, especially now it's only written to the build folder. In verbose mode, all commands run are printed, then why not show the compilation database was generated?

cmaglie:
> Why did you removing the "Writing compilation database..." message?

Mainly because:
* it was inside a `CompilationDatabase` class, from the caller point of view calling `SaveToFile` should just save or return an error, without any other output
* it used `fmt.Printf` instead of the logging system

BTW, you have a good point, I'll put a message output on the caller side.

PR #1082, opened 2020-11-20
Add new commands to edit config files

silvanocerza:
**Please check if the PR fulfills these requirements**

- [x] The PR has no duplicates (please search among the [Pull Requests](https://github.com/arduino/arduino-cli/pulls)
      before creating one)
- [x] The PR follows
      [our contributing guidelines](https://arduino.github.io/arduino-cli/latest/CONTRIBUTING/#pull-requests)
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

* **What kind of change does this PR introduce?**

New feature!

- **What is the current behavior?**

To change your config file you must edit it manually.

* **What is the new behavior?**

There are 4 new sub command to `config`:

* `add` to add one or more values to a setting key that contains a list
* `remove` to remove one or more values from a existing key that contains a list
* `set` to change the value of a setting key
* `delete` to delete a key and all its sub keys

All these commands accept only the CLI config values, you can't set your own settings with these.

- **Does this PR introduce a breaking change?**

Nope.

* **Other information**:

None.

---

See [how to contribute](https://arduino.github.io/arduino-cli/latest/CONTRIBUTING/)

PR #1093, opened 2020-12-01
Provide package indexes from file paths using "additional-paths" flag

giulcioffi:

      before creating one)
- [x] The PR follows

It introduces a new flag `additional-paths` that allows to provide package indexes from absolute file paths. In this case, the signature will not be checked. This can be useful for third party cores, since they will not be requested to upload their package indexes to a server. However it can also be useful for Arduino internal usage during testing phases.

Package indexes can be downloaded only from URL. In case of Arduino cores, the signature is always checked and if a valid signature is not found, the core installation fails. There is no way to test a new Arduino package index that has no signature yet.

Package indexes can be provided also from the absolute file path and the signature will not be checked. An Arduino package index without a signature can be provided in this way.

No, it does not. It introduces a new feature that should not impact on the already implemented ones.

* **Other information**:
This PR will be needed for the Arduino core release automation workflow. Tests on the core with arduino-cli will need to run when there is no signature yet.

---

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -104,6 +104,7 @@ func createCliCommandTree(cmd *cobra.Command) {
 	cmd.PersistentFlags().StringVar(&outputFormat, "format", "text", "The output format, can be {text|json}.")
 	cmd.PersistentFlags().StringVar(&configFile, "config-file", "", "The custom config file (if not specified the default will be used).")
 	cmd.PersistentFlags().StringSlice("additional-urls", []string{}, "Comma-separated list of additional URLs for the Boards Manager.")
+	cmd.PersistentFlags().StringSlice("additional-paths", []string{}, "Comma-separated list of additional file paths for the Boards Manager.")
 	configuration.BindFlags(cmd, configuration.Settings)
 }
 
--- a/commands/instances.go
+++ b/commands/instances.go
@@ -200,9 +200,30 @@ func UpdateIndex(ctx context.Context, req *rpc.UpdateIndexReq, downloadCB Downlo
 	}
 
 	indexpath := paths.New(configuration.Settings.GetString("directories.Data"))
+	json_paths := []string{}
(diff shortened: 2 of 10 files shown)


cli/cli.go
```diff
@@ -104,6 +104,7 @@ func createCliCommandTree(cmd *cobra.Command) {
 	cmd.PersistentFlags().StringVar(&outputFormat, "format", "text", "The output format, can be {text|json}.")
 	cmd.PersistentFlags().StringVar(&configFile, "config-file", "", "The custom config file (if not specified the default will be used).")
 	cmd.PersistentFlags().StringSlice("additional-urls", []string{}, "Comma-separated list of additional URLs for the Boards Manager.")
+	cmd.PersistentFlags().StringSlice("additional-paths", []string{}, "Comma-separated list of additional file paths for the Boards Manager.")
```
silvanocerza, line 107: You need to add this new key also in `cli/config/validate.go`, otherwise you won't be able to edit from the CLI.
See #1082

commands/instances.go
```diff
@@ -200,9 +201,31 @@ func UpdateIndex(ctx context.Context, req *rpc.UpdateIndexReq, downloadCB Downlo
 	}
 
 	indexpath := paths.New(configuration.Settings.GetString("directories.Data"))
+	jsonPaths := []string{}
+	jsonPaths = append(jsonPaths, configuration.Settings.GetStringSlice("board_manager.additional_paths")...)
+
+	for _, x := range jsonPaths {
```
silvanocerza, line 207: You're not using `jsonPaths` anywhere else other than the `for`, you could do this to simplify:
```
for _, x := range configuration.Settings.GetStringSlice("board_manager.additional_paths") {
```

Same thing for the other in `createInstance()`.

silvanocerza (changes requested):
There are some small changes to make but overall the PR is good.

silvanocerza (approved):
LGTM :+1:

PR #1117, opened 2020-12-28
ARDUIFINE: global compiler options extracted from ctags parsing

d-a-v:
- [X] The PR has no duplicates (please search among the [Pull Requests](https://github.com/arduino/arduino-cli/pulls) before creating one)

----------------

Example: these lines in the `.ino` file:
```
    const char* ARDUIFINE_abc = "MYLIB_BUFFER_LEN=1234";
    const char* ARDUIFINE_def = "MYLIB_PARAM";
    const char* ARDUINOGLOBAL_ghi = "-DMYLIB2_BUFFER_LEN=1234";
    const char* ARDUINOGLOBAL_jkl = "-DMYLIB2_PARAM";
```
will globally enable the following options in the compiler command line:
```
        -DMYLIB_BUFFER_LEN=1234 -DMYLIB_PARAM -DMYLIB2_BUFFER_LEN=1234 -DMYLIB2_PARAM
```

That, ***including during compilation of the core and libraries***.

If they are not otherwise used by the sketch, the symbols `ARDUIFINExxx` and `ARDUINOGLOBALyyy` are excluded from the final binary at link time.

This PR aims at giving a solution to https://github.com/arduino/Arduino/issues/421
and is maybe related to #846.

---------

arduifine
(the missing `M` is fixed)

### DIFF
--- a/docs/sketch-build-process.md
+++ b/docs/sketch-build-process.md
@@ -27,6 +27,9 @@ compiler:
   rare cases, prototype generation may fail for some functions. To work around this, you can provide your own prototypes
   for these functions.
 - `#line` directives are added to make warning or error messages reflect the original sketch layout.
+- Special and optional sketch variables named ARDUIFINE and ARDUINOGLOBAL are parsed, and their content is used to
+  provide more arguments to the compiler commands. Those arguments are used for every compiled source file, including
+  the core, internal and external libraries. Their content is displayed in the message area.
 
 No pre-processing is done to files in a sketch with any extension other than .ino or .pde. Additionally, .h files in the
 sketch are not automatically #included from the main sketch file. Further, if you want to call functions defined in a .c
--- a/legacy/builder/builder_utils/utils.go
+++ b/legacy/builder/builder_utils/utils.go
@@ -229,7 +229,9 @@ func compileFileWithRecipe(ctx *types.Context, sourcePath *paths.Path, source *p
 	properties := buildProperties.Clone()
 	properties.Set(constants.BUILD_PROPERTIES_COMPILER_WARNING_FLAGS, properties.Get(constants.BUILD_PROPERTIES_COMPILER_WARNING_FLAGS+"."+ctx.WarningsLevel))
 	properties.Set(constants.BUILD_PROPERTIES_INCLUDES, strings.Join(includes, constants.SPACE))
-	properties.Set(constants.BUILD_PROPERTIES_INCLUDES, properties.Get(constants.BUILD_PROPERTIES_INCLUDES) + " " + ctx.Arduifines + " ")
+	if len(ctx.Arduifines) > 0 {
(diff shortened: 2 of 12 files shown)


d-a-v:
I would have preferred to propose something like
```
#arduifine BLAH 1
```
or
```
#pragma comment("ARDUIFINE: -Dblah")
```
but that would have required parsing the whole .`ino` files.
The simpler the better: the ctags parser is already here.

d-a-v:
About security, the compiler command line is affected but user is warned in the message console.
Additional parsing can be added to ensure that not too special characters are used in the macro defined by "ARDUIFINE" and "ARDUINOGLOBAL" feature can be disabled if it is considered as too dangerous.
@facchinm this is close to [this comment](https://github.com/arduino/Arduino/issues/421#issuecomment-327003003).

This PR, apart from adding more noise to an old feature request, is another attempt to finally find an answer to https://github.com/arduino/Arduino/issues/421. 
Per this PR, I'd like to bring to anyone interested's attention that the feature is needed.
One of the important use-cases is that it allows users to avoid modifying libraries when some parameters are to be tuned.
(this feature is provided with other IDE able to compile Arduino projects like platformIO or sloeber)

I'm aware that I updated in the `legacy/` directory. Are there specific considerations about this directory that should be known by contributors ?

I will be happy to close this PR if any other way is found to close https://github.com/arduino/Arduino/issues/421 which has been opened *for 8 years*. Attempts or explorations are quite numerous, like [this commit](https://github.com/facchinm/arduino-builder/commit/904250febf1dea20408bdd1d1fa8950195f5b7a1)(from @facchinm in #421), or one from the link list compiled by @per1234 in #846's OP. I'll also be glad to update this PR until it fits with maintainers idea of what are arduino-cli / arduino-builder.

d-a-v:
> [My opinion is that there are valid use cases for macro-based interfaces [...] but that it is not appropriate to expect normal users to define global macros in order to use a library](https://github.com/arduino-libraries/ArduinoBearSSL/pull/45#issuecomment-815004577)

@per1234, this change proposal even as a PoC is orthogonal to your above statement.
You make me believe this opinion is close to official Arduino's, in which case I should close this PR :)

#846's title "*Provide dedicated sets of "extra flags" properties for platform developers **and users***" did make me think that it is "legal" to provide such facility to users.

I have indeed two use-cases on widely used arduino libraries:

- When a ~bug~ request takes too long to be fixed upstream in a published and popular library, [sometimes this kind of "global define hack" is needed](https://github.com/knolleary/pubsubclient/issues/764),
- Nobody wants to locally modify a library in order to properly use it, but [sometimes configuring would result in a bigger code than when using a global define](https://github.com/mcci-catena/arduino-lmic/blob/1aafca50ae3f5050c05567b76d028e30711b2ba7/project_config/lmic_project_config.h) (@matthijskooijman and @terrillmoore know this one well)
 
IMHO It could be a normal use for an Arduino library to need a global define to get some extended behaviour, at no cost when 
they are not needed (= without state variable and related code that costs space on small hw).

I would really like clarification on how Arduino plans to address such issues.

PR #1145, opened 2021-01-20
Enhance lib install from git url and zip file

silvanocerza:

      before creating one)

Enhances the behaviour of `lib install` when using the `--git-url` or `--zip-file` flags.

Previously installing from a zip file would extract the content directly into the user library folder, this could have caused problem if the zip file wasn't structured correctly.

Installing from a git repository would clone the repo directly to the user library folder and leave it as it is. Also it would fail if the library is already installed and would not reinstall it.

When using the `--zip-file` flag we now extract to a temp directory to verify that the zip is structured correctly and to infer the name of the library being installed. If everything is fine we then copy it over to the user library folder to finalize the installation.

When using the `--git-url` flag after cloning the repository to the user library folder we delete the `.git` folder to make it a plain folder. Cloning should be faster too since we now clone using depth of one commit to avoid downloading useless files.

Both when installing from a zip file or from a git repository we delete the installed library folder if one with the same name is already installed. This fixes failures when trying to install with `--git-url` an already installed library.

Nope.

* **Other information**:

None.

---

### DIFF
--- a/arduino/libraries/librariesmanager/install.go
+++ b/arduino/libraries/librariesmanager/install.go
@@ -28,6 +28,7 @@ import (
 	"github.com/arduino/arduino-cli/arduino/utils"
 	paths "github.com/arduino/go-paths-helper"
 	"github.com/codeclysm/extract/v3"
+	"github.com/sirupsen/logrus"
 	"gopkg.in/src-d/go-git.v4"
 )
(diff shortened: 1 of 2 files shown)


ubidefeo (approved):
tested and working

PR #1157, opened 2021-01-28
Add back support for sketches with .pde extension and deprecate it

silvanocerza:

      before creating one)

This PR brings back the support for sketches ending with `.pde` extension. 
Also we deprecate it.

Trying to `compile`, `upload`, or `archive` a sketch with a `.pde` main file always fails.
Also examples' libraries with `.pde` extensions are not shown when calling `lib examples`.

`compile`, `upload`, or `archive` a sketch with a `.pde` main file now works correctly and shows a warning message recommending the user to rename all `.pde` files `.ino`.
Examples' libraries with `.pde` extensions are now correctly shown when calling `lib examples`.

No.

* **Other information**:

Each commit fixes a different command, it should be easier to review each commit separately.

---

### DIFF
--- a/cli/compile/compile.go
+++ b/cli/compile/compile.go
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"os"
 
+	"github.com/arduino/arduino-cli/arduino/sketches"
 	"github.com/arduino/arduino-cli/cli/feedback"
 	"github.com/arduino/arduino-cli/cli/output"
 	"github.com/arduino/arduino-cli/configuration"
@@ -127,6 +128,15 @@ func run(cmd *cobra.Command, args []string) {
 	}
 
 	sketchPath := initSketchPath(path)
+
+	// .pde files are still supported but deprecated, this warning urges the user to rename them
+	if files := sketches.CheckForPdeFiles(sketchPath); len(files) > 0 {
+		feedback.Error("Sketches with .pde extension are deprecated, please rename the following files to .ino:")
+		for _, f := range files {
+			feedback.Error(f)
+		}
+	}
+
 	// We must read this from settings since the value is set when the binding is accessed from viper,
 	// accessing it from cobra would only read it if the flag is explicitly set by the user and ignore
 	// the config file and the env vars.
--- a/commands/sketch/archive.go
+++ b/commands/sketch/archive.go
@@ -23,8 +23,10 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/arduino/arduino-cli/arduino/globals"
 	rpc "github.com/arduino/arduino-cli/rpc/commands"
 	paths "github.com/arduino/go-paths-helper"
+	"github.com/pkg/errors"
 )
 
 // ArchiveSketch FIXMEDOC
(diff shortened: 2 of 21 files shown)


test/test_upload.py
```diff
@@ -195,3 +198,143 @@ def test_compile_and_upload_combo_with_custom_build_path(run_command, data_dir,
...
+        wait_for_board()
+        assert run_command(f"upload -b {board.fqbn} -p {board.address} {sketch_path}")
+
+        # Upload from sketch file
+        wait_for_board()
+        assert run_command(f"upload -b {board.fqbn} -p {board.address} {sketch_file}")
+
+        # Upload from build folder
+        sketch_path_md5 = hashlib.md5(bytes(sketch_path)).hexdigest().upper()
+        build_dir = Path(tempfile.gettempdir(), f"arduino-sketch-{sketch_path_md5}")
```
cmaglie, line 265: You should not compute this value here but get it from `compile --format json` output.

test/test_upload.py
```diff
@@ -195,3 +198,143 @@ def test_compile_and_upload_combo_with_custom_build_path(run_command, data_dir,
...
+
+        # Compile both sketches and copy binaries in the same directory same build directory
+        assert run_command(f"compile --clean -b {board.fqbn} {sketch_one_path}")
+        assert run_command(f"compile --clean -b {board.fqbn} {sketch_two_path}")
+
+        sketch_path_md5 = hashlib.md5(bytes(sketch_one_path)).hexdigest().upper()
+        build_dir_one = Path(tempfile.gettempdir(), f"arduino-sketch-{sketch_path_md5}")
+
+        sketch_path_md5 = hashlib.md5(bytes(sketch_two_path)).hexdigest().upper()
+        build_dir_two = Path(tempfile.gettempdir(), f"arduino-sketch-{sketch_path_md5}")
```
cmaglie, line 311: again, build path should be get from `compile --format json`, there may be other places to change

cli/compile/compile.go
```diff
@@ -127,6 +128,15 @@ func run(cmd *cobra.Command, args []string) {
...
 	sketchPath := initSketchPath(path)
+
+	// .pde files are still supported but deprecated, this warning urges the user to rename them
+	if files := sketches.CheckForPdeFiles(sketchPath); len(files) > 0 {
+		feedback.Error("Sketches with .pde extension are deprecated, please rename the following files to .ino:")
+		for _, f := range files {
+			feedback.Error(f)
+		}
+	}
+
```
cmaglie, line 139: This piece is duplicated in compile.go, upload.go and archive.go.
On compile.go and upload.go there is also `initSketchPath()` that is duplicated as well.
This makes me wonder two things:
* in archive.go we are missing some cases because we miss an `initSketchPath(...)`
* we must move everything (the current initSketchPath and the pde check) into a common function `args.InitSketchParh(path)` or something like that...

commands/sketch/archive.go
```diff
@@ -42,22 +44,38 @@ func ArchiveSketch(ctx context.Context, req *rpc.ArchiveSketchReq) (*rpc.Archive
...
+		if candidateSketchMainFile.Exist() {
+			if sketchMainFile == nil {
+				sketchMainFile = candidateSketchMainFile
+			} else {
+				return nil, errors.Errorf("multiple main sketch files found (%v, %v)",
+					sketchMainFile,
+					candidateSketchMainFile,
+				)
+			}
+		}
```
cmaglie, line 68: This is a duplicate code: the same loop exists in `sketch.NewSketchFromPath`

cmaglie (changes requested):

cli/compile/compile.go
silvanocerza: This is probably something that we need to tackle in a systematic way and not in this PR or it would make it explode.

cmaglie (approved):

PR #1170, opened 2021-02-04
Better error message for boards without bootloader

cmaglie:
Before:

```
$ arduino-cli upload -b arduino:avr:gemma -p /dev/ttyACM0
avrdude: invalid baud rate specified '{upload.speed}'
Error during Upload: uploading error: uploading error: exit status 1
$
```

After:

```
$ arduino-cli upload -b arduino:avr:gemma -p /dev/ttyACM0
Error during Upload: a programmer is required to upload for this board
$
```

### DIFF
--- a/commands/upload/upload.go
+++ b/commands/upload/upload.go
@@ -228,6 +228,10 @@ func runProgramAction(pm *packagemanager.PackageManager,
 		}
 	}
 
+	if !uploadProperties.ContainsKey("upload.protocol") && programmer == nil {
+		return fmt.Errorf("a programmer is required to upload for this board")
+	}
+
 	// Set properties for verbose upload
 	if verbose {
 		if v, ok := uploadProperties.GetOk("upload.params.verbose"); ok {


matthijskooijman:
Is this really the right place and right check for this? Looking at `boards.txt`, it would make more sense to check for the presence of `upload.tool`, which I think is used to lookup additional properties from `platform.txt` during upload. If that entry would be missing, that would be a clearer sign that regular upload is not supported.

One caveat is that currently e.g. the `gemma` *does* have a `upload.tool` specified, presumably since without it some other bit breaks (or used to break, maybe), so this would require changes in the platforms as well.

I haven't tested any of this, just commenting based on a quick review...

cmaglie:
> One caveat is that currently e.g. the gemma does have a upload.tool specified, presumably since without it some other bit breaks (or used to break, maybe), so this would require changes in the platforms as well.

Not only the `gemma` but a lot of other platforms like attiny, xmega, etc... They use this trick to force the Arduino IDE to perform an "upload-with-programmer" even if the user clicks on "upload".

The relevant part of the Arduino IDE is this: https://github.com/arduino/Arduino/blob/b3d609f4de066289088739747a743076cf4ea023/arduino-core/src/cc/arduino/packages/uploaders/SerialUploader.java#L87-L91 

It's also specified in https://arduino.github.io/arduino-cli/dev/platform-specification/#upload-using-programmer-by-default

I'm going to merge it, but I'm happy to discuss a way to improve it if you have any idea.

PR #1181, opened 2021-02-10
Do not try to export binaries if --only-compilation-database is set

cmaglie:
Otherwise an empty `build/fqbn/...` folder is created if `always-export-binaries` option is set via config file.


silvanocerza (changes requested):
We need an integration test to verify this new behaviour, the rest looks good.

ubidefeo (approved):
Tested in the latest build and LS operations will not create the `build` folder.
Works fine

silvanocerza (approved):

silvanocerza:
Rebased to fix conflicts.

PR #1201, opened 2021-02-26
Skip unused broken symlinks

matthijskooijman:

      before creating one)
- [X] The PR follows

Bug fix

1. When a sketch file (i.e. `.ino`, `.cpp`, etc.) is unreadable, it is silently skipped when loading the sketch.
2. When a sketch contains a broken symlink, sketch loading fails even when the file would not be otherwise loaded (i.e. not a `.ino`, `.cpp`, etc type of file).

1. When a sketch file (i.e. `.ino`, `.cpp`, etc.) is unreadable (i.e. no permission or otherwise unreadable), an error is thrown.
2. When a sketch contains a broken symlink, it is skipped if the file would not be otherwise loaded (i.e. not a `.ino`, `.cpp`, etc type of file).

Sketches that contain unreadable files that are not actually needed can no longer be compiled (they show an error instead of silently skipping the unreadable file), but I would consider such sketches an unsupported corner case.

* **Other information**:
The trigger for this PR was that I had a symlink in my sketch to the `compile_commands.json` in the build folder, so my editor would find that file and use it for code completion. However, after a reboot, the symlink would become broken, and `arduino-cli compile` (even with `--only-compilation-database` would fail, requiring me to remove the symlink and then recreate it afterwards.

The first commit was not strictly required for this problem, but it seemed like a good idea to not silently ignore unreadable files, and raising this error simplifies the second commit, while still raising an error as before on symlinks that would be loaded.

I've omitted testcases from this PR, since a significant of the code paths to test would result in `os.Exit`, which I think is not supported directly in the test framework. Are there any strategies in place for handling this already (see [this post](https://stackoverflow.com/questions/26225513/how-to-test-os-exit-scenarios-in-go) for some suggestions if not)?

### DIFF
--- a/arduino/builder/sketch.go
+++ b/arduino/builder/sketch.go
@@ -154,12 +154,15 @@ func SketchLoad(sketchPath, buildPath string) (*sketch.Sketch, error) {
 	var files []string
 	rootVisited := false
 	err = simpleLocalWalk(sketchFolder, maxFileSystemDepth, func(path string, info os.FileInfo, err error) error {
-		if err != nil {
+		// Ignore ErrNotExist (broken symlinks), which will be
+		// reported by os.Open below if the file would not be
+		// skipped based on the name
+		if err != nil && !errors.Is(err, os.ErrNotExist) {
 			feedback.Errorf("Error during sketch processing: %v", err)
 			os.Exit(errorcodes.ErrGeneric)
 		}
 
-		if info.IsDir() {
+		if info != nil && info.IsDir() {
 			// Filters in this if-block are NOT applied to the sketch folder itself.
 			// Since the sketch folder is the first one processed by simpleLocalWalk,
 			// we can set the `rootVisited` guard to exclude it.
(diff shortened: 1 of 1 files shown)


matthijskooijman:
Any chance of merging this? It's a fairly straightforward bugfix :-)

matthijskooijman:
Hm, I just rebased this on top of latest master, and discovered that his fix no longer works entirely as expected due to #1235. That PR added support for .json files, so this means that a broken symlink to `compile_commands.json` now again triggers an error. This is because this PR ensures that broken symlinks that would not be loaded as a sketch file at all are ignored, but with #1235 `compile_commands.json` is now a valid sketch file that must be loaded, and thus also reports an error if it is a broken symlink...

I'm not directly sure what a good approach for this is, though. Maybe add an explicit ignore for `compile_commands.json` (since I don't think it should be listed as a sketch file in any case), and then also apply the changes in the PR?

matthijskooijman:
Closing in favor of #1438.

PR #1204, opened 2021-03-02
Update serial discovery to 1.2.0

cmaglie:
This updates the tool to v1.2.0, changelog:

* added support for macosx big-sur
* fixed wrong USB serial number detection in Windows for USB composite devices


cmaglie:
nevermind, wrong release :sigh:

PR #1207, opened 2021-03-02
Update serial discovery

cmaglie:
This updates the tool to v1.2.1, changelog:

* added support for macosx big-sur
* fixed wrong USB serial number detection in Windows for USB composite devices

### DIFF
--- a/commands/bundled_tools_serial_discovery.go
+++ b/commands/bundled_tools_serial_discovery.go
@@ -32,15 +32,15 @@ import (
 )
 
 var (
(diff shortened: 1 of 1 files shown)


ubidefeo (approved):
tested and working correctly

PaulStoffregen:
Is Pluggable Discovery going to be supported?

cmaglie:
> Is Pluggable Discovery going to be supported?

Yes! We postponed it a couple of times but it's definitely in our planning.
I'll share more info as we start working on it.

PR #1214, opened 2021-03-08
Add overwrite option to zip and git lib install for gRPC interface

silvanocerza:

      before creating one)
- [x] The PR follows

Enhances two gRPC interface functions

The gRPC interface fails if a zip and git library is already installed. 
The temporary directory used to extract the zip archive is not cleaned after installation.

An `Overwrite` flag can now be set to specify what to do when a library with the same name is installed.
The temporary directory used to extract the zip archive is now cleaned after installation.

The CLI will always overwrite the installed library if it exists.

No breaking changes.

* **Other information**:

None.

---


arduino/libraries/librariesmanager/install.go
```diff
@@ -126,11 +130,29 @@ func (lm *LibrariesManager) InstallZipLib(ctx context.Context, archivePath strin
...
+	defer func() {
+		// Clean up install dir if installation failed
+		files, err := installPath.Parent().ReadDir()
+		if err != nil {
+			return
+		}
+		if len(files) == 0 {
+			installPath.Parent().RemoveAll()
+		}
+	}()
```
cmaglie, line 149: IIUC:
```
libsDir              == "/home/user/sketchbook/libraries"
installPath          == "/home/user/sketchbook/libraries/LIBNAME"
installPath.Parent() == "/home/user/sketchbook/libraries"
```

so instead of `installPath.Parent()` you can use `libsDir` that makes the goal of this piece of code much more explicit: create the `libraries` directory if not exist and remove it when it's no more needed

cmaglie (changes requested):

arduino/libraries/librariesmanager/install.go
silvanocerza: Looking at this again I think I made a mistake actually, I don't want to verify if `/home/user/sketchbook/libraries` is empty but `/home/user/sketchbook/libraries/LIBNAME`. So `installPath` is correct but I don't have to get its parent. :thinking:

cmaglie (approved):

PR #1224, opened 2021-03-16
Add #line tags to AdditionalFiles when copying to build directory

cmaglie:
Improves error messages.

This is the patch from @oschonrock #707 rebased on `master`.


ubidefeo (approved):
It works!!!
nice one, @cmaglie

PR #1234, opened 2021-03-24
Add library validation step when installing from zip or git

silvanocerza:

      before creating one)
- [x] The PR follows

Enhances an existing command.

Calling `lib install` with `--zip-file` or `--git-url` doesn't check if the used zip or git repository contain a valid library.

This affects also the gRPC interface.

Calling `lib install` with `--zip-file` or `--git-url` now checks if the used zip or git repository contain a valid library and fails if it not.

A valid library must contain an header file with the same name as the library and a `library.properties` file as documented.
https://arduino.github.io/arduino-cli/latest/library-specification

This affects also the gRPC interface.

Nope.

* **Other information**:

None.

---


test/test_lib.py
```diff
@@ -868,3 +905,71 @@ def test_install_zip_lib_with_macos_metadata(run_command, data_dir, downloads_di
...
+    repo_dir = Path(data_dir, "lib-without-header")
+    repo = Repo.init(repo_dir)
+    lib_properties = Path(repo_dir, "library.properties")
+    lib_properties.touch()
+    repo.index.add([str(lib_properties)])
+    repo.index.commit("First commit")
+
+    res = run_command(f"lib install --git-url {repo_dir}", custom_env=env)
+    assert res.failed
+    assert 'library is not valid: missing header file "lib-without-header.h"' in res.stderr
```
umbynos, line 962: maybe it would be nice to check also the [cleanup of the install directory](https://github.com/arduino/arduino-cli/pull/1234/files#diff-7201cbc75bf18b11ac71b3a4a12caae78081c2d6be4e164db68db73ba10a0ea6R229)?

arduino/libraries/librariesmanager/install.go
```diff
@@ -133,6 +133,11 @@ func (lm *LibrariesManager) InstallZipLib(ctx context.Context, archivePath strin
 
 	extractionPath := paths[0]
 	libraryName := extractionPath.Base()
+
+	if err := validateLibrary(libraryName, extractionPath); err != nil {
+		return err
```
umbynos, line 138: Probably a stupid question: why the extraction path is not cleared as done [here](https://github.com/arduino/arduino-cli/pull/1234/files#diff-7201cbc75bf18b11ac71b3a4a12caae78081c2d6be4e164db68db73ba10a0ea6R229)?

arduino/libraries/librariesmanager/install.go
```diff
@@ -239,3 +251,23 @@ func parseGitURL(gitURL string) (string, error) {
...
+	// Checks also root folder for legacy reasons.
+	// For more info see:
+	// https://arduino.github.io/arduino-cli/latest/library-specification/#source-code
+	libraryHeader := name + ".h"
+	if !dir.Join("src", libraryHeader).Exist() && !dir.Join(libraryHeader).Exist() {
+		return fmt.Errorf(`library is not valid: missing header file "%s"`, libraryHeader)
+	}
+
+	// Verifies library contains library.properties
+	if !dir.Join("library.properties").Exist() {
```
umbynos, line 268: maybe would be the case to check also if the `library.property` is not empty? Or includes the required fields?

arduino/libraries/librariesmanager/install.go
silvanocerza: There's a `defer tmpDir.RemoveAll()` at line `108`, we don't do the same when installing from git url since we clone directly in the installation folder in that case.

test/test_lib.py
silvanocerza: Yep, I totally forgot.

arduino/libraries/librariesmanager/install.go
silvanocerza: Not really sure about that, ideally that's Arduino Lint job.
Bringing the lint tool as a dependency to check on these things would be a good idea but the lint tool has the CLI as a dependency, it would open a huge can of worms I think.

umbynos: Nevermind, you are right, probably was a bad idea :space_invader:

umbynos (approved):
LGTM :racehorse:

PR #1251, opened 2021-03-30
Better gRPC error handling

cmaglie:
This is a proposal to uniform gRPC error handling.

For now only the `upload` command has been ported, as a proof-of-concept. It will be extended to all the `arduino-cli` code base if approved.

This PR, once fully implemented, should:

* change all gRPC API implementations to return a `status.Status` instead of an `error` (the `Status` message is how the gRPC transports errors on RPC calls).
* application-specific errors must be mapped to gRPC `codes`
* application-specific errors that may be useful to "identify" client-side are reported also in the `Status.details` field (for example the `UploadRequiresProgrammerError`)


commands/upload/upload.go
```diff
@@ -37,23 +37,25 @@ import (
...
+		return nil, status.Newf(codes.InvalidArgument, "Sketch not found: %s", err)
 	}
 
 	pm := commands.GetPackageManager(req.GetInstance().GetId())
 
-	err = runProgramAction(
+	return &rpc.UploadResponse{}, runProgramAction(
 		pm,
 		sketch,
 		req.GetImportFile(),
```
silvanocerza, line 61: Shouldn't we check for errors from `runProgramAction` and return a `nil` `rpc.UploadResponse` and the `status.Status`?

silvanocerza (commented):
I like this, it's a great improvement that's very much needed. I just have some doubt about the returning an `rpc.UploadResponse` even when the `status.Status` might contain errors.

commands/upload/upload.go
cmaglie: Yes you're correct we should return `nil, somestatus` or `response, nil` (I've taken a shortcut here, I'll fix it later)

per1234 (commented):
Is this considered a breaking change to the API?

If so, does the standard process for documenting it need to be done still?
https://arduino.github.io/arduino-cli/latest/CONTRIBUTING/#pull-request-checklist

cmaglie:
🤦🏼 right, we should prepare a migration guide.

I'll write down some notes later today...

cli/lib/args.go
```diff
@@ -75,13 +75,13 @@ func ParseLibraryReferenceArgs(args []string) ([]*LibraryReferenceArg, error) {
 // ParseLibraryReferenceArgAndAdjustCase parse a command line argument that reference a
 // library and possibly adjust the case of the name to match a library in the index
 func ParseLibraryReferenceArgAndAdjustCase(instance *rpc.Instance, arg string) (*LibraryReferenceArg, error) {
-	libRef, err := ParseLibraryReferenceArg(arg)
+	libRef, _ := ParseLibraryReferenceArg(arg)
```
silvanocerza, line 78: Shouldn't we fix this by handling this `err` instead of ignoring it?

cli/lib/check_deps.go
```diff
@@ -51,13 +51,13 @@ func runDepsCommand(cmd *cobra.Command, args []string) {
...
-	deps, err := lib.LibraryResolveDependencies(context.Background(), &rpc.LibraryResolveDependenciesRequest{
+	deps, stat := lib.LibraryResolveDependencies(context.Background(), &rpc.LibraryResolveDependenciesRequest{
 		Instance: instance,
 		Name:     libRef.Name,
 		Version:  libRef.Version,
 	})
-	if err != nil {
-		feedback.Errorf(tr("Error resolving dependencies for %[1]s: %[2]s"), libRef, err)
+	if stat != nil {
+		feedback.Errorf(tr("Error resolving dependencies for %[1]s: %[2]s", libRef, err))
```
silvanocerza, line 60:
```suggestion
		feedback.Errorf(tr("Error resolving dependencies for %[1]s: %[2]s"), libRef, err)
```

client_example/main.go
```diff
@@ -65,8 +66,12 @@ func main() {
 
 	settingsClient := settings.NewSettingsServiceClient(conn)
 
+	callUpload(client, initInstance(client))
+	return
 	// Now we can call various methods of the API...
-
+	log.Println("calling LibraryDownload(WiFi101@0.15.2)")
+	callLibDownload(client, initInstance(client))
+	return
```
silvanocerza, line 74: I'd remove this and the above `return`.

commands/board/list.go
```diff
@@ -190,19 +192,19 @@ func List(instanceID int32) (r []*rpc.DetectedPort, e error) {
...
 	pm := commands.GetPackageManager(instanceID)
 	if pm == nil {
-		return nil, errors.New(tr("invalid instance"))
+		return nil, status.New(codes.InvalidArgument, tr("Invalid instance"))
 	}
 
 	ports, err := commands.ListBoards(pm)
 	if err != nil {
-		return nil, errors.Wrap(err, fmt.Sprintf(tr("error getting port list from %s"), "serial-discovery"))
+		return nil, status.New(codes.FailedPrecondition, errors.Wrap(err, tr("Error getting port list from serial-discovery")).Error())
```
silvanocerza, line 200:
```suggestion
		return nil, status.New(codes.FailedPrecondition, errors.Wrap(err, fmt.Sprintf(tr("error getting port list from %s"), "serial-discovery")).Error())
```

cli/lib/args.go
per1234: Probably already clear, but I did this because this assignment of `err` was ignored and the `error` type from the declaration was incompatible with the change of `lib.LibrarySearch()`'s return type from `error` to `*status.Status`.

If it's to be ignored, I think it's best to make that explicit by using `_`. But if an error return from `ParseLibraryReferenceArg()` is significant in this context, then of course it's better to handle it.

cli/core/upgrade.go
```diff
@@ -95,10 +95,15 @@ func runUpgradeCommand(cmd *cobra.Command, args []string) {
...
-			feedback.Errorf(tr("Error during upgrade: %v"), err)
+		if err != nil {
+			if d := err.Details(); len(d) > 0 {
+				if _, ok := d[0].(*rpc.AlreadyAtLatestVersionError); ok {
+					feedback.Printf(tr("Platform %s is already at the latest version", platformRef))
+					continue
+				}
+			}
+
+			feedback.Errorf(tr("Error during upgrade: %v", err))
```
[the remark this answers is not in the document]
cmaglie: On the CLI it still did not fail... are you talking about the gRPC call?

cli/lib/args.go
cmaglie: Yep, the `err` was ignored also before this PR (it was overwritten without being checked). This should be fixed in another PR BTW.

cli/core/upgrade.go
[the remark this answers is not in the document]
silvanocerza: That shouldn't happen. :thinking: 

`os.Exit(errorcodes.ErrGeneric)` is always called even if the details contains `rpc.AlreadyAtLatestVersionError`, previously we used to call `os.Exit(errorcodes.ErrGeneric)` only if `err` was different from `core.ErrAlreadyLatest` and not `nil`.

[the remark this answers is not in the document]
silvanocerza: AAAAH! Nevermind, my eyes are betraying me, I thought we were looping through the `err.Details()` result so the `continue` would refer to that but it doesn't. :facepalm:

silvanocerza:
The translation files must be updated, the rest is good. I'll approve as soon that's fixed. :+1:

cmaglie:
After trying the current solution for a bit I found two annoying problems:

1. the error messages are not consistent, just to make an example the snippet:
   ```go
   	packageManager := i.PackageManager
   	if packageManager == nil {
   		return nil, status.New(codes.InvalidArgument, tr("Invalid instance"))
   	}
   ```
   is repeated a lot in the codebase, but there are cases where the message is different:
   ```go
   		return nil, errors.Errorf(tr("incorrect instance")
   		return nil, errors.Errorf(tr("unable to find an instance with ID: %d"), instanceID)
   ```

2. developers that imports the `commands` package directly will have the following code broken:
   ```go
   detailsResp, err := board.Details(context.Background(), detailsReq)
   if err != nil {
       fmt.Println("Error getting board details:", err)
       return
   }
   ```
   even if it seems perfectly reasonable it prints an address instead of an error message, because `*status.Status` did not 
implement a `String()` method, this is not idiomatic.

**PROPOSAL**

To address the two problems above, while rebasing this PR on master, I'd like to try a different approach:

1. to avoid repetition of error strings we define a bunch of common errors objects in `commands/errors.go`, something like:
   ```go
   type InvalidInstanceError struct{}

   func (e *InvalidInstanceError) Error() string {
          return tr("Invalid instance")
   }
   ```

2. to make life easier for developers importing the `commands` package I'll try to:
   * set back the return type to `error` instead of `*status.Status`.
   * define a CommandError interface:
      ```go
      type CommandError interface {
             ToRPC() error
      }
      ```
   * implement the CommandError interface on all errors returned from the `commands` public API, in this case the `InvalidInstanceError` show above will become:
     ```go
     type InvalidInstanceError struct{}

     func (e *InvalidInstanceError) Error() string {
            return tr("Invalid instance")
     }
     func (e *InvalidInstanceError) ToRPCStatus() *status.Status {
            return status.New(codes.InvalidArgument, e.Error())
     }
     ```
   * the wrappers in `deamon.go` will use an helper function to convert the `error` in the correct `status.Status`:
     ```go
     func convertError(err error) error {
            if rpcErr, ok := err.(CommandError); ok {
                    return rpcErr.ToRPCStatus().Err()
             }
            return err
     }
     ```

silvanocerza:
Yes, I 100% agree with this new design. I think it also make it easier to handle errors internally and recover them gracefully.
Let's go! :racehorse:

per1234 (dismissed):

cli/lib/check_deps.go
```diff
@@ -51,13 +51,13 @@ func runDepsCommand(cmd *cobra.Command, args []string) {
 		os.Exit(errorcodes.ErrBadArgument)
 	}
 
-	deps, err := lib.LibraryResolveDependencies(context.Background(), &rpc.LibraryResolveDependenciesRequest{
+	deps, stat := lib.LibraryResolveDependencies(context.Background(), &rpc.LibraryResolveDependenciesRequest{
```
silvanocerza, line 54: `stat` is of type `error` we should probably rename it `err`, the name is kinda confusing otherwise.

commands/board/list.go
```diff
@@ -231,7 +231,7 @@ func Watch(instanceID int32, interrupt <-chan bool) (<-chan *rpc.BoardListWatchR
 	runErrs := dm.RunAll()
 	if len(runErrs) == len(dm.IDs()) {
 		// All discoveries failed to run, we can't do anything
-		return nil, fmt.Errorf("%v", runErrs)
+		return nil, &commands.UnavailableError{Message: tr("Error starting board discoveries"), Cause: runErrs[0]}
```
silvanocerza, line 234: I don't like that we only take the first `error`, we lose some information this way. 😕

commands/upload/upload.go
```diff
@@ -234,9 +237,9 @@ func runProgramAction(pm *packagemanager.PackageManager,
 	} else if programmer != nil {
 		action = "program"
 	}
-	uploadToolID, err := getToolID(props, action, port.Protocol)
-	if err != nil {
-		return err
+	uploadToolID, st := getToolID(props, action, port.Protocol)
+	if st != nil {
+		return st
```
silvanocerza, line 242:
```suggestion
	uploadToolID, err := getToolID(props, action, port.Protocol)
	if err != nil {
		return err
```

commands/upload/upload_test.go
```diff
@@ -180,7 +180,7 @@ func TestUploadPropertiesComposition(t *testing.T) {
 	testRunner := func(t *testing.T, test test, verboseVerify bool) {
 		outStream := &bytes.Buffer{}
 		errStream := &bytes.Buffer{}
-		err := runProgramAction(
+		status := runProgramAction(
```
silvanocerza, line 183: `runProgramAction` returns an `error`, we should change `status` to `err` to avoid confusion.

silvanocerza (changes requested):
All in all it's a great improvement but there are some small things to fix.

commands/board/list.go
cmaglie: Fixed with a `fmt.Errorf("%v", errs)` for now, a better PR is coming after this one.

silvanocerza (approved):

PR #1263, opened 2021-04-13
upload: do not give up if no port is specified for 1200 bps touch

cmaglie:
There are some particular boards where the serial port is required only to switch the board in bootloader mode but the actual upload is performed in other ways.

This patch allows to upload when the board is already in bootloader mode, otherwise, the cli will exit with the error:

```
no upload port provided
```

Previously the CLI would exit with an error anyway, so this change should not make any harm to other boards.


facchinm (approved):

silvanocerza (approved):

ubidefeo (approved):
Tested in IDE 2 and working

PR #1274, opened 2021-04-23
[breaking] Refactor initialization steps

silvanocerza:

      before creating one)
- [x] The PR follows

A big refactoring of the initialization steps run by the CLI both when run "one shot" or in daemon mode.

Currently the gRPC `Init` function that creates and initializes a new `CoreInstance` to handle platforms and libraries that are both installed and installable returns an error and fail completely if even just one installed platform or library fails to load for any reason.

This issue breaks completely the CLI or any gRPC consumer when the initialization step fails because no `CoreInstance` is created, thus preventing any other further operation.

The `Init` gRPC function has now been split into two separate functions, `Create` to create a new `CoreInstance` and folders necessary to correctly execute the CLI, and `Init` that loads indexes, platforms and libraries.

The `Create` function will only fail if the creation of the necessary folders will fail, in any other case a valid `CoreInstance` is returned.

The `Init` function has been greatly improved, now failure to load a single platform or library doesn't completely stop the loading, but a message is sent to the client to notify the failure, the client will decide how to handle the error.

The CLI will print an error to the user whenever a platform or library fails to load but execution won't stop, it will only fail if the platform or library is needed to run successfully the command.

The `Rescan` gRPC function has been removed completely in favour of `Init`.

`UpdateIndex` and `UpdateLibrariesIndex` gRPC functions now don't implicitly reload the indexes, it has to be done explicitly with the `Init` function.

Yes. Further information to handle the breaking change can be found in the `UPGRADING.md` file.

* **Other information**:

None.

---


per1234 (changes requested):

commands/core/install.go
```diff
@@ -54,9 +54,9 @@ func PlatformInstall(ctx context.Context, req *rpc.PlatformInstallRequest,
 		return nil, err
 	}
 
-	_, err = commands.Rescan(req.GetInstance().GetId())
-	if err != nil {
-		return nil, err
+	_, status := commands.Init(&rpc.InitRequest{Instance: &rpc.Instance{Id: req.Instance.Id}})
```
cmaglie, line 57: This may be changed to:
```suggestion
	_, status := commands.Init(&rpc.InitRequest{Instance: req.Instance})
```

silvanocerza:
Rebased after #1278 merge.

cmaglie (approved):

PR #1327, opened 2021-06-18
serial-discovery now talks pluggable discovery protocol v1

cmaglie:
* Updated serial-discovery
* `discovery.Discovery` class now talks pluggable discovery [protocol v1](https://github.com/arduino/tooling-rfcs/pull/2)
* Also added a grpc `Port` for future use

This change should not break anything.


arduino/discovery/discovery.go
```diff
@@ -117,24 +122,30 @@ func (disc *PluggableDiscovery) jsonDecodeLoop(in io.Reader, outChan chan<- *dis
 	closeAndReportError := func(err error) {
 		disc.statusMutex.Lock()
 		disc.alive = false
+		if debug {
+			fmt.Println("ERR:", err)
```
silvanocerza, line 126: Can't we just use `logrus.debug`?

arduino/cores/packagemanager/identify.go
```diff
@@ -41,6 +41,7 @@ func (pm *PackageManager) IdentifyBoard(idProps *properties.Map) []*cores.Board
 		}
 		return false, true
 	}
+	// FIXME: TODO...
```
silvanocerza, line 44: I'd remove this, it scares me. :sweat_smile:

silvanocerza (changes requested):
All in all it's fine for me, but I think debug printing could be done in a better way.

arduino/discovery/discovery.go
cmaglie: I guess I'll just remove it...

arduino/cores/packagemanager/identify.go
cmaglie: LOL, ok done

arduino/discovery/discovery.go
silvanocerza: I very much prefer if we keep some debugging capabilities, I'd just prefer if we do it in a more structured way.

silvanocerza (approved):

PR #1333, opened 2021-06-23
Add support for Pluggable Discoveries

silvanocerza:

      before creating one)
- [x] The PR follows

This PR introduces quite a bit of changes to add support for Pluggable Discoveries as specified by the related [RFC](https://github.com/arduino/tooling-rfcs/blob/main/RFCs/0002-pluggable-discovery.md).

Currently the Arduino CLI uses a bundled `serial-discovery` tool to search for supported boards connected to serial ports, there is no support for boards connected to the network or 3rd party boards that require specific tools to be discovered. 

There is currently no way for 3rd party platforms to specify a custom tool to run to discover boards.

After this PR is merged any platform will be able to specify custom tools to run at discovery time to find supported boards, for example when calling `arduino-cli board list`. 

This change is fully retro compatible with platform that don't specify a discovery tool, the bundled `serial-discovery` tool will be used as a fallback. 

Discovery tools will be internally loaded automatically when the Arduino CLI is initialized and run when necessary.

Yes, they are documented in the `docs/UPGRADING.md` file.

* **Other information**:

None.

---


arduino/cores/packagemanager/loader.go
```diff
@@ -590,3 +599,83 @@ func (pm *PackageManager) LoadToolsFromBundleDirectory(toolsPath *paths.Path) er
...
 }
+
+// LoadDiscoveries returns a list of PluggableDiscoveries supported by the specfied PlatformRelease.
+// Returns error if:
+// * A PluggableDiscovery instance can't be created
+// * Tools required by the PlatformRelease cannot be found
+// * Command line to start PluggableDiscovery has malformed or mismatched quotes
+func (pm *PackageManager) LoadDiscoveries(release *cores.PlatformRelease) ([]*discovery.PluggableDiscovery, error) {
+	res := []*discovery.PluggableDiscovery{}
+	discoveryProperties := release.Properties.SubTree("discovery").Clone()
```
cmaglie, line 610:
```suggestion
	discoveryProperties := release.Properties.SubTree("discovery")
```

arduino/cores/packagemanager/loader.go
```diff
@@ -590,3 +599,83 @@ func (pm *PackageManager) LoadToolsFromBundleDirectory(toolsPath *paths.Path) er
...
+	//    "discovery.required": "PLATFORM:DISCOVERY_NAME",
+	//
+	// Case 2:
+	//    "discovery.required.0": "PLATFORM:DISCOVERY_ID_1",
+	//    "discovery.required.1": "PLATFORM:DISCOVERY_ID_2",
+	//
+	// If both indexed and unindexed properties are found the unindexed are ignored
+	for _, id := range discoveryProperties.ExtractSubIndexLists("required") {
+		tool := release.Platform.Package.Tools[id]
+		toolRelease := tool.GetLatestInstalled()
```
cmaglie, line 628: `tool, ok := ....` to check for `nil`

arduino/cores/packagemanager/package_manager.go
```diff
@@ -429,6 +439,39 @@ func (pm *PackageManager) InstalledBoards() []*cores.Board {
...
+			rel := tool.GetLatestInstalled()
+			if rel != nil {
+				foundTools[rel.Tool.Name] = rel
+			}
+		}
+	}
+	// replace the default tools above with the specific required by the current platform
+	requiredTools := []*cores.ToolRelease{}
+	platform.ToolDependencies.Sort()
+	for _, toolDep := range platform.ToolDependencies {
```
cmaglie, line 460: Missing `DiscoveryDependencies`?

cmaglie (changes requested):

per1234:
I get a panic now if I try to upload without specifying a port:
```
$ ./arduino-cli version
arduino-cli.exe alpha Version: git-snapshot Commit: 5ab1ecde Date: 2021-07-22T16:58:19Z

$ ./arduino-cli compile -u -b arduino:avr:uno /tmp/FooSketch
[signal 0xc0000005 code=0x0 addr=0x20 pc=0xd4c6e9]

goroutine 1 [running]:
github.com/arduino/arduino-cli/cli/compile.run(0xc000755600, 0xc00011d380, 0x1, 0x4)
        C:/Users/per/go/src/github.com/arduino/arduino-cli/cli/compile/compile.go:206 +0x969
github.com/spf13/cobra.(*Command).execute(0xc000755600, 0xc00011d340, 0x4, 0x4, 0xc000755600, 0xc00011d340)
        C:/Users/per/go/pkg/mod/github.com/spf13/cobra@v1.0.1-0.20200710201246-675ae5f5a98c/command.go:846 +0x2c2
github.com/spf13/cobra.(*Command).ExecuteC(0xc00014fb80, 0x0, 0xc000210c60, 0x0)
        C:/Users/per/go/pkg/mod/github.com/spf13/cobra@v1.0.1-0.20200710201246-675ae5f5a98c/command.go:950 +0x375
github.com/spf13/cobra.(*Command).Execute(...)
        C:/Users/per/go/pkg/mod/github.com/spf13/cobra@v1.0.1-0.20200710201246-675ae5f5a98c/command.go:887
main.main()
        C:/Users/per/go/src/github.com/arduino/arduino-cli/main.go:31 +0x8f
```
Previously there was a friendly error message:
```
$ ./arduino-cli version
arduino-cli.exe alpha Version: git-snapshot Commit: fc367b77 Date: 2021-07-22T16:55:44Z

$ ./arduino-cli compile -u -b arduino:avr:uno /tmp/FooSketch
Error during Upload: uploading error: no upload port provided
```

docs/platform-specification.md
```diff
@@ -681,9 +761,112 @@ Other upload parameters can also be defined for the board. For example, in the A
...
+over the network. In this case the recipe must use the special placeholder {upload.field.FIELD_NAME} where FIELD_NAME
+must be declared separately in the recipe using the following format:
+
+    tools.UPLOAD_RECIPE_ID.upload.field.FIELD_NAME=FIELD_LABEL
+    tools.UPLOAD_RECIPE_ID.upload.field.FIELD_NAME.secret=true
+
+FIELD_LABEL is the label shown in the graphical prompt to ask the user to enter the value of the field. The secret
+property is optional and it should be set to true if the field is a secret (like passwords or token).
+
+Let’s see how a complete example will look like:
```
per1234, line 859:
```suggestion
Let's see a complete example:
```

docs/platform-specification.md
```diff
@@ -681,9 +761,112 @@ Other upload parameters can also be defined for the board. For example, in the A
...
+
+    tools.arduino_ota.upload.field.username=Username
+    tools.arduino_ota.upload.field.password=Password
+    tools.arduino_ota.upload.field.password.secret=true
+    tools.arduino_ota.upload.pattern="{runtime.tools.arduinoOTA.path}/bin/arduinoOTA"
+                                    -address {upload.port.address} -port 65280
+                                    -username "{upload.field.username}
+                                    -password "{upload.field.password}"
+                                    -sketch "{build.path}/{build.project_name}.bin"
```
per1234, line 868:
```suggestion
    tools.arduino_ota.upload.pattern="{runtime.tools.arduinoOTA.path}/bin/arduinoOTA" -address {upload.port.address} -port 65280 -username "{upload.field.username} -password "{upload.field.password}" -sketch "{build.path}/{build.project_name}.bin"
```
It was reasonable in the special case where you wanted to make it easy to see the before and after and made an explicit note about it being broken for clarity, but I think it will cause a lot of confusion to platform authors to do that in other parts of the specification, since the properties data format doesn't provide any way to break things into multiple lines AFAIK.

docs/platform-specification.md
```diff
@@ -756,7 +943,17 @@ The `program` action is triggered via the **Sketch > Upload Using Programmer** f
 compiled sketch to a board using an external programmer.
 
 The **program.tool** property determines the tool to be used for this action. This property is typically defined for
-each programmer in [programmers.txt](#programmerstxt):
+each programmer in [programmers.txt](#programmerstxt) and uses the same syntax as `upload` action:
```
per1234, line 946:
```suggestion
each programmer in [programmers.txt](#programmerstxt) and uses the same syntax as
[the `upload` action](#sketch-upload-configuration):
```

docs/platform-specification.md
```diff
@@ -786,7 +983,18 @@ the board.
 1. `bootloader` is used to flash the bootloader to the board
 
 The **bootloader.tool** property determines the tool to be used for the `erase` and `bootloader` actions both. This
-property is typically defined for each board in boards.txt:
+property is typically defined for each board in boards.txt and uses the same syntax as `upload` action:
```
per1234, line 986:
```suggestion
property is typically defined for each board in boards.txt and uses the same syntax as
[the `upload` action](#sketch-upload-configuration):
```

docs/platform-specification.md
```diff
@@ -786,7 +983,18 @@ the board.
...
+
+    [......]
+    uno.bootloader.tool.serial=avrdude
+    [......]
+    leonardo.upload.tool.serial=avrdude
+    leonardo.upload.tool.network=arduino_ota
+    [......]
+    duemilanove.upload.tool.default=avrdude
```
per1234, line 994:
```suggestion
    leonardo.bootloader.tool.serial=avrdude
    leonardo.bootloader.tool.network=arduino_ota
    [......]
    duemilanove.bootloader.tool.default=avrdude
```

docs/platform-specification.md
```diff
@@ -1009,3 +1217,7 @@ software is in use:
   This behavior
   [can be configured](https://arduino.github.io/arduino-cli/latest/commands/arduino-cli_core_install/#options)
 - **Arduino Pro IDE**: (since 0.1.0) runs the script for any installed platform.
+
+```
+
+```
```
per1234, line 1223:
```suggestion
```

per1234 (changes requested):

docs/package_index_json-specification.md
```diff
@@ -80,6 +80,7 @@ Each tool describes a binary distribution of a command line tool. A tool can be:
 - a file preprocessor
 - a debugger
 - a program that performs a firmware upgrade
+- a [Pluggable Discovery](pluggable-discovery-specification.md)
```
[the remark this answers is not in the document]
silvanocerza: I wasn't sure either but decided to go with the capitalized one since it's the name of the feature. But am ok if we want to make it lowercase.

docs/platform-specification.md
```diff
@@ -625,6 +625,64 @@ The tool configuration properties are available globally without the prefix. For
...
+if it supports also boards connected to the network it can use the builtin `mdns-discovery`:
+
+```
+discovery.required.0=builtin:serial-discovery
+discovery.required.1=builtin:mdns-discovery
+```
+
+Since the above syntax requires adding a discovery in the `discoveryDependencies` field in the platform
+`package_index.json` it might be cumbersome to do it before releasing it so we also provide another syntax to ease
+development:
```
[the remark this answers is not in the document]
silvanocerza: It should be possible to keep both with different IDs but they would both be started by the CLI if they're in the `platform.txt`, either one must be omitted or commented out.

docs/platform-specification.md
```diff
@@ -651,26 +709,48 @@ The Upload action is triggered when the user clicks on the "Upload" button on th
...
+    uno.upload.tool.default=avrdude
+    [......]
+    leonardo.upload.tool.default=avrdude
+    leonardo.upload.tool.network=arduino_ota
+    [......]
+
+`default` is also used when no upload address is provided by the user, this can be done only for boards that use upload
+tools with builtin port detection like `openocd`.
+
+For backward compatibility with IDE 1.8.15 and older the previous syntax is still supported:
```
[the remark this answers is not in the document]
silvanocerza: We can't know exactly which version of all products will contain this feature though, I think my also be misleading if we reference versions that are still not released. Probably we should update this when new versions are released.

silvanocerza:
> I get a panic now if I try to upload without specifying a port:
> 
> ```
> $ ./arduino-cli version
> arduino-cli.exe alpha Version: git-snapshot Commit: 5ab1ecde Date: 2021-07-22T16:58:19Z
> 
> $ ./arduino-cli compile -u -b arduino:avr:uno /tmp/FooSketch
> Sketch uses 444 bytes (1%) of program storage space. Maximum is 32256 bytes.
> Global variables use 9 bytes (0%) of dynamic memory, leaving 2039 bytes for local variables. Maximum is 2048 bytes.
> panic: runtime error: invalid memory address or nil pointer dereference
> [signal 0xc0000005 code=0x0 addr=0x20 pc=0xd4c6e9]
> 
> goroutine 1 [running]:
> github.com/arduino/arduino-cli/cli/compile.run(0xc000755600, 0xc00011d380, 0x1, 0x4)
>         C:/Users/per/go/src/github.com/arduino/arduino-cli/cli/compile/compile.go:206 +0x969
> github.com/spf13/cobra.(*Command).execute(0xc000755600, 0xc00011d340, 0x4, 0x4, 0xc000755600, 0xc00011d340)
>         C:/Users/per/go/pkg/mod/github.com/spf13/cobra@v1.0.1-0.20200710201246-675ae5f5a98c/command.go:846 +0x2c2
> github.com/spf13/cobra.(*Command).ExecuteC(0xc00014fb80, 0x0, 0xc000210c60, 0x0)
>         C:/Users/per/go/pkg/mod/github.com/spf13/cobra@v1.0.1-0.20200710201246-675ae5f5a98c/command.go:950 +0x375
> github.com/spf13/cobra.(*Command).Execute(...)
>         C:/Users/per/go/pkg/mod/github.com/spf13/cobra@v1.0.1-0.20200710201246-675ae5f5a98c/command.go:887
> main.main()
>         C:/Users/per/go/src/github.com/arduino/arduino-cli/main.go:31 +0x8f
> ```
> 
> Previously there was a friendly error message:
> 
> ```
> $ ./arduino-cli version
> arduino-cli.exe alpha Version: git-snapshot Commit: fc367b77 Date: 2021-07-22T16:55:44Z
> 
> $ ./arduino-cli compile -u -b arduino:avr:uno /tmp/FooSketch
> Sketch uses 444 bytes (1%) of program storage space. Maximum is 32256 bytes.
> Global variables use 9 bytes (0%) of dynamic memory, leaving 2039 bytes for local variables. Maximum is 2048 bytes.
> Error during Upload: uploading error: no upload port provided
> ```

That's no good. It should now be possible to upload without specifying a port for boards that use tools that automatically detected the correct port, but it certainly must not crash like this for those that don't.

Thanks for the great review anyway. :pray:

docs/package_index_json-specification.md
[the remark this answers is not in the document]
per1234: I don't have a super strong opinion on which way it should be, but I think this is the time to make a conscious decision and follow it consistently from here on. Otherwise, conscientious writers may waste time trying to determine the correct way to write this term in the face of inconsistent precedents.

@cmaglie do you have a preference re: "Pluggable Discovery" vs. "pluggable discovery"?

docs/platform-specification.md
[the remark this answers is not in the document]
per1234: OK. I was thinking that if there was a way for them to coexist then we might want to find a way to explain that, since I foresee this will be something of interest to platform developers. Since there isn't, I think the system is explained well enough.

docs/platform-specification.md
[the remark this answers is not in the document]
per1234: Yeah, it's really not a great situation. My concern is that the need to add the information will be forgotten. But guessing the versions risks the introduction of incorrect information.

I think that this will be very important information for platform developers using pluggable discovery because they will need to clearly document the minimum supported tool versions.

docs/package_index_json-specification.md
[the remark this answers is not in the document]
per1234: Oops, I forgot that Cristian is away right now.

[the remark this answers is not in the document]
silvanocerza: Yep, it's up to us. Probably it's going to be better to use the lowercase one, as you said it's not a proper noun so it doesn't come natural capitalizing it. :shrug:

docs/pluggable-discovery-specification.md
```diff
@@ -0,0 +1,368 @@
...
+  "error": true,
+  "message": "Unknown command XXXX"
+}
+```
+
+### Board identification
+
+The `properties` associated to a port can be used to identify the board attached to that port. The algorithm is simple:
+
+- each board listed in the platform file `boards.txt` may declare a set of `upload_port.*` properties
```
per1234, line 286:
```suggestion
- each board listed in the platform file [`boards.txt`](platform-specification.md#boardstxt) may declare a set of
  `upload_port.*` properties
```
The `upload_port.*` properties need to be documented in the platform specification.

docs/pluggable-discovery-specification.md
```diff
@@ -0,0 +1,368 @@
...
+  "ports": [
+    {
+      "address":       <-- THE ADDRESS OF THE PORT
+      "label":         <-- HOW THE PORT IS DISPLAYED ON THE GUI
+      "protocol":      <-- THE PROTOCOL USED BY THE BOARD
+      "protocolLabel": <-- HOW THE PROTOCOL IS DISPLAYED ON THE GUI
+      "properties": {
+                       <-- A LIST OF PROPERTIES OF THE PORT
+      }
+    }, {
```
per1234, line 133:
```suggestion
    },
    {
```
The style used by all other JSON in the documentation.

docs/pluggable-discovery-specification.md
```diff
@@ -0,0 +1,368 @@
...
+
+```JSON
+  "port": {
+    "address": "/dev/ttyACM0",
+    "properties": {
+      "pid": "0x804e",
+      "vid": "0x2341",
+      "serialNumber": "EBEABFD6514D32364E202020FF10181E",
+      "name": "ttyACM0"
+    }
```
per1234, line 304:
````suggestion
```
  "port": {
    "address": "/dev/ttyACM0",
    "properties": {
      "pid": "0x804e",
      "vid": "0x2341",
      "serialNumber": "EBEABFD6514D32364E202020FF10181E",
      "name": "ttyACM0"
    },
    ...
````
Even though it is handled correctly by the Markdown renderers, the intentionally omitted missing closing brace in this code block really freaks out GitHub's diff viewer, which I think could cause confusion to future contributors.

File: docs/pluggable-discovery-specification.md [hunk 16]
```diff
@@ -0,0 +1,368 @@
...
+```
+myboard.name=My Wonderful Arduino Compatible Board
+myboard.upload_port.0.pears=20
+myboard.upload_port.0.apples=30
+myboard.upload_port.1.pears=30
+myboard.upload_port.1.apples=40
+```
+
+will match on both `pears=20, apples=30` and `pears=30, apples=40` but not `pears=20, apples=40`, in that sense each
+"set" of identification properties is indepentent from each other and cannot be mixed for port matching.
```
per1234, line 368:
```suggestion
"set" of identification properties is independent from each other and cannot be mixed for port matching.
```

docs/pluggable-discovery-specification.md
```diff
@@ -0,0 +1,368 @@
...
+- a `label` that is the human readable form of the `address` (it may be for example `ttyACM0` or
+  `SSH on 192.168.10.100`)
+
+- `protocol` is the protocol identifier (such as `serial` or `dfu` or `ssh`)
+
+- `protocolLabel` is the `protocol` in human readable form (for example `Serial port` or `DFU USB` or `Network (ssh)`)
+
+- `properties` is a list of key/value pairs that represent information relative to the specific port
+
+To make the above more clear let’s show an example with the `serial_discovery`:
```
per1234, line 155: I don't understand what `serial_discovery` is. I don't find any other occurrences of this identifier anywhere else in the documentation or the RFC.

docs/pluggable-discovery-specification.md
```diff
@@ -0,0 +1,368 @@
...
+The `properties` associated to a port can be used to identify the board attached to that port. The algorithm is simple:
+
+- each board listed in the platform file `boards.txt` may declare a set of `upload_port.*` properties
+- if each `upload_port.*` property has a match in the `properties` set coming from the discovery then the board is a
+  “candidate” board attached to that port.
+
+Some port `properties` may not be precise enough to uniquely identify a board, in that case more boards may match the
+same set of `properties`, that’s why we called it “candidate”.
+
+Let’s see an example to clarify things a bit, let's suppose that we have the following `properties` coming from the
```
per1234, line 293:
```suggestion
  "candidate" board attached to that port.

Some port `properties` may not be precise enough to uniquely identify a board, in that case more boards may match the
same set of `properties`, that's why we called it "candidate".

Let's see an example to clarify things a bit, let's suppose that we have the following `properties` coming from the
```
I don't trust this highfalutin "smart quotes" stuff and we don't use it in the rest of the documentation. It was introduced from Discourse, which forces them on you.

docs/pluggable-discovery-specification.md
```diff
@@ -0,0 +1,368 @@
...
+    "properties": {
+      "pid": "0x804e",
+      "vid": "0x2341",
+      "serialNumber": "EBEABFD6514D32364E202020FF10181E",
+      "name": "ttyACM0"
+    }
+```
+
+in this case we can use `vid` and `pid` to identify the board. The `serialNumber`, instead, is unique for that specific
+instance of the board so it can't be used to identify the board model. Let’s suppose we have the following `boards.txt`:
```
per1234, line 308:
```suggestion
instance of the board so it can't be used to identify the board model. Let's suppose we have the following `boards.txt`:
```

mkdocs.yml
```diff
@@ -121,6 +122,7 @@ nav:
   - sketch-specification.md
   - library-specification.md
   - platform-specification.md
+  - Pluggable Discovery specification: pluggable-discovery-specification.md
```
per1234, line 125: Was there a specific reason for the placement here? I'm thinking that, since more platform developers will be interested in writing a package index than will be interested in writing a pluggable discovery tool, the order of the two should be swapped. But there might be another logical reason for the current order that I'm missing.

rpc/cc/arduino/cli/commands/v1/upload.proto
```diff
@@ -52,6 +53,12 @@ message UploadRequest {
   // If set to true, the actual upload will not be performed but a trace output
   // will be printed stdout. This is for debugging purposes.
   bool dry_run = 10;
+  // User provided fields usually used by upload tools that need authentication
+  // or in any case fields that can be customized by the user at upload time
+  // and cannot be known previously.
+  // For more info:
+  // https://github.com/arduino/tooling-rfcs/blob/main/RFCs/0002-pluggable-discovery.md#user-provided-fields
```
per1234, line 60:
```suggestion
  // https://github.com/arduino/tooling-rfcs/blob/main/RFCs/0002-pluggable-discovery.md#user-provided-fields
```
I wonder if it makes more sense to reference the platform specification instead of the RFC in this sort of comment?

It seems that you have documented most of the important information from the RFC in the CLI docs specifications now, so it seems like the RFC can serve more as a decision record while the specifications are the actively maintained documentation that is updated as the system evolves over time.

rpc/cc/arduino/cli/commands/v1/upload.proto
```diff
@@ -133,3 +152,30 @@ message ListProgrammersAvailableForUploadRequest {
...
+  // with the FQBN.
+  string protocol = 3;
+}
+
+message UserField {
+  // Id of the tool that supports this field
+  string tool_id = 1;
+  // Name used internally to store and retrieve this field
+  string name = 2;
+  // Label is the text shown to the user when needs to input this field
```
per1234, line 170:
```suggestion
  // Label is the text shown to the user when they need to input this field
```

per1234 (changes requested):

mkdocs.yml
silvanocerza: I just thought that since it's more related to the platforms' specification it would make sense for it to come after those. I can move below the package index specification if you think it's better.

rpc/cc/arduino/cli/commands/v1/upload.proto
silvanocerza: This kind of a vicious circle really, when I wrote that comment I still didn't write the documentation so I had only the RFC available.
Also I think there are other occurences in this PR where I do the same.

docs/pluggable-discovery-specification.md
silvanocerza: I changed it, should be clearer now.

docs/pluggable-discovery-specification.md
```diff
@@ -0,0 +1,368 @@
+These tools must be in the form of executables that can be launched as a subprocess using a `platform.txt` command line
```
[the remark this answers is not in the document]
silvanocerza: I added a small introduction with https://github.com/arduino/arduino-cli/pull/1333/commits/fb1dde149b745e1c684355e870c60c8e4805d461.

rpc/cc/arduino/cli/commands/v1/upload.proto
per1234: >I think there are other occurences in this PR where I do the same.

Correct. I picked on at random to comment on to see what your opinion was.

What do you think? Is the specification a better reference to direct the reader to, or the RFC?

mkdocs.yml
per1234: After a bit more consideration, I think it's OK where it is.

docs/pluggable-discovery-specification.md
per1234: @silvanocerza I see that my suggestion to linkify `boards.txt` was accepted, but I still didn't find any mention of the `upload_port` properties in the platform specification, so I think that has not been done.

docs/pluggable-discovery-specification.md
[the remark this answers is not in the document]
per1234: Looks great!

docs/pluggable-discovery-specification.md
```diff
@@ -0,0 +1,368 @@
...
+recipe. They will communicate to the parent process via stdin/stdout, in particular a discovery will accept commands as
+plain text strings from stdin and will send answers back in JSON format on stdout. Each tool will implement the commands
+to list and enumerate ports for a specific protocol as specified in this document.
```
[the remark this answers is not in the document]
per1234: This change was reverted by https://github.com/arduino/arduino-cli/commit/fb1dde149b745e1c684355e870c60c8e4805d461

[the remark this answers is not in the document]
silvanocerza: Yep, I messed up a bit, will fix right away.

rpc/cc/arduino/cli/commands/v1/upload.proto
silvanocerza: Yeah, the specification is probably the best thing. I'll update the links, hope I'll get them correctly.

docs/pluggable-discovery-specification.md
silvanocerza: Yeah, completely forgot to add that on the platform specification.

docs/pluggable-discovery-specification.md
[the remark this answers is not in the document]
silvanocerza: Fixed: https://github.com/arduino/arduino-cli/commit/a42e2931a5aa3673543969dc27a86ce013524416

docs/pluggable-discovery-specification.md
silvanocerza: Added in https://github.com/arduino/arduino-cli/commit/efdb11c5cb7f7945e47a515f704830ec931f3fc5

rpc/cc/arduino/cli/commands/v1/upload.proto
silvanocerza: Done: https://github.com/arduino/arduino-cli/commit/b59b6f6a43835760ee056d7fedd0c71611c640c7

per1234: Looks great. Thanks!

docs/platform-specification.md
```diff
@@ -430,6 +430,25 @@ following properties are automatically generated:
...
+    uno.upload_port.pid.0=0x0043
+    uno.upload_port.vid.1=0x2341
+    uno.upload_port.pid.1=0x0001
+    uno.upload_port.vid.2=0x2A03
+    uno.upload_port.pid.2=0x0043
+    uno.upload_port.vid.3=0x2341
+    uno.upload_port.pid.3=0x0243
+
+In this case we're using the board's `vid` and `pid` to identify it but `upload_port.*` properties can be anything that
+can help identify a certain board. For more detailed information see the
```
per1234, line 448:
```suggestion
In this case we're using the board's USB VID/PID pair to identify it but `upload_port.*` properties can be anything that
can help identify a certain board. For more detailed information see the
```
The platform properties are named with `vid` and `pid`, but the "vid" and "pid" in this sentence are referring to the USB vendor ID (VID) and product ID (PID), which the reader may not be familiar with. This change will at least provide them with some additional search terms. I wanted to linkify the text when I was working on https://arduino.github.io/arduino-cli/dev/platform-specification/#board-vidpid, but I didn't find a straightforward authoritative explanation to link to.

docs/platform-specification.md
```diff
@@ -430,6 +430,25 @@ following properties are automatically generated:
 - `{build.variant.path}`: The path to the selected board variant folder (inside the
   [variant platform](#platform-terminology), for example hardware/arduino/avr/variants/micro)
 
+If the platform supports pluggable discovery it may also declare a set of `upload_port.*` properties, these properties
+will be used to identify a board by the discovery process when plugged in.
+
+For example we could declare a series of `upload_port.vid` and `upload_port.pid` for the Uno like so:
```
per1234, line 436:
```suggestion
For example we could declare a series of `upload_port.vid` and `upload_port.pid` properties for the Uno like so:
```

docs/pluggable-discovery-specification.md
per1234: Thanks!

docs/pluggable-discovery-specification.md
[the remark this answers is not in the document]
per1234: Looks perfect now!

silvanocerza:
For the record in commit https://github.com/arduino/arduino-cli/pull/1333/commits/d9a3385a216ea8492f539d32b993b40bf497c7f1 I changed how pluggable discoveries must be declare using a `platform.txt` properties. This makes the documentation divert from the [RFC](https://github.com/arduino/tooling-rfcs/blob/main/RFCs/0002-pluggable-discovery.md#discovery-tools-integration).

Previously a platform developer had to declare a pluggable discovery using one of the following syntaxes.

```
    discovery.required=VENDOR_ID:DISCOVERY_NAME
```

```
    discovery.required.0=VENDOR_ID:DISCOVERY_0_NAME
    discovery.required.1=VENDOR_ID:DISCOVERY_1_NAME
```

```
    discovery.DISCOVERY_ID.pattern=DISCOVERY_RECIPE
```

Now they've all been changed to: 

```
    pluggable_discovery.required=VENDOR_ID:DISCOVERY_NAME
```

```
    pluggable_discovery.required.0=VENDOR_ID:DISCOVERY_0_NAME
    pluggable_discovery.required.1=VENDOR_ID:DISCOVERY_1_NAME
```

```
    pluggable_discovery.DISCOVERY_ID.pattern=DISCOVERY_RECIPE
```

So it's not `discovery.*` but `pluggable_discovery.*`.

We took this decision after discovering (pun not intended) that the [Java IDE](https://github.com/arduino/arduino) supports a rudimentary version of the pluggable discovery, it's undocumented and wasn't meant to be used by platforms developers, turns out that some developers figured out how it works and implemented discoveries compatible with the Java IDE. This wouldn't be an issue if they used the same protocol declared in the RFC and a different property in the `platform.txt` but they don't, so to avoid breaking the Java IDE when existing platforms get updated to support the pluggable discovery specification we decided to err on the safe side and change the property used by the Arduino CLI to load a platform's discovery.

So a developer can declare both `discovery.*` and `pluggable_discovery.*` to support both the Java IDE and the Arduino CLI without breaking anything.

commands/upload/upload.go
```diff
@@ -349,21 +426,30 @@ func runProgramAction(pm *packagemanager.PackageManager,
...
-			uploadProperties.Set("serial.port.file", actualPort)
+		uploadProperties.Set("serial.port", actualPort.Address)
+		if actualPort.Protocol == "serial" {
+			// This must be done only for serial ports
+			portFile := strings.TrimPrefix(actualPort.Address, "/dev/")
+			uploadProperties.Set("serial.port.file", portFile)
 		}
 	}
 
+	// Get Port properties gathered using Pluggable Discovery
```
per1234, line 444:
```suggestion
	// Get Port properties gathered using pluggable discovery
```

docs/package_index_json-specification.md
per1234, line 83:
```suggestion
- a [pluggable discovery](pluggable-discovery-specification.md)
```

docs/platform-specification.md
```diff
@@ -625,6 +644,51 @@ The tool configuration properties are available globally without the prefix. For
...
+
+    pluggable_discovery.required=VENDOR_ID:DISCOVERY_NAME
+
+instead if it needs multiple discoveries:
+
+    pluggable_discovery.required.0=VENDOR_ID:DISCOVERY_0_NAME
+    pluggable_discovery.required.1=VENDOR_ID:DISCOVERY_1_NAME
+
+A platform that supports only boards connected via serial ports can easily use the `builtin` package's
+`serial-discovery` without creating a custom Pluggable Discovery:
```
per1234, line 663:
```suggestion
`serial-discovery` without creating a custom pluggable discovery:
```

docs/platform-specification.md
```diff
@@ -625,6 +644,51 @@ The tool configuration properties are available globally without the prefix. For
 property can be used as **{cmd.path}** inside the recipe, and the same happens for all the other avrdude configuration
 variables.
 
+#### Pluggable Discovery
```
per1234, line 647:
```suggestion
#### Pluggable discovery
```
Even though it could be argued that [title case](https://en.wikipedia.org/wiki/Title_case) rules should apply here, most of the document's section headings use sentence case and I think that it's best to follow the established style convention.

docs/platform-specification.md
```diff
@@ -681,9 +767,109 @@ Other upload parameters can also be defined for the board. For example, in the A
...
     [.....]
 
+If necessary the same property can be defined multiple times for different protocols:
+
+    leonardo.upload.serial.maximum_size=28672
+    leonardo.upload.network.maximum_size=256
+
+The two above properties will be available as **{upload.serial.maximum_size}** and **{upload.network.maximum_size}**.
+
+#### Properties from Pluggable Discovery
```
per1234, line 780:
```suggestion
#### Properties from pluggable discovery
```

docs/platform-specification.md
```diff
@@ -681,9 +767,109 @@ Other upload parameters can also be defined for the board. For example, in the A
...
+
+If a platform supports Pluggable Discovery it can also use the port's properties returned by a discovery. For example,
```
per1234, line 782:
```suggestion
If a platform supports pluggable discovery it can also use the port's properties returned by a discovery. For example,
```

docs/UPGRADING.md
```diff
@@ -4,6 +4,66 @@ Here you can find a list of migration guides to handle breaking changes between
...
+  // The possible boards attached to the port.
+  repeated BoardListItem matching_boards = 1;
+  // The port details
+  Port port = 2;
+}
+```
+
+The properties previously contained directly in the message are now stored in the `port` property.
+
+These changes are necessary for the Pluggable Discovery.
```
per1234, line 60:
```suggestion
These changes are necessary for the pluggable discovery.
```

per1234 (changes requested):
Follow-up on the decision from https://github.com/arduino/arduino-cli/pull/1333#discussion_r675041689

docs/pluggable-discovery-specification.md
```diff
@@ -0,0 +1,367 @@
+Discovery tools are a special kind of tool used to find supported boards, a platform developer can create its own
```
per1234, line 1:
```suggestion
Discovery tools are a special kind of tool used to find supported boards, a platform developer can create their own
```
I don't think "its" is appropriate here ([reference](https://en.wikipedia.org/wiki/Gender_neutrality_in_languages_with_gendered_third-person_pronouns#It_and_one_as_gender-neutral_pronouns)).

My suggested replacement is debatable, but I think it's reasonable. See the answer to [this question](https://www.chicagomanualofstyle.org/qanda/data/faq/topics/Pronouns.html#:~:text=the%20singular%20%E2%80%9Cthey%2C%E2%80%9D) for a reasonable take on the subject.

per1234 (changes requested):

silvanocerza:
Rebased to fix conflicts after merge of #1370.

per1234:
@silvanocerza would you like me to submit a PR to the `scerza/pluggable-tools` branch to internationalize all the output strings that are added by this PR?

By the way, I like the `arguments` package :+1:.

silvanocerza:
> @silvanocerza would you like me to submit a PR to the `scerza/pluggable-tools` branch to internationalize all the output strings that are added by this PR?

That would be great.

> By the way, I like the `arguments` package +1.

Yeah, me too. Thank @cmaglie for the idea. :grin:

per1234 (changes requested):

silvanocerza:
Rebased to fix conflicts.

silvanocerza:
Rebased to fix conflicts.

cmaglie (dismissed):

test/test_upload_mock.py
```diff
@@ -138,15 +162,15 @@ def test_upload_sketch(
         "ARDUINO_SKETCHBOOK_DIR": session_data_dir,
     }
 
-    if package_index:
+    # Install everything just once
+    if not os.path.isdir(session_data_dir + "/packages"):
```
silvanocerza, line 1173: Am not sure this works on Windows. :thinking: 

I'd change it to `if not Path(session_data_dir, "packages").is_dir():`.

test/test_upload_mock.py
```diff
@@ -179,8 +1202,16 @@ def test_upload_sketch(
...
     # Fake compilation, we just need the folder to exist
     build_dir = generate_build_dir(sketch_path)
+    if programmer != "":
+        programmer_arg = "-P " + programmer
+    else:
+        programmer_arg = ""
+    if upload_port != "":
+        port_arg = "-p " + upload_port
+    else:
+        port_arg = ""
```
silvanocerza, line 1196: I'd change this to:

```
programmer_arg = f"-P {programmer}" if programmer else ""
port_arg = f"-p {upload_port}" if upload_port else ""
```

test/test_upload_mock.py
```diff
@@ -478,24 +462,24 @@ def generate_expected_output(
...
+            '0x2000; shutdown"\n',
             "win32": '"{data_dir}/packages/arduino/tools/openocd/0.10.0-arduino7/bin/openocd.exe" '
             "-d2 -s "
             '"{data_dir}/packages/arduino/tools/openocd/0.10.0-arduino7/share/openocd/scripts/" '
             "-f "
             '"{data_dir}/packages/arduino/hardware/samd/1.8.11/variants/arduino_zero/openocd_scripts/arduino_zero.cfg" '
             '-c "telnet_port disabled; program '
-            "{{build_dir}/{sketch_name}.ino.bin} verify reset 0x2000; "
-            'shutdown"\n',
+            "{bo}{build_dir}/{sketch_name}.ino.bin{bc} verify reset "
```
silvanocerza, line 481: This is not the best approach I think, you can just use double braces to make `format()` work without failing.
So this string becomes `{{{build_dir}/{sketch_name}.ino.bin}} verify reset ` and you won't need `format(bo="{", bc="}")` anymore.

arduino/cores/packagemanager/loader.go
```diff
@@ -356,9 +356,62 @@ func (pm *PackageManager) loadPlatformRelease(platform *cores.PlatformRelease, p
...
+
+		// Merge the converted properties in the root configuration
+		platform.Properties.Merge(convertedProps)
+
+		// Add the network upload to the boards using the old method
+		for _, board := range platform.Boards {
+			oldUploadTool := board.Properties.Get("upload.tool")
+			if oldUploadTool == toolName && !board.Properties.ContainsKey("upload.tool.network") {
+				board.Properties.Set("upload.tool.network", convertedToolName)
+				// fmt.Printf("ADDED: %s.upload.tool.network=%s\n", board, convertedToolName)
```
silvanocerza, line 384: Remove this.

arduino/cores/packagemanager/loader.go
```diff
@@ -356,9 +356,62 @@ func (pm *PackageManager) loadPlatformRelease(platform *cores.PlatformRelease, p
...
+		pattern = strings.ReplaceAll(pattern, "{network.password}", "{upload.field.password}")
+	}
+	props.Set("upload.pattern", pattern)
+
+	prefix := "tools." + newToolName + "."
+	res := properties.NewMap()
+	for _, k := range props.Keys() {
+		v := props.Get(k)
+		res.Set(prefix+k, v)
+		// fmt.Println("ADDED:", prefix+k+"="+v)
```
silvanocerza, line 421: Remove this.

arduino/cores/packagemanager/loader.go
```diff
@@ -356,9 +356,62 @@ func (pm *PackageManager) loadPlatformRelease(platform *cores.PlatformRelease, p
 		return fmt.Errorf(tr("loading boards: %s"), err)
 	}
 
+	if !platform.PluggableDiscoveryAware {
+		convertLegacyPlatformToPluggableDiscovery(platform)
+	}
```
silvanocerza, line 361: I'm strongly against this. 

This is not being retrocompatible, we never were compatible with those platforms, this is adding a new feature to the CLI that will remain here forever and will need to be maintained.

Platform maintainers will keep using the old syntax if we do this, that syntax is not even documented. We have a new specification to support network upload and we must push platform maintaners to use it, if we keep supporting old things it will be harder and harder to move forward with newer ones.

cmaglie: IMHO we are not creating legacy but supporting the existing ecosystem. If we remove this function the network upload, already supported by the Java IDE, will not work with the Arduino CLI on all the currently existing platforms (at least until the 3rd party developers adopt the pluggable discovery and make a new upstream release).

IMHO we should start by supporting the current platforms and leave the possibility to the upstream maintainers to upgrade at their own pace. They will eventually upgrade to pluggable discovery because it's a clearly superior specification, but until then the users will benefit from the network upload with all the currently available platforms.

test/test_upload_mock.py
cmaglie: it works in Windows, but your suggestion is better 👍🏼

arduino/cores/packagemanager/loader.go
silvanocerza: I don't think it's a good idea but since @cmaglie and I couldn't settle this debate @rsora did it for us and we're going forward with the conversion.

Nonetheless I'd like some unit tests for the `convertLegacyPlatformToPluggableDiscovery` function.

cmaglie (approved):

silvanocerza:
🎉

rsora:
image

matthijskooijman:
I noticed a regression that I expect is caused by this PR. When using arduino-cli from git now, uploads using (our slightly customized version of) the [STM32 core](https://github.com/stm32duino/Arduino_Core_STM32) fail with:

    Error during Upload: Property 'upload.tool.serial' is undefined


The board in question indeed does not define `upload.tool.serial`, but still uses the pre-pluggable syntax of defining `upload.tool`, which should be translated to `upload.tool.default` by `convertUploadToolsToPluggableDiscovery()`. However, this core does not define `upload.tool` on the board directly, but defines it in a board menu: https://github.com/stm32duino/Arduino_Core_STM32/blob/974b35690cbf4680ce5746c61c69fb20935bae2f/boards.txt#L522

To reproduce:

```
$ arduino-cli compile --fqbn STMicroelectronics:stm32:Nucleo_32:pnum=NUCLEO_F031K6,upload_method=serialMethod --upload --port /dev/ttyACM0 
Sketch uses 7612 bytes (23%) of program storage space. Maximum is 32768 bytes.
Global variables use 828 bytes (20%) of dynamic memory, leaving 3268 bytes for local variables. Maximum is 4096 bytes.
Error during Upload: Property 'upload.tool.serial' is undefined
```

I guess this is fairly easy to fix in the core, though (just define both `upload.tool.default` (or `upload.tool.<protocol>`) in addition to the legacy `upload.tool`, but it might be useful to fix compatibility here too. An obvious fix would be to implement the fallback when actually doing the upload, rather than when loading the boards).

@fpistm, you might find this interesting too :-)

cmaglie:
@matthijskooijman you're my nightmare :-)

> An obvious fix would be to implement the fallback when actually doing the upload, rather than when loading the boards).

yes I think it's the thing to do, I'll move your comment in a new issue, so we can keep track of it.

fpistm:
> 
> 
> @matthijskooijman you're my nightmare :-)
> 

@cmaglie  Mine too 😜

matthijskooijman:
Lol, thanks for the compliments :-p

PR #1363, opened 2021-07-20
Applied taskfile lint rules to DistTask.yml

cmaglie:
- moved vars section to top https://taskfile.dev/#/styleguide?id=use-the-correct-order-of-keywords
- removed wrapping spaces in variables https://taskfile.dev/#/styleguide?id=dont-wrap-vars-in-spaces-when-templating

### DIFF
--- a/DistTasks.yml
+++ b/DistTasks.yml
@@ -16,6 +16,11 @@ version: "3"
 #
 # The project MUST contain a LICENSE.txt file in the root folder or packaging will fail.
 
+vars:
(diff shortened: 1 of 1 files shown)


per1234 (approved):
I ran a test release in my fork using this taskfile because I'm paranoid like that and all is well.

Thanks Cristian!

cmaglie:
Well, this is just linting, we already changed the taskfile some times ago :-)
Thanks for testing it!

PR #1416, opened 2021-08-26
Add ways to let users verify if new CLI released

silvanocerza:

      before creating one)
- [x] The PR follows

Adds a new feature and change behaviour of an existing command.

Users are never notified of new releases of the Arduino CLI.

If more than 24 hours have passed since the last check the Arduino CLI verifies if the currently executed version is the latest available, if not it notifies the user that a new release is available. e.g.

```
$ arduino-cli core list            
ID                  Installed Latest Name                                        
arduino:samd        1.8.10    1.8.11 Arduino SAMD Boards (32-bits ARM Cortex-M0+)
esp32:esp32         1.0.6     1.0.6  esp32                                       
SPRESENSE:spresense 2.0.2     2.2.1  Spresense Reference Board                   



A new release of arduino-cli is available: 0.18.0 → 0.18.3
https://arduino.github.io/arduino-cli/latest/installation/#latest-packages

$ arduino-cli version
arduino-cli alpha Version: 0.18.0 Commit: 6507e27c Date: 2021-08-26T13:57:48Z


A new release of arduino-cli is available: 0.18.0 → 0.18.3
https://arduino.github.io/arduino-cli/latest/installation/#latest-packages

$ arduino-cli version --format json
{
  "Application": "arduino-cli",
  "VersionString": "0.18.0",
  "LatestVersion": "0.18.3",
  "Commit": "6507e27c",
  "Status": "alpha",
  "Date": "2021-08-26T13:57:48Z"
}
```

The message is printed on the `stderr` output so it won't disrupt existing parsers and the JSON output.

This behaviour you can disable it by setting the `updater.disable_notification` config or the env var `ARDUINO_UPDATER_DISABLE_NOTIFICATION` to `true`.

`arduino-cli version` command will always force a check for new versions even if the above config or env var are set.

Nope.

* **Other information**:

Heavily inspired by https://github.com/cli/cli.

---


cli/updater/updater.go
```diff
@@ -0,0 +1,138 @@
...
+	return latestVersion
+}
+
+// NotifyNewVersionIsAvailable prints information about the new latestVersion
+func NotifyNewVersionIsAvailable(latestVersion string) {
+	feedback.Errorf("\n\n%s %s → %s\n%s",
+		ansi.Color(tr("A new release of Arduino CLI is available:"), "yellow"),
+		ansi.Color(globals.VersionInfo.VersionString, "cyan"),
+		ansi.Color(latestVersion, "cyan"),
+		ansi.Color("https://arduino.github.io/arduino-cli/latest/installation/#latest-packages", "yellow"))
```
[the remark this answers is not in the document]
per1234: `github.com/mattn/go-colorable` is already being used for the verbose output:
https://github.com/arduino/arduino-cli/blob/b8d30dcd169861019f7b5a20916b5521d84867d5/cli/cli.go#L152

Looks like it is proposed that it be used for all the non-verbose output as well:
https://github.com/arduino/arduino-cli/pull/1251/files#diff-ef44b92f1ba3916c39c98fc4a25b67da847cb2a07a5d1253e1978365338888a6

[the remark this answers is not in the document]
silvanocerza: I partly fixed this in a separate PR #1418, I'll also need to remove this `ansi` package and use the `color` one that's already used in other parts of the CLI.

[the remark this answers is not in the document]
silvanocerza: Fixed, now works fine now: 
image

cli/version/version.go
```diff
@@ -39,5 +43,29 @@ func NewCommand() *cobra.Command {
 }
 
 func run(cmd *cobra.Command, args []string) {
-	feedback.Print(globals.VersionInfo)
+	if strings.Contains(globals.VersionInfo.VersionString, "git-snapshot") {
```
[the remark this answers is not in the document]
silvanocerza: Fixed with ca4c678110a68acbb595430609e87cccdcfe46e2.

cli/updater/updater.go
```diff
@@ -0,0 +1,138 @@
...
+func getLatestRelease() string {
+	client, err := httpclient.New()
+	if err != nil {
+		return ""
+	}
+
+	// We just use this URL to check if there's a new release available and
+	// never show it to the user, so it's fine to use the Linux one for all OSs.
+	URL := "https://downloads.arduino.cc/arduino-cli/arduino-cli_latest_Linux_64bit.tar.gz"
+	res, err := client.Get(URL)
```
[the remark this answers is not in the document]
silvanocerza: Fixed with 2991e9fe84ed2fd71a5cc98ccb1cb241abfcdd6f.

docs/installation.md
```diff
@@ -40,6 +40,10 @@ as a parameter like this:
 curl -fsSL https://raw.githubusercontent.com/arduino/arduino-cli/master/install.sh | sh -s 0.9.0
 ```
 
+The Arduino CLI verifies every 24 hours if there are new releases, if you don't like this behaviour you can disable it
+by setting the [`updater.enable_notification` config](configuration.md#configuration-keys) or the
```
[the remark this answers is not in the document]
silvanocerza: Fixed with fd1bc798b49510e641b2b19d00acd5249178b0e6.

cli/cli.go
```diff
@@ -151,6 +155,20 @@ func preRun(cmd *cobra.Command, args []string) {
...
 	feedback.SetErr(colorable.NewColorableStderr())
 
+	updaterMessageChan = make(chan *semver.Version)
+	go func() {
+		if cmd.Name() == "version" {
+			// The version command checks by itself if there's a new available version
+			updaterMessageChan <- nil
+		}
+		// Starts checking for updates
+		currentVersion, err := semver.Parse(globals.VersionInfo.VersionString)
```
cmaglie, line 165: Is VersionString supposed to be always semver compliant? if yes you can use `semver.MustParse` to avoid checking the error

cli/updater/updater.go
```diff
@@ -0,0 +1,138 @@
...
+// based on https://github.com/watson/ci-info/blob/HEAD/index.js
+func isCI() bool {
+	return os.Getenv("CI") != "" || // GitHub Actions, Travis CI, CircleCI, Cirrus CI, GitLab CI, AppVeyor, CodeShip, dsari
+		os.Getenv("BUILD_NUMBER") != "" || // Jenkins, TeamCity
+		os.Getenv("RUN_ID") != "" // TaskCluster, dsari
+}
+
+// getLatestRelease queries the official Arduino download server for the latest release,
+// if there are no errors or issues a version string is returned, in all other case an empty string.
+func getLatestRelease() string {
```
cmaglie, line 111: We can make this function return a `*semver.Version` (or `nil` if fails or not valid semver)

cli/updater/updater.go
```diff
@@ -0,0 +1,138 @@
...
+	return latestVersion
+}
+
+// NotifyNewVersionIsAvailable prints information about the new latestVersion
+func NotifyNewVersionIsAvailable(latestVersion string) {
+	feedback.Errorf("\n\n%s %s → %s\n%s",
+		color.YellowString(tr("A new release of Arduino CLI is available:")),
+		color.CyanString(globals.VersionInfo.VersionString),
+		color.CyanString(latestVersion),
+		color.YellowString("https://arduino.github.io/arduino-cli/latest/installation/#latest-packages"))
```
cmaglie, line 77: Is this URL going to stay? Will it move at some point from `arduino.github.io` to `arduino.cc/something`?

cli/version/version.go
```diff
@@ -39,5 +43,29 @@ func NewCommand() *cobra.Command {
 }
 
 func run(cmd *cobra.Command, args []string) {
-	feedback.Print(globals.VersionInfo)
+	if strings.Contains(globals.VersionInfo.VersionString, "git-snapshot") || strings.Contains(globals.VersionInfo.VersionString, "nightly") {
+		// We're using a development version, no need to check if there's a
+		// new release available
+		feedback.Print(globals.VersionInfo)
+		return
+	}
```
cmaglie, line 51: I don't know if this is sensible here... I'd leave the `version` command to always check for the latest version (even in dev builds)

```suggestion
```

cli/updater/updater.go
```diff
@@ -0,0 +1,138 @@
...
+	return checkForUpdate(currentVersion)
+}
+
+// ForceCheckForUpdate always returns the latest available version if greater than
+// the one running, nil in all other cases
+func ForceCheckForUpdate(currentVersion *semver.Version) *semver.Version {
+	return checkForUpdate(currentVersion)
+}
+
+func checkForUpdate(currentVersion *semver.Version) *semver.Version {
```
cmaglie, line 51:
```suggestion
	return ForceCheckForUpdate(currentVersion)
}

// ForceCheckForUpdate always returns the latest available version if greater than
// the one running, nil in all other cases
func ForceCheckForUpdate(currentVersion *semver.Version) *semver.Version {
```

cmaglie (changes requested):

cli/cli.go
silvanocerza: It's not always semver compliant, it can also be `git-snapshot` or `nightly-<timestamp>` so I can't use `semver.MustParse`.

cli/updater/updater.go
silvanocerza: Ideally we'll change it in the future when the Arduino CLI will be added to https://www.arduino.cc/en/software.

cli/version/version.go
silvanocerza: I can't, it would fail because `globals.VersionInfo.VersionString` can't be parsed as semver.

cli/updater/updater.go
silvanocerza: I don't think that would help that much, this `getLatestRelease` is only used in `checkForUpdate` and there we handle eventual errors in parsing the version string, if I change `getLatestRelease` to return a `*semver.Version` I'd have to check for the parsing error inside it and handle the `nil` version in `checkForUpdate` too. 

Don't seem much of an improvement to me really. 🤔

cli/updater/updater.go
silvanocerza: Right!

cli/cli.go
cmaglie: Shouldn't those versions be changed to `0.0.0-git` or `0.0.0-nightly-<timestamp>`?
These non-semver versions plugged in feels "wrong" to me...

Also we already had:

```go
var (
	defaultVersionString = "0.0.0-git"
	versionString        = ""
	commit               = ""
	status               = "alpha"
	date                 = ""
	tr                   = i18n.Tr
)
```
in `version.go` IMHO we should follow the same pattern and force semver compliance.

per1234 (approved):

cli/updater/updater.go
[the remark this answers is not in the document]
per1234: Looks great Silvano! I think it's important to give some care to the Windows users, since my observations indicate it is by far the most common OS for Arduino users in general (I don't know how much that shifts when it comes to CLI users specifically though).

cli/cli.go
silvanocerza: I get what you mean but probably would be better if done in a separate PR.

cmaglie (approved):

PR #1424, opened 2021-08-31
Better handling of functions returing multiple errors

cmaglie:
This PR replaces `[]error` with a `*multierror.Error`.

* the returned `multierror.Error` is compatible with the idiomatic `error`
* makes the code much more readable
* the error output is formatted better
* it's `Unwrap`-able

No breaking changes.


cmaglie:
This change may lead to some weird and hard-to-find bugs due to `if err != nil` checks failing (unfortunately [not all `nil`s are equal in go](https://www.calhoun.io/when-nil-isnt-equal-to-nil/)...).

Rejecting because the advantages are not worth the side effects.

PR #1425, opened 2021-09-01
Removed unnecessary `fmt.*printf` before `tr(...)`

cmaglie:
* Removed some `fmt.Sprintf` since `tr` already provides formatting in printf-style.
* Simplified some strings.
* Added some missing line feed.

* Fixed [static initialization problem on some strings](https://github.com/arduino/arduino-cli/pull/1425#discussion_r701014277)


cli/debug/debug.go
```diff
@@ -62,7 +61,7 @@ func NewCommand() *cobra.Command {
 	debugCommand.Flags().StringVarP(&fqbn, "fqbn", "b", "", tr("Fully Qualified Board Name, e.g.: arduino:avr:uno"))
 	port.AddToCommand(debugCommand)
 	debugCommand.Flags().StringVarP(&programmer, "programmer", "P", "", tr("Programmer to use for debugging"))
-	debugCommand.Flags().StringVar(&interpreter, "interpreter", "console", fmt.Sprintf(tr("Debug interpreter e.g.: %s, %s, %s, %s, %s"), "console", "mi", "mi1", "mi2", "mi3"))
+	debugCommand.Flags().StringVar(&interpreter, "interpreter", "console", tr("Debug interpreter e.g.: %s", "console, mi, mi1, mi2, mi3"))
```
per1234, line 64: The reason I intentionally chose not to use this capability when I did the i18n campaign is because I think it makes the code unclear regarding which components will be internationalized. I could accept it if the function was somehow using the extra arguments to enhance the translation effort, but it's not. My feeling is that the `i18n.Tr` function should have one job: internationalization of strings, just as `fmt.Sprintf` has its one job.

But I do understand that my approach resulted in quite a bit of extra code that is not really necessary, even more so due to the technique of substituting out parts of the strings that should not be translated. It wasn't clear exactly how much the impact of my decision would be when I made it at the start of the work. So I think it's a reasonable change to make. I just wanted to share my decision process.

cmaglie: Yep, I figured it went like that :-)

legacy/builder/constants/constants.go
```diff
@@ -106,10 +104,10 @@ var MSG_LIB_LEGACY = tr("(legacy)")
 var MSG_LIBRARIES_MULTIPLE_LIBS_FOUND_FOR = tr("Multiple libraries were found for \"{0}\"")
 var MSG_LIBRARIES_NOT_USED = tr(" Not used: {0}")
 var MSG_LIBRARIES_USED = tr(" Used: {0}")
-var MSG_LIBRARY_CAN_USE_SRC_AND_UTILITY_FOLDERS = fmt.Sprintf(tr("Library can't use both '%[1]s' and '%[2]s' folders. Double check {0}"), "src", "utility")
+var MSG_LIBRARY_CAN_USE_SRC_AND_UTILITY_FOLDERS = tr("Library can't use both '%[1]s' and '%[2]s' folders. Double check {0}", "src", "utility")
```
cmaglie, line 107: Here we have these "mixed" messages with the gettext-style `{0}` markers and the printf-style `%[1]s` markers.

AFAIR this was done at the times to let the Java IDE translate the messages.
Basically in the `arduino-builder`, using the legacy `MachineLogger`, the messages are passed as-is (i.e. with the markers `{0}` left intact) together with the value to substitute. It's the duty of the Java IDE to translate the message and replace the markers with the given values.

```
$ arduino-builder -compile -logger=machine -hardware ... -tools ... -libraries ... -fqbn=arduino:avr:uno -ide-version=10815 -build-path /tmp/arduino_build_474903 -verbose /tmp/sketch_sep02a/sketch_sep02a.ino
[.....]
===info ||| Progress {0} ||| [25.00]
===info ||| Progress {0} ||| [50.00]
===info ||| Progress {0} ||| [75.00]
===info ||| Sketch uses {0} bytes ({2}%%) of program storage space. Maximum is {1} bytes. ||| [444 32256 1]
===info ||| Global variables use {0} bytes ({2}%%) of dynamic memory, leaving {3} bytes for local variables. Maximum is {1} bytes. ||| [9 2048 0 2039]
===info ||| Progress {0} ||| [100.00]
```

This must be completely replaced and the translation should be made in the CLI.
I see also that if I use the `HumanLogger` with the parameter `-logger=human` then the output is:
```
$ arduino-builder -compile -logger=human -hardware ... -tools ... -libraries ... -fqbn=arduino:avr:uno -ide-version=10815 -build-path /tmp/arduino_build_474903 -verbose /tmp/sketch_sep02a/sketch_sep02a.ino
```

so the HumanLogger is able to replace the `{0}`, at this point I'm wondering if we should keep the `%s` or use the gettext notation.

per1234: Yeah, these are very ugly. I remembered this comment, so didn't worry about them too much because I figured they were going to go away at some point:
https://github.com/arduino/arduino-cli/pull/961#discussion_r512819675

> I'm wondering if we should keep the %s or use the gettext notation.

I think people contributing to the code will be more likely to be familiar with the printf style. I don't know which would be friendly to the translators though. Something we noticed is that Transifex recognizes some of these printf verbs and makes them into a token that is treated like a single character in the editor. But it doesn't recognize all of them. For example, here you can see it gave the `%s` special treatment, but not the `%v`:
Clipboard01
But I see that it should be possible to configure Transifex to recognize whatever syntax we use for placeholders:
https://docs.transifex.com/translation-checks/setting-translation-checks
So this means that it's probably not so important that the translator be familiar with the exact placeholder syntax in use, just as long as they can recognize the significance of the index numbers in the placeholders.

---
FYI, these particular strings are not correctly handled by the `i18n` package because they are declared at package level before the call to `i18n.Init()`, and so are just returned as-is. Details [here](https://arduino.atlassian.net/browse/ATL-1601).

cmaglie: Ok it seems that I can't put the panic in the Tr function otherwise the unit-tests will fail:
```
=== RUN   TestBoardOptions
--- FAIL: TestBoardOptions (0.00s)
panic: i18n.Tr called before i18n.Init() [recovered]
	panic: i18n.Tr called before i18n.Init()

goroutine 9 [running]:
testing.tRunner.func1.2(0xc56e00, 0xeaee20)
	/opt/hostedtoolcache/go/1.16.7/x64/src/testing/testing.go:1143 +0x332
testing.tRunner.func1(0xc0002a2f00)
	/opt/hostedtoolcache/go/1.16.7/x64/src/testing/testing.go:1146 +0x4b6
panic(0xc56e00, 0xeaee20)
	/opt/hostedtoolcache/go/1.16.7/x64/src/runtime/panic.go:965 +0x1b9
github.com/arduino/arduino-cli/i18n.Tr(0xdceacd, 0x28, 0x0, 0x0, 0x0, 0x0, 0x1)
	/home/runner/work/arduino-cli/arduino-cli/i18n/i18n.go:54 +0xd9
github.com/arduino/arduino-cli/arduino/cores.(*Board).GetBuildProperties(0x13bf060, 0xc0003b6150, 0xc000339400, 0x0, 0x0)
	/home/runner/work/arduino-cli/arduino-cli/arduino/cores/board.go:109 +0x69d
github.com/arduino/arduino-cli/arduino/cores.(*Board).GeneratePropertiesForConfiguration(0x13bf060, 0xdb5f54, 0xe, 0xc000309ef0, 0xc911e0, 0xc0003b6000)
	/home/runner/work/arduino-cli/arduino-cli/arduino/cores/board.go:141 +0x28f
github.com/arduino/arduino-cli/arduino/cores.TestBoardOptions(0xc0002a2f00)
	/home/runner/work/arduino-cli/arduino-cli/arduino/cores/board_test.go:298 +0x4bd
testing.tRunner(0xc0002a2f00, 0xe052a8)
	/opt/hostedtoolcache/go/1.16.7/x64/src/testing/testing.go:1193 +0xef
created by testing.(*T).Run
	/opt/hostedtoolcache/go/1.16.7/x64/src/testing/testing.go:1238 +0x2b3
FAIL	github.com/arduino/arduino-cli/arduino/cores	0.021s
=== RUN   TestIndexParsing
--- FAIL: TestIndexParsing (0.00s)
panic: i18n.Tr called before i18n.Init() [recovered]
	panic: i18n.Tr called before i18n.Init()
```

so I'll remove it for now, we need to come up with a better idea.
BTW it has been very helpful to find all occurrences of the bug, very neat!

commands/upload/upload.go
```diff
@@ -530,7 +526,7 @@ func determineBuildPathAndSketchName(importFile, importDir string, sk *sketch.Sk
 	// Case 1: importFile flag has been specified
 	if importFile != "" {
 		if importDir != "" {
-			return nil, "", fmt.Errorf(fmt.Sprintf(tr("%s and %s cannot be used together"), "importFile", "importDir"))
+			return nil, "", fmt.Errorf(tr("%[1]s and %[2]s cannot be used together", "importFile", "importDir"))
```
per1234, line 529:
```suggestion
			return nil, "", fmt.Errorf(tr("%s and %s cannot be used together", "importFile", "importDir"))
```
See explanation above.

legacy/builder/resolve_library.go
```diff
@@ -31,9 +31,9 @@ func ResolveLibrary(ctx *types.Context, header string) *libraries.Library {
 
 	logger := ctx.GetLogger()
 	if ctx.Verbose {
-		logger.Println(constants.LOG_LEVEL_INFO, fmt.Sprintf(tr("Alternatives for %[1]s: %[2]s"), header, candidates))
+		logger.Println(constants.LOG_LEVEL_INFO, tr("Alternatives for %s:", header), candidates)
```
per1234, line 34:
```suggestion
		logger.Println(constants.LOG_LEVEL_INFO, tr("Alternatives for %[1]s: %[2]s", header, candidates))
```
If I managed to follow the code correctly, I think this didn't work as intended because `logger.Println()` eventually passes the `format` and `a...` arguments through `fmt.Sprintf()`, so you get something like `Alternatives for SomeHeader.h:%!(EXTRA string=some candidates)`.

legacy/builder/wipeout_build_path_if_build_options_changed.go
```diff
@@ -46,7 +46,7 @@ func (s *WipeoutBuildPathIfBuildOptionsChanged) Run(ctx *types.Context) error {
 
 	var prevOpts *properties.Map
 	if err := json.Unmarshal([]byte(previousBuildOptionsJson), &prevOpts); err != nil || prevOpts == nil {
-		ctx.GetLogger().Println(constants.LOG_LEVEL_DEBUG, constants.MSG_BUILD_OPTIONS_INVALID+constants.MSG_REBUILD_ALL, constants.BUILD_OPTIONS_FILE)
+		ctx.GetLogger().Println(constants.LOG_LEVEL_DEBUG, ", "+tr("rebuilding all"), constants.BUILD_OPTIONS_FILE)
```
per1234, line 49:
```suggestion
		ctx.GetLogger().Println(constants.LOG_LEVEL_DEBUG, tr("{0} invalid, rebuilding all"), constants.BUILD_OPTIONS_FILE)
```

per1234 (changes requested):

cli/core/list.go
```diff
@@ -85,7 +85,7 @@ func (ir installedResult) String() string {
 	for _, p := range ir.platforms {
 		name := p.Name
 		if p.Deprecated {
-			name = fmt.Sprintf(tr("[DEPRECATED] %s"), name)
+			name = fmt.Sprintf("[%s] %s", tr("DEPRECATED"), name)
```
silvanocerza, line 88: Am not sure about this, won't this cause some issues for languages that are read from right to left?

cli/core/search.go
```diff
@@ -115,7 +115,7 @@ func (sr searchResults) String() string {
 		for _, item := range sr.platforms {
 			name := item.GetName()
 			if item.Deprecated {
-				name = fmt.Sprintf(tr("[DEPRECATED] %s"), name)
+				name = fmt.Sprintf("[%s] %s", tr("DEPRECATED"), name)
```
silvanocerza, line 118: Same as above.

legacy/builder/print_used_and_not_used_libraries.go
```diff
@@ -47,10 +47,10 @@ func (s *PrintUsedAndNotUsedLibraries) Run(ctx *types.Context) error {
 		if len(libResResult.NotUsedLibraries) == 0 {
 			continue
 		}
-		logger.Fprintln(os.Stdout, logLevel, constants.MSG_LIBRARIES_MULTIPLE_LIBS_FOUND_FOR, header)
-		logger.Fprintln(os.Stdout, logLevel, constants.MSG_LIBRARIES_USED, libResResult.Library.InstallDir)
+		logger.Fprintln(os.Stdout, logLevel, tr("Multiple libraries were found for \"{0}\""), header)
+		logger.Fprintln(os.Stdout, logLevel, " "+tr("Used: {0}"), libResResult.Library.InstallDir)
```
silvanocerza, line 51: I'd keep the space in the string, for right to left languages it should be on the other side I guess?

```suggestion
		logger.Fprintln(os.Stdout, logLevel, tr(" Used: {0}"), libResResult.Library.InstallDir)
```

legacy/builder/print_used_and_not_used_libraries.go
```diff
@@ -47,10 +47,10 @@ func (s *PrintUsedAndNotUsedLibraries) Run(ctx *types.Context) error {
...
 		for _, notUsedLibrary := range libResResult.NotUsedLibraries {
-			logger.Fprintln(os.Stdout, logLevel, constants.MSG_LIBRARIES_NOT_USED, notUsedLibrary.InstallDir)
+			logger.Fprintln(os.Stdout, logLevel, " "+tr("Not used: {0}"), notUsedLibrary.InstallDir)
```
silvanocerza, line 53: Same as above.

```suggestion
			logger.Fprintln(os.Stdout, logLevel, tr(" Not used: {0}"), notUsedLibrary.InstallDir)
```

silvanocerza (changes requested):
I have some concerns about certain strings when they'll be translated for right to left languages.

legacy/builder/print_used_and_not_used_libraries.go
cmaglie: My main concern here is that the translator will not even notice it and will just translate the string dropping the space. This is further confirmed by looking at the Transifex GUI (in this case on the Italian language but I bet that it's the same for many others):

image

as you can see the translated string has lost the whitespace.
IMHO we should keep formatting tricks out of the translated strings and add the space on the left or the right depending on the locale. But this implies that we need to handle the right-to-left support in a more profound way.

silvanocerza: Yeah, I agree. Probably we should try check the [golang/x/text](https://pkg.go.dev/golang.org/x/text) package and see if it helps.

All in all we should also rework how we output in the CLI.

per1234 (approved):

cli/core/list.go
cmaglie: I've read some documents about proper RTL support (links below) to make my mind.

I don't think putting this string in the translation helps either: if the translator makes a good job, the string will appear reversed (`%s [إهمال]` that is correct) but still left-aligned (that is incorrect).

I see the `[DEPRECATED]` piece more like a "label" widget that is placed near the name than a part of the sentence.

The table that comes out from the `core list`:

```
ID                       Installed Latest Name                                                                                 
arduino:avr              1.8.3     1.8.3  Arduino AVR Boards                                                                   
arduino:mbed             2.1.0     2.4.1  [DEPRECATED] [DEPRECATED - Please install standalone packages] Arduino Mbed OS Boards
```

to properly render it in RTL we should do it from right to left like this:

```
                                                                            Name Latest Installed           ID
                                                              Arduino AVR Boards  1.8.3     1.8.3  arduino:avr
[DEPRECATED - Please install standalone packages] Arduino Mbed OS Boards [إهمال]  2.4.1     2.1.0 arduino:mbed
```

But this requires implementing RTL support on the table renderer (and in this case the `[DEPRECATED]` label should be laid out accordingly to flow direction, but again, not a translator duty).

Until we correctly support BiDi I think it's better to keep layouts out of the translations.

https://terminal-wg.pages.freedesktop.org/bidi/
https://terminal-wg.pages.freedesktop.org/bidi/bidi-intro/why-terminals-are-special.html
https://rtl.wtf/wild/mixed-content-terminal/#what-s-going-on

silvanocerza: Let's wait for that then ok. 👍

silvanocerza (approved):

PR #1449, opened 2021-09-13
Added parsing of pluggable monitors in package_index.json

cmaglie:

Added functions for parsing pluggable monitor fields in package_index.json.
The added functions are sort of duplicated from the ones that handle pluggable discoveries, I have another PR to factor out the common part but is a bit more involved, so for now I'll push this one so we can move forward.

No


arduino/cores/status.go
```diff
@@ -127,6 +127,33 @@ func (packages Packages) GetPlatformReleaseDiscoveryDependencies(release *Platfo
...
+
+		// We always want to use the latest available release for monitors
+		latestRelease := tool.LatestRelease()
+		if latestRelease == nil {
+			return nil, fmt.Errorf(tr("can't find latest release of %s"), monitor.Name)
+		}
+		res = append(res, latestRelease)
+	}
+	return res, nil
+}
```
silvanocerza, line 155: This is identical to `GetPlatformReleaseDiscoveryDependencies` apart from line 137, we could make generic.

silvanocerza (changes requested):
Small change to make, rest is good.

arduino/cores/status.go
cmaglie: Yes, also `MonitorDependencies` is identical to `DiscoveryDependencies`.

The problem is that `release.MonitorDependencies` is an array of `MonitorDependecy`... we can make it generic if we change the type of `release.DiscoveryDependencies` and `release.MonitorDependencies` to be the same type (for example reusing`ToolDependencies` for both) but this change starts a chain reaction... I've another branch ready for that but I prefer to push it in another PR for review and not delay this one.

silvanocerza: Oh right right, the type is different. Let's leave as it is now for the time being.

silvanocerza (approved):

PR #1482, opened 2021-09-30
Pluggable monitor fix and refactor

cmaglie:

      before creating one)
- [x] The PR follows

Fixed a race condition in the pluggable monitor client due to a stateLock contention:
  1. many methods (for example `Close`) had the stateLock held for the entire scope of the function
  2. at the same time, the `"port_closed"` message can be received at any moment in the decode loop, and it requires a stateLock as well, but it may be blocked until the `Close` function get the `close` message (that may happen later in the decodeLoop)

  In this case the worst case scenario is that the decode loop is blocked for 10 seconds until the timeout occurs and `Close` exits, but this is not ideal. This bug has been fixed by removing the `state` from the pluggable monitor client, since it's not really useful in any case.

No

* **Other information**:
Some other refactorings have been made in the same PR to simplify and make error handling and logging more coherent.


arduino/monitor/monitor.go
```diff
@@ -116,54 +113,54 @@ func (mon *PluggableMonitor) String() string {
...
+	}
+	if expectedEvt == "" {
+		// No message processing required for this call
+		return msg, nil
+	}
+	if msg.EventType != expectedEvt {
+		return msg, fmt.Errorf(tr("communication out of sync, expected '%[1]s', received '%[2]s'"), expectedEvt, msg.EventType)
 	}
+	if msg.Message != "OK" || msg.Error {
+		return msg, fmt.Errorf(tr("command '%[1]s' failed: %[1]s"), expectedEvt, msg.Message)
```
silvanocerza, line 157: Is this error message correct? :thinking:

cmaglie: 🤦🏼

arduino/monitor/monitor.go
```diff
@@ -321,58 +286,34 @@ func (mon *PluggableMonitor) Open(port *rpc.Port) (io.ReadWriter, error) {
...
-		return fmt.Errorf(tr("calling %[1]s: %[2]w"), "QUIT", err)
-	} else if msg.EventType != "quit" {
-		return errors.Errorf(tr("communication out of sync, expected 'quit', received '%s'"), msg.EventType)
-	} else if msg.Message != "OK" || msg.Error {
-		return errors.Errorf(tr("command failed: %s"), msg.Message)
+	if _, err := mon.waitMessage(time.Second*10, "quit"); err != nil {
+		return err
 	}
-	mon.killProcess()
+	_ = mon.killProcess()
```
silvanocerza, line 317: We could at least log the error here instead of completely ignoring it.

silvanocerza (approved):

Issue #1456, opened 2021-09-16
Q: how to create a new sketch?


4ntoine:
### Current behavior

No `CreateSketch` method in `commands.proto`, but there are `LoadSketch` and `ArchiveSketch`.

This is available with cmd `arduino-cli sketch new ...` so i guess it should be available via gRPC too. 

### Expected behavior

`CreateSketch` method in `commands.proto`.

### Environment

- CLI version (output of `arduino-cli version`): 0.19.0
- OS and platform:

PR #1498, opened 2021-10-06
Update gRPC API to create a new sketch

4ntoine:

      before creating one)
- [x] The PR follows

Updates gRPC (daemon) API to support creating a new sketch

Not implemented (https://github.com/arduino/arduino-cli/issues/1456)

Implemented and available via gRPC

No breaking changes

* **Other information**:
Still requires some polishing (fix i18n) - at least 1 string needs to be translated (error message).
No tests were added in the repo, but i tested it with Dart gRPC client manually instead.

---


4ntoine:
No idea on what's wrong with "Check internationalization" task..

4ntoine:
Yeah, originally i started with this design and then i decided that the
caller does not have to be that smart and specify the path per every call
but that receive the path instead. Also if we design it to accept the path
every time we will be unable to do 'sketch list' as they can be in
different dirs every time. So i'd prefer to have a single dir for them as a
setting.

What do you think?

вт, 12 окт. 2021 г., 17:46 Silvano Cerza ***@***.***>:

4ntoine:
Rebased to upstream master, tested from cli and via gRPC - seems to be ok

PR #1517, opened 2021-10-18
libdefs: a way for user to configure libraries

d-a-v:

      before creating one)
- [X] The PR follows
- [X] Not a breaking change

This proposal gives the ability for any sketch to use additional configuration files for libraries that are aware of this feature.

This is yet another proposal for #846 and https://github.com/arduino/Arduino/issues/421: Compiler flags are not addressed, global defines for sketch *and* libraries is the target.
It would additionally replace and close #1117.

A library source file (header or code) could optionally use this kind of snippet:

`CoolLibrary/CoolLibrary.h`
-----
```cpp
#if defined __has_include
#  if __has_include (<CoolLibrary_user_config.h>)
#    pragma message "CoolLibrary: Including local project config file <CoolLibrary_user_config.h>"
#    include <CoolLibrary_user_config.h>
#  endif
#endif

#ifndef COOLLIBRARY_BUFFER_SIZE
#define COOLLIBRARY_BUFFER_SIZE 42  // default value if sketch did not redefined it in its local config file
#endif 
```

`CoolLibrary/CoolLibrary.cpp`
-----
```cpp
#include "CoolLibrary.h"

// COOLLIBRARY_BUFFER_SIZE is 42 by default per header file
// but it can also be overridden by configuration file from sketch directory and taken into account in cpp file
static const unsigned char libbuf [COOLLIBRARY_BUFFER_SIZE];
...
```

A sketch using this library can create a local subdirectory called `libdefs/` and store, in this specific case, a file named `CoolLibrary_user_config.h`. Name and content of this file need to match the specific library requirements.

The goal for the library maintainer is to willingly allow users to change default values, or provide some specific configuration (like `#define COOLLIBRARY_DEBUG_MODE`) ***without*** having to force users to modify the library files (header or code). #1117 may give more details on the same subject with additional examples.

For security, files added in `libdefs/` are accessible during the compilation phases of libraries and sketch, but not when core source files are involved.

### DIFF
--- a/i18n/data/en.po
+++ b/i18n/data/en.po
@@ -395,7 +395,7 @@ msgstr "Compiling core..."
 msgid "Compiling libraries..."
 msgstr "Compiling libraries..."
 
-#: legacy/builder/phases/libraries_builder.go:135
+#: legacy/builder/phases/libraries_builder.go:136
 msgid "Compiling library \"{0}\""
 msgstr "Compiling library \"{0}\""
 
@@ -1358,7 +1358,7 @@ msgstr "Library installed"
 msgid "Library name"
 msgstr "Library name"
 
-#: legacy/builder/phases/libraries_builder.go:91
+#: legacy/builder/phases/libraries_builder.go:92
 msgid "Library {0} has been declared precompiled:"
 msgstr "Library {0} has been declared precompiled:"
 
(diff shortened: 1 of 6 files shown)


ubidefeo:
@d-a-v 
also take vision of this issue
https://github.com/arduino/tooling-rfcs/issues/9

d-a-v:
> For a beginner, having to create a folder on the Sketch level, means finding the Sketch folder and creating folder libdefs containing LIBRARY_NAME_user_config.h which in our opinion is way more cumbersome than offering a GU in which to add a number of custom fields, which will be propagated as #define directives during the pre-compile steps.

This is true and I have much simplified this in #1524 which requires no other change in the current tools. No need to ask or change anything from libraries, a regular per-project GUI-editable header file backward compatible with a number of libraries. Also compatible with Arduino 1.8.

I will close this PR and #1117 and see if #1524 finds your interest.

PR #1524, opened 2021-10-20
automatic inclusion of `sketch_globals.h` for `sketch.ino` and libraries

d-a-v:
      before creating one)
- [X] The PR follows
- [X] No breaking change

User is given the ability to add local overriding for cpp-defines used by sketch ***and libraries***.
This is done through a new and optional user file editable from Arduino-GUI and which must stand beside `someSketch.ino` under the name `someSketch_globals.h`.

When it exists, this file is automatically included first in all compilation units (`.ino`, `.cpp`) through the compiler's `-include file` command line option.

This is yet another proposal for #846 and arduino/Arduino#421, and follows #1117 and #1517.

Features:
- Libraries can be configured without modification
- No change required from configurable libraries which support macro overrides.
- Self contained projects archive embedding library configuration
- Single C++ header file per project to configure all used libraries
- Compatible with Arduino 1.8's GUI
- platform's Core source files *do not* include the user global header file (*edit*)

follows https://github.com/arduino/arduino-cli/pull/1517#issuecomment-947448275 .


legacy/builder/gcc_preproc_runner.go
```diff
@@ -61,7 +61,8 @@ func prepareGCCPreprocRecipeProperties(ctx *types.Context, sourceFilePath *paths
 	properties.SetPath(constants.BUILD_PROPERTIES_SOURCE_FILE, sourceFilePath)
 	properties.SetPath(constants.BUILD_PROPERTIES_PREPROCESSED_FILE_PATH, targetFilePath)
 
-	includesStrings := utils.Map(includes.AsStrings(), utils.WrapWithHyphenI)
+    ctx.SetGlobalIncludeOption()
```
z0nt, line 64: Please check the spacing throughout the change.

z0nt:
I'd suggest renaming to:
> automatic inclusion of sketch_globals.h for sketch.ino and libraries

legacy/builder/gcc_preproc_runner.go
d-a-v: All 4-spaces are now replaced by tabs.

PaulStoffregen:
Before merging this, please consider the long term effect on the Arduino ecosystem.  Historically, Arduino has rejected this idea.  If it's now considered okay, then great.

d-a-v:
I can't see any adverse effect for backward compatibility.
If you have one in mind please elaborate.

z0nt:
Having a single `sketch_globals.h` is probably not very safe.  This approach may cause macro name clashing from different libraries.  We cannot assume that all libraries use prefixed macro names.  I'd suggest to have one config file per library - `LibFoo_config.h`, `LibBar_config.h`, etc.  When LibFoo is being built only `LibFoo_config.h` is included (if exists).  Let me know what you think.

d-a-v:
I think a way for using global defines is really lacking in the Arduino build system for a user point of view, in the same sense where in any other build environment that I know of, there is always a place where one can add some custom #defines, which are enabled in a global manner.

Allowing to use a separate and distinct file per library is an interesting idea, which indeed would avoid name clashing.
However I can't seem to be able to identify a single build environment where one could add external custom set of defines per included library / directory (makefiles and cmakefiles don't do that, eclipse has a global configuration place, platformio too, I have no knowledge for others).
So it appears as just not natural to me.
Also I believe its use would be quite not easy for everyone, because it is not rare that an Arduino library is stored into TheLibrary/src/LibraryHeader1or2.h. Then on what base the NameForALibrary_globals.h should be chosen in a safe and predictible manner ?

Having a single file named after user's sketch appears more simple to me, and sticks to the Arduino simplicity.

About library name clashing, this is indeed an issue.
However if the proposed feature is (ever) adopted in Arduino, library authors would surely accept little and backward compatible changes in their library. Also, as some Arduino libraries are also used in other build environments they are already prepared for this use and probably already have unique sets of defines.

Issue #1447, opened 2021-09-13
Q: back-end scenario


4ntoine:
## Question

I've [asked](https://stackoverflow.com/questions/69082483/how-to-get-compiled-binary-content-not-just-path-with-arduino-cli) on SO but did not get any answers unfortunately.
Sorry for cross posting here.

Is there any opportunity to get new files content, compiled binaries content done vis cli vis gRPC calls?

### Current behavior

All the methods return arduino cli host local fs paths. This limits the use cases and require direct access
to cli host file system or alternative services to provide access (remote fs, ssh, etc).

### Expected behavior

Expected to have an opportunity to pass the content of the files, not just paths. This could be done by just returning `oneof` depending on the (new) request arguments (paths/content).
 
### Environment

- CLI version (output of `arduino-cli version`): 0.19.0
- OS and platform:

### Additional context

How does it work with "arduino create"?

silvanocerza:
The Sketch build path is generated automatically if one is not set explicitly by the user, on terminal it's done using the `--build-path` flag of the `compile` command, on gRPC using `CompileRequest.build_path`.

The build path is generated by this function:

 https://github.com/arduino/arduino-cli/blob/58ddae7ba97c5aad19e7e6c030715f7db83e8348/arduino/sketch/sketch.go#L285-L293

The `upload` command, and the relative gRPC function, first searches for the binaries in the build path that would be generated using the specified Sketch path, that's only done if the `--import-dir` or `--import-file` are not specified. On gRPC they're `UploadRequest.import_file` and `UploadRequest.import_dir` respectively.

Does this help? 

> How does it work with "arduino create"?

Well, that's quite complicated, also it doesn't use the Arduino CLI fully as of now.

4ntoine:
@silvanocerza thanks for the feedback. I was not clear enough. All this assumes that the client and the service are on the same host. But i'm talking  about the use case when the client and the service are on different hosts. So it means not paths but actual content should be available over gRPC (or additional service available that returns the files content)

silvanocerza:
Interesting. :thinking: 

I seem to understand from the SO question that you already managed to compile a Sketch on your host so I'll assume that's working correctly.

What is not clear to me is what you're trying to accomplish after that, do you want to get the compiled binaries from the host to upload them through the client? 

[...]

PR #1536, opened 2021-11-03
#1447 - back-end usage scenario

4ntoine:
Add .proto file (messages and a service), generate go files.
(Draft to demonstrate the proposal).

      before creating one)
- [ ] The PR follows

The MR Draft demonstrates the back-end scenario when client and server are on different hosts thus making direct files access problematic. So the client can't read neither the sketches content, nor the compiled binaries.

The feature is not implemented

There is "file system" service allowing to request file content.

* **Other information**:
This is an alternative to returning file content every here and there.

---


rpc/cc/arduino/cli/files/v1/files.proto
```diff
@@ -0,0 +1,45 @@
...
+}
+
+enum ContentType {
+  CONTENT_TYPE_RAW_UNSPECIFIED = 0;
+  CONTENT_TYPE_ZIP = 1;
+}
+
+message LoadFileRequest {
+  // path as it's returned in LoadSketchResponse for instance
+  string path = 1;
```
4ntoine, line 35: we could use `repeated` to be able to return few file entires with just 1 call

rpc/cc/arduino/cli/files/v1/files.proto
```diff
@@ -0,0 +1,45 @@
...
+  // Optional
+  ContentType type = 2;
+}
+
+message LoadFileResponse {
+  // File content
+  bytes content = 1;
+  // is skipped in case of error
+  ContentType type = 2;
```
4ntoine, line 44: could be removed to avoid duplication. The reason it's still here is that the server might still return not the requested type (for any reason)

rpc/cc/arduino/cli/files/v1/files.proto
```diff
@@ -0,0 +1,45 @@
...
+message LoadFileRequest {
+  // path as it's returned in LoadSketchResponse for instance
+  string path = 1;
+  // Optional
+  ContentType type = 2;
+}
+
+message LoadFileResponse {
+  // File content
+  bytes content = 1;
```
4ntoine, line 42: or it could be smth like (depends on the design if whether we want to let the client know error code to add custom behaviour):
```
oneof {
    bytes content = 1;
    grpc.status error =2;
}

**Comment by silvanocerza** [comment]
Interesting proposal but I have some concerns. 

Protocol buffers are not meant to be used with large data sets as per [documentation](https://developers.google.com/protocol-buffers/docs/techniques#large-data), sending the whole content of a file via gRPC is not a great idea I think.

But the major concern is about security, if the `arduino-cli` is running on a backend and any client can ask the content of any file that has permission to read you open up a BIG hole I think. Am not really confortable with this.

In the end I think this must not be a concern of the CLI, files exchange should be handled by another service.

**Comment by 4ntoine** [comment]
@silvanocerza Yeah, your concerns are more than reasonable. I've been also thinking about that. It seems if we limit the paths to `user` and `build` dirs only (to be able to load the sketches and compiled binaries) it should be ok. Also we might register the service to the server only if daemon is started with a new `--with-file-system` flag (switched off by default), so users who intentionally want to run it for back-end use case will be able to do it (and the rest will be not affected at all).

I also agree that external (dedicated) service should do it better, but imo it's very small additional feature to provide a full set of needed functional so it seems to be not a big deal. I doubt that average user will be able to install smth like https://github.com/hubot-grpc/filesystem-grpc and want to deal with dockers, daemons, etc.

But instead it will allow users to use their Raspi's to compile for Arduino from any mobile client (smartphones, tablet, smart devices, etc).

**Comment by ubidefeo** [comment]
@4ntoine 
>But instead it will allow users to use their laptops/desktops or even Raspi's with arduino-cli to compile for Arduino from any mobile client (smartphones, tablet, smart devices, etc).

This is a use case I was interested in, but I have opted for running CLI on a Pi, editing files over SSH and then running the appropriate command over SSH with the board attached to the Pi's USB port.
Have you thought about "mapping" the remote FS using SAMBA?
You could in theory configure the CLI to have the paths required be subfolders of that mounted path.
Not sure it would work on iPad, though
```
directories:
  data: /Users/ubi/Library/Arduino15
  downloads: /Users/ubi/Library/Arduino15/staging
  user: /data/Dropbox/AppsDocuments/Arduino
```

4ntoine:
@ubidefeo 

yes, i tried different scenarios including samba (and it worked).

I'm thinking about "average user" and Arduino is awesome due to it's simplicity. So imo if it's costs only 42 lines of code as in this MR lot's of people will say "thank you guys for doing our life easier"

4ntoine:
@silvanocerza will [it](https://github.com/arduino/arduino-cli/pull/1536#issuecomment-961291825) work for you to address security concerns? i can update the MR

facchinm:
Closing as stale

PR #1574, opened 2021-11-26
Optimize `core` operations, improving on the user input

umbynos:

      before creating one)
- [X] The PR follows

Feature

installing / uninstalling / upgrading / downloading cores is not case insesitive

now it is's case-insensitive, so it works if you try to `core install Arduino:samd`

no
* **Other information**:

---

### DIFF
--- a/cli/arguments/reference.go
+++ b/cli/arguments/reference.go
@@ -18,6 +18,10 @@ package arguments
 import (
 	"fmt"
 	"strings"
+
+	"github.com/arduino/arduino-cli/cli/instance"
+	"github.com/arduino/arduino-cli/commands/core"
+	rpc "github.com/arduino/arduino-cli/rpc/cc/arduino/cli/commands/v1"
 )
 
 // Reference represents a reference item (core or library) passed to the CLI
@@ -79,6 +83,21 @@ func ParseReference(arg string) (*Reference, error) {
 		return nil, fmt.Errorf(tr("invalid empty core architecture '%s'"), arg)
 	}
 	ret.Architecture = toks[1]
+
+	// We try to optimize the search and help the user
+	platforms, _ := core.GetPlatforms(&rpc.PlatformListRequest{
+		Instance:      instance.CreateAndInit(),
+		UpdatableOnly: false,
+		All:           true,
+	})
+	for _, platform := range platforms {
+		if strings.EqualFold(platform.GetId(), ret.PackageName+":"+ret.Architecture) {
+			toks = strings.Split(platform.GetId(), ":")
+			ret.PackageName = toks[0]
+			ret.Architecture = toks[1]
+			break // is it ok to stop at the first result?
+		}
+	}
 
 	return ret, nil
 }
--- a/cli/arguments/reference.go
+++ b/cli/arguments/reference.go
@@ -84,20 +84,25 @@ func ParseReference(arg string) (*Reference, error) {
 	}
 	ret.Architecture = toks[1]
 
-	// We try to optimize the search and help the user
+	// Now that we have the required informations in `ret` we can
+	// try to use core.GetPlatforms to optimize what the user typed
+	// (by replacing the PackageName and Architecture in ret with the content of core.GetPlatform())
 	platforms, _ := core.GetPlatforms(&rpc.PlatformListRequest{
 		Instance:      instance.CreateAndInit(),
 		UpdatableOnly: false,
-		All:           true,
+		All:           true, // this is true because we want also the installable platforms
 	})
+	var found = 0
 	for _, platform := range platforms {
 		if strings.EqualFold(platform.GetId(), ret.PackageName+":"+ret.Architecture) {
 			toks = strings.Split(platform.GetId(), ":")
-			ret.PackageName = toks[0]
-			ret.Architecture = toks[1]
-			break // is it ok to stop at the first result?
+			found++
 		}
 	}
-
+	// replace the returned Reference only if only one occurrence is found, otherwise leave it as is
+	if found == 1 {
+		ret.PackageName = toks[0]
+		ret.Architecture = toks[1]
+	}
 	return ret, nil
 }
--- a/cli/arguments/reference.go
+++ b/cli/arguments/reference.go
@@ -37,10 +37,10 @@ func (r *Reference) String() string {
 
 // ParseReferences is a convenient wrapper that operates on a slice of strings and
 // calls ParseReference for each of them. It returns at the first invalid argument.
-func ParseReferences(args []string, parseArch bool) ([]*Reference, error) {
+func ParseReferences(args []string) ([]*Reference, error) {
 	ret := []*Reference{}
 	for _, arg := range args {
-		reference, err := ParseReference(arg, parseArch)
+		reference, err := ParseReference(arg)
 		if err != nil {
 			return nil, err
 		}
@@ -49,10 +49,8 @@ func ParseReferences(args []string, parseArch bool) ([]*Reference, error) {
 	return ret, nil
 }
 
-// ParseReference parses a string and returns a Reference object. If `parseArch` is passed,
-// the method also tries to parse the architecture bit, i.e. string must be in the form
-// "packager:arch@version", useful to represent a platform (or core) name.
-func ParseReference(arg string, parseArch bool) (*Reference, error) {
+// ParseReference parses a string and returns a Reference object.
+func ParseReference(arg string) (*Reference, error) {
 	ret := &Reference{}
 	if arg == "" {
 		return nil, fmt.Errorf(tr("invalid empty core argument"))
@@ -69,20 +67,18 @@ func ParseReference(arg string, parseArch bool) (*Reference, error) {
 		ret.Version = toks[1]
 	}
 
-	if parseArch {
-		toks = strings.Split(ret.PackageName, ":")
-		if len(toks) != 2 {
-			return nil, fmt.Errorf(tr("invalid item %s"), arg)
-		}
-		if toks[0] == "" {
-			return nil, fmt.Errorf(tr("invalid empty core name '%s'"), arg)
-		}
-		ret.PackageName = toks[0]
-		if toks[1] == "" {
-			return nil, fmt.Errorf(tr("invalid empty core architecture '%s'"), arg)
-		}
-		ret.Architecture = toks[1]
+	toks = strings.Split(ret.PackageName, ":")
+	if len(toks) != 2 {
+		return nil, fmt.Errorf(tr("invalid item %s"), arg)
 	}
+	if toks[0] == "" {
+		return nil, fmt.Errorf(tr("invalid empty core name '%s'"), arg)
+	}
+	ret.PackageName = toks[0]
+	if toks[1] == "" {
+		return nil, fmt.Errorf(tr("invalid empty core architecture '%s'"), arg)
+	}
+	ret.Architecture = toks[1]
 
 	return ret, nil
 }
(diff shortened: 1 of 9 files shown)


cli/arguments/reference.go
```diff
@@ -69,20 +73,39 @@ func ParseReference(arg string, parseArch bool) (*Reference, error) {
...
-			return nil, fmt.Errorf(tr("invalid empty core architecture '%s'"), arg)
+	}
+	if toks[1] == "" {
+		return nil, fmt.Errorf(tr("invalid empty core architecture '%s'"), arg)
+	}
+	ret.Architecture = toks[1]
+	// we do the same for architecture
+	if inst != nil { // do not perform the search if the instance is not defined
+		for _, platform := range targetPackage.Platforms {
+			if platform.Architecture == strings.ToLower(ret.Architecture) {
```
per1234, line 104:
```suggestion
			if strings.ToLower(platform.Architecture) == strings.ToLower(ret.Architecture) {
```
Architecture names are not always lower case.

For example:
https://github.com/redbear/redbear.github.io/blob/master/arduino/package_redbear_nRF5x_index.json#L14
```
                    "architecture": "nRF51822",
```

per1234 (changes requested):

cli/arguments/reference.go
```diff
@@ -37,10 +41,10 @@ func (r *Reference) String() string {
 
 // ParseReferences is a convenient wrapper that operates on a slice of strings and
 // calls ParseReference for each of them. It returns at the first invalid argument.
-func ParseReferences(args []string, parseArch bool) ([]*Reference, error) {
+func ParseReferences(args []string, inst *rpc.Instance) ([]*Reference, error) {
```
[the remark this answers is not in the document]
cmaglie: We should really start thinking about putting some of the modules under the `internal` package, the `cli` module is one that doesn't have much use outside the arduino-cli.

cli/arguments/reference.go
```diff
@@ -69,20 +73,39 @@ func ParseReference(arg string, parseArch bool) (*Reference, error) {
...
+		return nil, fmt.Errorf(tr("invalid empty core name '%s'"), arg)
+	}
+	ret.PackageName = toks[0]
+
+	// try to search in the packages, this way ParseReference can be case insensitive
+	// note that if the search fail for some reason the ret.Packagename does not get overwritten
+	var targetPackage *cores.Package
+	if inst != nil { // do not perform the search if the instance is not defined
+		pm := commands.GetPackageManager(inst.Id)
+		for _, targetPackage = range pm.Packages {
```
cmaglie, line 90: We should not access the `instance` field directly (like `pm.Packages`) but we must use the RPC-mapped functions to obtain the data. In this case, I think that `core.PlatformSearch` in the `commands/core` 

```
core.PlatformSearch(&rpc.PlatformSearchRequest{
		Instance:    inst,
		SearchArgs:  arguments,
		AllVersions: allVersions,
	})
```

should return the info we need.
I'm wondering if we can make private the `Packages` field in PackageManager to avoid accidentally bypassing the gRPC API from the command-line client.

cmaglie (changes requested):

cli/arguments/reference.go
umbynos: Actually I did use the `core.PlatformSearch` function at first, then @silvanocerza suggested not to use it because the function does other stuff and can be optimized a little.

cmaglie: What about `GetPlatformList` then:

```
// GetPlatforms returns a list of installed platforms, optionally filtered by
// those requiring an update.
func GetPlatforms(req *rpc.PlatformListRequest) ([]*rpc.Platform, error) {
```

silvanocerza: Using `GetPlatforms` is not efficient, it will always iterate all packages and all their platforms much like `PlatformSearch`.

In this case we need only find one platform, if we don't want to store packages and platforms using lower case keys this is the most efficient way.

umbynos: So what should we do about the implementation?

cmaglie: We carefully avoided accessing directly the `cores` package from the `cli` package (but instead we used the `command` package to do everything). This is required to keep a feature parity between the CLI and the gRPC interface (since both use the `commands` package you can do exactly the same things using the CLI and the gRPC).
If we access directly the `Packages` field we are introducing a "feature" in the CLI that is not present in the gRPC (eventually leading to code duplication/feature incoherence).

> Using GetPlatforms is not efficient, it will always iterate all packages and all their platforms much like PlatformSearch.

I don't think any user will ever be able to see this optimization since the bulk of the operations is the loading of the platforms at init time (that sometimes takes up to a second), the few milliseconds spent iterating a bunch of platforms is negligible.

Anyway, in the platform specification, the platform name **is** case sensitive, so we cannot stop at the first match, because there could be more than one platform matching: in that case the CLI should output an error and ask the user to disambiguate.

silvanocerza: > We carefully avoided accessing directly the `cores` package from the `cli` package (but instead we used the `command` package to do everything). This is required to keep a feature parity between the CLI and the gRPC interface (since both use the `commands` package you can do exactly the same things using the CLI and the gRPC).
If we access directly the `Packages` field we are introducing a "feature" in the CLI that is not present in the gRPC (eventually leading to code duplication/feature incoherence).

You're right, didn't think about this. :thinking: 

> Anyway, in the platform specification, the platform name **is** case sensitive, so we cannot stop at the first match, because there could be more than one platform matching: in that case the CLI should output an error and ask the user to disambiguate.

This concerns me though, if platform names are case sensitive and we make commands work without sensitivity I think it may cause issues further in the future. Is it actually worth making this change? :confused:

cli/arguments/reference.go
```diff
@@ -52,7 +56,7 @@ func ParseReferences(args []string, parseArch bool) ([]*Reference, error) {
 // ParseReference parses a string and returns a Reference object. If `parseArch` is passed,
 // the method also tries to parse the architecture bit, i.e. string must be in the form
 // "packager:arch@version", useful to represent a platform (or core) name.
-func ParseReference(arg string, parseArch bool) (*Reference, error) {
+func ParseReference(arg string, inst *rpc.Instance) (*Reference, error) {
```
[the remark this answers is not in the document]
umbynos: Sorry @per1234 I don't get why this is a braking change.. It's part of the `cli` package and not the grpc one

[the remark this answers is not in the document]
per1234: The API of the packages of the `github.com/arduino/arduino-cli` Go module is also a public API. It is one of the three pillars of Arduino CLI:

https://arduino.github.io/arduino-cli/dev/integration-options/#the-third-pillar-embedding

A breaking change to any of the three pillars is considered a breaking change, which must be clearly communicated to the users via the standard process:

https://arduino.github.io/arduino-cli/dev/CONTRIBUTING/#pull-request-checklist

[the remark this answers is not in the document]
umbynos: Oh crap! You are right (as always). Thanks Per :medal_military:

cli/arguments/reference.go
```diff
@@ -69,20 +73,39 @@ func ParseReference(arg string, parseArch bool) (*Reference, error) {
...
+	}
+	ret.PackageName = toks[0]
+
+	// try to search in the packages, this way ParseReference can be case insensitive
+	// note that if the search fail for some reason the ret.Packagename does not get overwritten
+	var targetPackage *cores.Package
+	if inst != nil { // do not perform the search if the instance is not defined
+		pm := commands.GetPackageManager(inst.Id)
+		for _, targetPackage = range pm.Packages {
+			if targetPackage.Name == strings.ToLower(ret.PackageName) {
```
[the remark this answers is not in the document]
umbynos: using [`strings.EqualFold`](https://pkg.go.dev/strings#EqualFold) should be sufficient. Thanks Per

cli/arguments/reference.go
umbynos: using [`strings.EqualFold`](https://pkg.go.dev/strings#EqualFold) should be sufficient. Thanks Per

cli/arguments/reference.go
umbynos: Ok I'm trying to use the `core.GetPlatform()` approach and see what happens

cli/arguments/reference.go
```diff
@@ -69,20 +73,41 @@ func ParseReference(arg string, parseArch bool) (*Reference, error) {
...
+	}
+	// replace the returned Reference only if only one occurrence is found, otherwise leave it as is
+	if found == 1 {
 		ret.PackageName = toks[0]
-		if toks[1] == "" {
-			return nil, fmt.Errorf(tr("invalid empty core architecture '%s'"), arg)
-		}
 		ret.Architecture = toks[1]
+	} else {
+		logrus.Warnf("Found %d platform for reference: %s", found, arg)
```
silvanocerza, line 110: I don't think we should just warn in this case, if we're not which platform the user actually meant to use we should return an error in my opinion.

cli/arguments/reference.go
```diff
@@ -49,10 +54,9 @@ func ParseReferences(args []string, parseArch bool) ([]*Reference, error) {
 	return ret, nil
 }
 
-// ParseReference parses a string and returns a Reference object. If `parseArch` is passed,
-// the method also tries to parse the architecture bit, i.e. string must be in the form
-// "packager:arch@version", useful to represent a platform (or core) name.
-func ParseReference(arg string, parseArch bool) (*Reference, error) {
+// ParseReference parses a string and returns a Reference object.
```
silvanocerza, line 57: We should document the fact that `ParseReference` tries to infer which platform reference should be returned.

docs/UPGRADING.md
```diff
@@ -2,6 +2,19 @@
 
 Here you can find a list of migration guides to handle breaking changes between releases of the CLI.
 
+## Unreleased
+
+### `ParseReferences()` function change in`arduino-cli` package
```
per1234, line 7:
```suggestion
### `github.com/arduino/arduino-cli/cli/arguments.ParseReferences` function change
```

This convention is in use previously in the document for referring to packages:

https://arduino.github.io/arduino-cli/dev/UPGRADING/#githubcomarduinoarduino-clii18n-package

My proposal extends that approach to provide fully qualified symbol references, as is use with [`go doc`](https://pkg.go.dev/cmd/doc)

docs/UPGRADING.md
```diff
@@ -2,6 +2,19 @@
...
+
+The `parseArch` parameter was removed since it was unused and was always true. This means that the architecture gets
+always parsed by the function.
+
+### `ParseReference()` function change in`arduino-cli` package
```
per1234, line 12:
```suggestion
### `github.com/arduino/arduino-cli/cli/arguments.ParseReference` function change
```
See explanation in previous comment.

per1234 (changes requested):

per1234 (approved):

cli/arguments/reference.go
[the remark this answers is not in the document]
Maclain27: - Hi guys I am progrommer and I already had the program for self_balance_robot I need this cli to finish it allow me in.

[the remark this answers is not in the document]
Maclain27: > - Hi guys I am progrommer and I already had the program for self_balance_robot I need this cli to finish it allow me in!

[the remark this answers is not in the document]
Maclain27: - Add access to the repositories to get fuction.

silvanocerza (approved):

PR #1686, opened 2022-03-07
Add secure boot support for compile command.

umbynos:

      before creating one)
- [X] The PR follows

feature

The secure boot is now supported in the compile command

I've added three flags to support the override of the default security keys used to sign and encrypt a binary:
- `--keys-keychain` allows specifying the directory used to search for the security keys
- `--sign-key` indicates the name+extension of the sign key to use
- `--encrypt-key` indicates the name+extension of the encryption key to use

The keys are red from the properties.Map representing the `platform.txt` file and are overwritten, even if `security=sien` board option is not used. Because we don't know if it will be the standard for secure boot. 
 
* **Other information**:

---


matthijskooijman:
I have not been involved with any development leading up to this, but at first glance it seems these extra options added are a bit arbitrary. Various platforms support all kinds of options in their `platform.txt`, and I wonder if it makes sense to make these particular options special and understood by arduino-cli. Going onto this train of development might eventually lead to a ton of different options, each supported by different platforms, each with their own compatibility needs...

IOW, I wonder:
 - Are these options really needed? Can the existing `--build-property` option not be used instead?
 - If these are needed, are they generic enough? I.e. will they work for a broad range of platforms and usecases, or will more (similar but not quite the same) options be needed in the future for other platforms?
 - Should this PR not include an update for the platform.txt documentation, to clearly document the particular keys updated by this option, with a clear contract about how they are intended to be used and set?

per1234 (changes requested):

commands/compile/compile.go
```diff
@@ -118,13 +118,34 @@ func Compile(ctx context.Context, req *rpc.CompileRequest, outStream, errStream
...
+	InstalledPlatformRelease := pm.GetInstalledPlatformRelease(targetPlatform)
+	if targetPlatform == nil || InstalledPlatformRelease == nil {
 		return nil, &arduino.PlatformNotFoundError{
 			Platform: fmt.Sprintf("%s:%s", fqbn.Package, fqbn.PlatformArch),
 			Cause:    fmt.Errorf(tr("platform not installed")),
 		}
 	}
 
+	// At the current time we do not have a way of knowing if a board supports the secure boot or not,
+	// so, if the flags to override the default keys are used, we try override the corresponding property in the properties.txt nonetheless.
```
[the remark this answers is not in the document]
umbynos: Exactly, should I rephrase better the sentence?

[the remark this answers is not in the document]
per1234: I provided a suggestion, but you are welcome to phrase it as you like. The part I found confusing was "properties.txt".

umbynos:
`--board-options` PR created #1688

cli/compile/compile.go
```diff
@@ -100,6 +105,14 @@ func NewCommand() *cobra.Command {
 		tr("List of custom build properties separated by commas. Or can be used multiple times for multiple properties."))
 	compileCommand.Flags().StringArrayVar(&buildProperties, "build-property", []string{},
 		tr("Override a build property with a custom value. Can be used multiple times for multiple properties."))
+	compileCommand.Flags().StringSliceVar(&boardOptions, "board-options", []string{},
```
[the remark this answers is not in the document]
umbynos: Done in #1688

umbynos:
Docs updated

docs/platform-specification.md
```diff
@@ -768,6 +768,60 @@ All the tools launched to compile or upload a sketch will have the following env
...
+```
+
+The currently we support the secure boot only with `build.keys.type=public_keys` but in the future other ways can be
+added. The security keys can be added with:
+
+- `keys.keychain` indicates the path of the dir where to search for the custom keys to sign and encrypt a binary.
+- `keys.sign_key` indicates the name of the custom signing key to use to sign a binary during the compile process.
+- `keys.encrypt_key` indicates the name of the custom encryption key to use to encrypt a binary during the compile
+  process.
+
```
per1234, line 824:
```suggestion
### Secure Boot

Some boards supports the secure boot. Basically the compiled sketch can be signed and encrypted with a [tool](#tools)
before being flashed to the target board. The bootloader of the board is then responsible for starting the compiled
sketch if the matching keys are used.

To be able to correctly carry out all the operations at the end of the build we can leverage the
[post build hooks](#pre-and-post-build-hooks-since-arduino-ide-165) to sign and encrypt a binary by using
`recipe.hooks.objcopy.postobjcopy.NUMBER.pattern` key in [`platform.txt`](#platformtxt). The security keys used are
defined in the boards file, this way there could be different keys for different boards.

```
[...]
## Create output secure image (bin file)
recipe.hooks.objcopy.postobjcopy.1.pattern={build.postbuild.cmd}
#
# IMGTOOL
#

tools.imgtool.cmd=imgtool
tools.imgtool.build.pattern=sign --key "{build.keys.keychain}/{build.keys.sign_key}" --encrypt "{build.keys.keychain}/{build.keys.encrypt_key}" "{build.path}/{build.project_name}.bin" "{build.path}/{build.project_name}.bin" --align {build.alignment} --max-align {build.alignment} --version {build.version} --header-size {build.header_size} --pad-header --slot-size {build.slot_size}
[...]

```

By having only `tools.TOOL_NAME.cmd` and `tools.TOOL_NAME.build.pattern`, we can customize the behavior with a
[custom board option](#custom-board-options). Then in the [`boards.txt`](#boardstxt) we can define the new option to use
a different `postbuild.cmd`:

```
[...]
menu.security=Security setting

envie_m7.menu.security.none=None
envie_m7.menu.security.sien=Signature + Encryption

envie_m7.menu.security.sien.build.postbuild.cmd="{tools.imgtool.cmd}" {tools.imgtool.build.pattern}
envie_m7.menu.security.none.build.postbuild.cmd="{tools.imgtool.cmd}" exit

envie_m7.menu.security.sien.build.keys.type=public_keys
envie_m7.menu.security.sien.build.keys.keychain={runtime.hardware.path}/Default_Keys
envie_m7.menu.security.sien.build.keys.sign_key=default-signing-key.pem
envie_m7.menu.security.sien.build.keys.encrypt_key=default-encrypt-key.pem
[...]
```

The currently we support the secure boot only with `build.keys.type=public_keys` but in the future other ways can be
added. The security keys can be added with:

- `keys.keychain` indicates the path of the dir where to search for the custom keys to sign and encrypt a binary.
- `keys.sign_key` indicates the name of the custom signing key to use to sign a binary during the compile process.
- `keys.encrypt_key` indicates the name of the custom encryption key to use to encrypt a binary during the compile
  process.

```

This is really valuable content, but I don't think it is appropriate for the specification. This would be better published as a guide for one possible use case of the flags.

The specification should only document the property names and the Arduino CLI flags that set them. I think the most appropriate place for them is under the "Build Process" section:

https://github.com/arduino/arduino-cli/blob/e325b1674181e47799ebc02bb76c2cbc8d15fce3/docs/platform-specification.md#build-process

Make sure to use the correct property names:

- `build.keys.keychain`
- `build.keys.sign_key`
- `build.keys.encrypt_key`

per1234 (changes requested):
My opinion is that we should postpone the merge of this.

Start out by using the existing mechanism, as suggested by @matthijskooijman:
```
arduino-cli compile --build-property build.keys.sign_key=<some path>  --build-property build.keys.encrypt_key=<some path>
```
I really don't think it is so much more difficult for the advanced users who will use Arduino CLI and secure boot.

Once the first platform providing "secure boot" capability has been released and in use by the users, any requirements for facilitating its use may become more clear.

cli/compile/compile.go
```diff
@@ -100,6 +104,12 @@ func NewCommand() *cobra.Command {
 		tr("List of custom build properties separated by commas. Or can be used multiple times for multiple properties."))
 	compileCommand.Flags().StringArrayVar(&buildProperties, "build-property", []string{},
 		tr("Override a build property with a custom value. Can be used multiple times for multiple properties."))
+	compileCommand.Flags().StringVar(&keysKeychain, "keys-keychain", "",
```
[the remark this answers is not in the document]
umbynos: At first, I asked the same question, but having these three flags instead of 2 results in a more future-proof implementation:
In the future, `keys-keychain` could point to a real keychain instead of a simple directory and the other two flags could indicate the actual name of the keys inside the keychain.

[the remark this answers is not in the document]
per1234: Make sense. Thanks for the clarification.

legacy/builder/setup_build_properties.go
```diff
@@ -126,6 +127,17 @@ func (s *SetupBuildProperties) Run(ctx *types.Context) error {
...
+	if buildProperties.ContainsKey("build.keys.type") {
+		if buildProperties.Get("build.keys.type") == "public_keys" {
+			if !buildProperties.ContainsKey("build.keys.keychain") || !buildProperties.ContainsKey("build.keys.sign_key") || !buildProperties.ContainsKey("build.keys.encrypt_key") {
+				return errors.Errorf("%s core does not specify correctly default sign and encryption keys", ctx.BuildCore)
+			}
+		} else {
+			return errors.New("\"build.keys.type\" key only supports \"public_keys\" value for now")
+		}
+	}
+
```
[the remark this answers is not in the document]
umbynos: Yeah, this property is not used at the current time. But as I explained above, in the future we could add support for other approaches (like [a real keychain](https://github.com/arduino/arduino-cli/pull/1686/files#r826749584)). If we add something like this later, it could result in a breaking change. It was an idea by @cmaglie

ubidefeo:
In my opinion, as we offer facilities to burn a bootloader, we should enhance `arduino-cli` with the functionalities we need for secure-boot platforms.
Starting with our choice of MCUboot, which we might use in more upcoming boards, our CLI needs to be able to handle operations such as key generation, binary encryption and upload of such binaries.

Although the use of `--build-property` is a valid option, semantically I'd like to keep such operations into their respective realms.
Nothing prevents a user to work their way through a convoluted breakdown of properties and use them in the canonical way, but I'd prefer to keep the operations easily documented also through the command line itself (`-h`).

Once we provide support for our preferred way of doing key generation to support MCUboot, developers of other platforms and boards flavours might decide to implement other secure bootloaders and use the same commands which will leverage their own scripts and tools.

The idea of providing a local keychain is in my opinion valuable as more advanced use cases present themselves, hence we might end up implementing some form of keychain management through the CLI.
As we discussed yesterday in an internal meeting with @facchinm and @umbynos , generation of `.pem` files using `ecdsa-p256` would not be done through `imgtool` but directly in Go to simplify things and reduce clutter, but we still want to offer platform developers a gateway into writing their own implementation of secure bootloaders or other means of generating keys to be used in yet unspecified contexts

legacy/builder/setup_build_properties.go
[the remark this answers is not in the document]
per1234: The default value allows it to be added later without breakage.

docs/platform-specification.md
umbynos: Got it, what do you think is the best place to put this content? `/docs/secure-boot.md`?

per1234: I'm not sure. I have pondered this in relation to some other guide/tutorial type content we have considered, such for creating Arduino [boards platforms](https://github.com/arduino/arduino-cli/issues/847) (https://github.com/arduino/arduino-cli/issues/847) or [precompiled libraries](https://forum.arduino.cc/t/precompiling-a-big-library/626128/7).

I'm kind of on the fence about whether that type of content belongs in the Arduino CLI repository, but it is developer targeted and might benefit from being maintained alongside the code that implements those capabilities.

So I think the best would be to create a sub-section of the documentation as a container for guides, like `docs/guides/secure-boot.md`

legacy/builder/setup_build_properties.go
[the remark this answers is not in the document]
umbynos: https://github.com/arduino/arduino-cli/pull/1686/commits/80489c467a8ce7015cf2d8980d12ebd200859eff

docs/platform-specification.md
umbynos: Thanks for the useful suggestion. I'll do it right away

umbynos: Done in https://github.com/arduino/arduino-cli/commit/974f8b597f2198809f779ce67e6a0f64da88eee0

docs/platform-specification.md
```diff
@@ -155,6 +155,16 @@ the name of the architecture is set as well.
 
 There are some other **{build.xxx}** properties available, that are explained in the boards.txt section of this guide.
 
+Some of them allows to specify trusted security credentials (sign and encryption keys) that can be used for the secure
+boot:
```
per1234, line 159:
```suggestion
Some of them allow specifying trusted security credentials (signing and encryption keys) that can be used by a
["secure boot" system](guides/secure-boot.md):
```
- Minor grammar
- Correct terminology ("sign key" -> "signing key")
- Add reference link to secure boot guide.

legacy/builder/setup_build_properties.go
```diff
@@ -126,6 +127,17 @@ func (s *SetupBuildProperties) Run(ctx *types.Context) error {
 
 	buildProperties.Merge(ctx.PackageManager.CustomGlobalProperties)
 
+	if !buildProperties.ContainsKey("build.keys.type") {
+		buildProperties.Set("build.keys.type", "public_keys") // The default is "pubblic_keys" for now
+	}
```
per1234, line 132:
```suggestion
```
This is not needed now. There is no use of the property, so it doesn't need to be set.

If it is later found that Arduino CLI must have special handling for multiple key types, some code of this sort should be added at that time in order to provide backwards compatibility for the existing platforms.

legacy/builder/setup_build_properties.go
```diff
@@ -126,6 +127,17 @@ func (s *SetupBuildProperties) Run(ctx *types.Context) error {
...
+	keychainProp := buildProperties.ContainsKey("build.keys.keychain")
+	signProp := buildProperties.ContainsKey("build.keys.sign_key")
+	encryptProp := buildProperties.ContainsKey("build.keys.encrypt_key")
+	// we verify that all the properties for the secure boot keys are defined or nono of them is defined.
```
per1234, line 136:
```suggestion
	// we verify that all the properties for the secure boot keys are defined or none of them is defined.
```

docs/guides/secure-boot.md
```diff
@@ -0,0 +1,54 @@
...
+envie_m7.menu.security.none.build.postbuild.cmd="{tools.imgtool.cmd}" exit
+
+envie_m7.menu.security.sien.build.keys.type=public_keys
+envie_m7.menu.security.sien.build.keys.keychain={runtime.hardware.path}/Default_Keys
+envie_m7.menu.security.sien.build.keys.sign_key=default-signing-key.pem
+envie_m7.menu.security.sien.build.keys.encrypt_key=default-encrypt-key.pem
+[...]
+```
+
+Currently we support the secure boot only with `build.keys.type=public_keys` but in the future other ways can be added.
```
per1234, line 48:
````suggestion
envie_m7.menu.security.sien.build.keys.keychain={runtime.hardware.path}/Default_Keys
envie_m7.menu.security.sien.build.keys.sign_key=default-signing-key.pem
envie_m7.menu.security.sien.build.keys.encrypt_key=default-encrypt-key.pem
[...]
```

````
The `build.keys.type` property is not used for anything. It can be added to the guide along with the code if the need is later found for Arduino CLI to have special handling for different key types.

File: test/testdata/platform_with_secure_boot/boards.local.txt [hunk 10]
```diff
@@ -0,0 +1,12 @@
+menu.security=Security setting
+
+uno.menu.security.none=None
+uno.menu.security.sien=Signature + Encryption
+
+uno.menu.security.sien.build.postbuild.cmd="{tools.imgtool.cmd}" {tools.imgtool.build.pattern}
+uno.menu.security.none.build.postbuild.cmd="{tools.imgtool.cmd}" exit
+
+uno.menu.security.sien.build.keys.type=public_keys
```
per1234, line 9:
```suggestion
```

per1234 (changes requested):

docs/platform-specification.md
per1234: I think it would be helpful to add a sentence or two near the start to provide a quick explanation of the reason for an Arduino boards platform developer to add a "secure boot" capability. That could include a link to an external reference that provides more detailed information on the general subject.

umbynos: Done in ea4ba1a4f91c998d99a51e48deccac1b431803d3

per1234: Looks good!

docs/guides/secure-boot.md
```diff
@@ -0,0 +1,54 @@
...
+```
+[...]
+## Create output secure image (bin file)
+recipe.hooks.objcopy.postobjcopy.1.pattern={build.postbuild.cmd}
+#
+# IMGTOOL
+#
+
+tools.imgtool.cmd=imgtool
+tools.imgtool.build.pattern=sign --key "{build.keys.keychain}/{build.keys.sign_key}" --encrypt "{build.keys.keychain}/{build.keys.encrypt_key}" "{build.path}/{build.project_name}.bin" "{build.path}/{build.project_name}.bin" --align {build.alignment} --max-align {build.alignment} --version {build.version} --header-size {build.header_size} --pad-header --slot-size {build.slot_size}
```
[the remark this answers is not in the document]
umbynos: Done in 8abfb326

docs/guides/secure-boot.md
```diff
@@ -0,0 +1,54 @@
...
+
+envie_m7.menu.security.sien.build.keys.type=public_keys
+envie_m7.menu.security.sien.build.keys.keychain={runtime.hardware.path}/Default_Keys
+envie_m7.menu.security.sien.build.keys.sign_key=default-signing-key.pem
+envie_m7.menu.security.sien.build.keys.encrypt_key=default-encrypt-key.pem
+[...]
+```
+
+Currently we support the secure boot only with `build.keys.type=public_keys` but in the future other ways can be added.
+The security keys can be added with:
```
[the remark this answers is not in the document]
umbynos: Done in 4912d484c56dc401a5cc581deed73ca3758a51fd

per1234 (changes requested):

per1234 (approved):

cmaglie (approved):

legacy/builder/setup_build_properties.go
```diff
@@ -126,6 +127,14 @@ func (s *SetupBuildProperties) Run(ctx *types.Context) error {
 
 	buildProperties.Merge(ctx.PackageManager.CustomGlobalProperties)
 
+	keychainProp := buildProperties.ContainsKey("build.keys.keychain")
+	signProp := buildProperties.ContainsKey("build.keys.sign_key")
+	encryptProp := buildProperties.ContainsKey("build.keys.encrypt_key")
+	// we verify that all the properties for the secure boot keys are defined or none of them is defined.
+	if !(keychainProp || signProp || encryptProp) && (keychainProp && signProp && encryptProp) {
```
per1234, line 134:
```suggestion
	if (keychainProp || signProp || encryptProp) && !(keychainProp && signProp && encryptProp) {
```
Logic error in check made it always evaluate as false.

legacy/builder/setup_build_properties.go
```diff
@@ -126,6 +127,14 @@ func (s *SetupBuildProperties) Run(ctx *types.Context) error {
...
+		return errors.Errorf("%s core does not specify correctly default sign and encryption keys", ctx.BuildCore)
```
per1234, line 135:
```suggestion
		return errors.Errorf("%s platform does not specify correctly default sign and encryption keys", targetPlatform.Platform)
```
The previous error message was incorrect because the build properties are not defined by the [core](https://arduino.github.io/arduino-cli/dev/platform-specification/#cores).

The previous error message was also prone to ambiguity. It produced this error message in my test:

```text
Error during build: arduino core does not specify correctly default sign and encryption keys
```

You can see the core of the platform I tested with is named `arduino` (as is the case with most of them):
https://github.com/arduino/ArduinoCore-mbed/tree/master/cores/arduino

Since the check is done on all installed platforms, rather than only the board being compiled for, this would leave the user wondering which of their platforms was being referred to. The platform identifier leaves no doubt:

```text
Error during build: arduino:mbed_portenta platform does not specify correctly default sign and encryption keys
```

per1234 (changes requested):

docs/platform-specification.md
```diff
@@ -155,6 +155,16 @@ the name of the architecture is set as well.
 
 There are some other **{build.xxx}** properties available, that are explained in the boards.txt section of this guide.
 
+Some of them allow specifying trusted security credentials (signing and encryption keys) that can be used by a
+["secure boot" system](guides/secure-boot.md):
+
+- `build.keys.keychain`: for the directory containing the keys
+- `build.keys.sign_key`: for the signing key
+- `build.keys.encrypt_key`: for the encryption key
```
per1234, line 163:
```suggestion
#### Security credential properties

Some of them allow specifying trusted security credentials (signing and encryption keys) that can be used by a
["secure boot" system](guides/secure-boot.md):

- `build.keys.keychain`: for the directory containing the keys
- `build.keys.sign_key`: for the signing key
- `build.keys.encrypt_key`: for the encryption key

If any of these properties are defined, the others are required.
```

If there is to be a requirement for all three of the properties to be present, then it must be documented in the specification.

I created a section for the information because I was concerned that it would not be clear that the requirement only applied to these specific properties.

per1234 (changes requested):

legacy/builder/setup_build_properties.go
umbynos: You are right, I redid the table of truth and I made a mistake indeed. Thanks a lot for spotting it

per1234 (approved):

umbynos:
image

PR #1735, opened 2022-05-18
Improve the performance of includes scanning

shaoyie:

      before creating one)
- [x] The PR follows

Enhancement

It's very slow to scan the includes for a large project when cache is invalid. And the cache is very easy to be invalidated because it caches the includes by a strict chain. Also because of the chain, the scanning can only be performed by single thread.

By applying the sync.Map instead of the array, we can introduce in the goroutine to improve the performance far more better.

No breaking

* **Other information**:

---

### DIFF
--- a/legacy/builder/container_find_includes.go
+++ b/legacy/builder/container_find_includes.go
@@ -96,7 +96,11 @@ import (
 	"encoding/json"
 	"fmt"
 	"os/exec"
+	"runtime"
 	"time"
+	"sync"
+	"strings"
+	"sync/atomic"
 
 	"github.com/arduino/arduino-cli/arduino/globals"
 	"github.com/arduino/arduino-cli/arduino/libraries"
@@ -106,6 +110,10 @@ import (
 	"github.com/arduino/go-paths-helper"
 	"github.com/pkg/errors"
 )
+
+var cache_valid bool
+var libGuard sync.Mutex
+var fileGuard sync.Mutex
 
 type ContainerFindIncludes struct{}
 
@@ -125,9 +133,11 @@ func (s *ContainerFindIncludes) findIncludes(ctx *types.Context) error {
 	cachePath := ctx.BuildPath.Join("includes.cache")
 	cache := readCache(cachePath)
 
-	appendIncludeFolder(ctx, cache, nil, "", ctx.BuildProperties.GetPath("build.core.path"))
+	// The entry with no_resolve prefix will be ignored in the preload phase 
+	// in case they are already added into context at here 
+	appendIncludeFolder(ctx, cache, nil, "no_resolve_1", ctx.BuildProperties.GetPath("build.core.path"))
 	if ctx.BuildProperties.Get("build.variant.path") != "" {
-		appendIncludeFolder(ctx, cache, nil, "", ctx.BuildProperties.GetPath("build.variant.path"))
+		appendIncludeFolder(ctx, cache, nil, "no_resolve_2", ctx.BuildProperties.GetPath("build.variant.path"))
 	}
 
 	sketch := ctx.Sketch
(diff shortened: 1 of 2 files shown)


facchinm:
I'm testing this PR and the speedup is amazing :star_struck: 
Still didn't find any corner case where it doesn't correctly resolve the includes

matthijskooijman:
@shaoyie, can you expand a little bit more on how this changes the include scanning process? The old caching code was intended to ensure that the results are always the same, with or without caching. I haven't dug into your new code deeply yet, but I have the impression that this might end up with different results depending on goroutine timing in some cases (I'm thinking of a case where one library exposes `a.h` and `b.h` and another exposes just `b.h`, inclusion of the latter can depend on whether `a.h` or `b.h` is resolved first).

In any case, speeding up this process is much welcomed, since it indeed is quite slow currently.

For a related (but distinctly different) improvement to this process, see https://github.com/arduino/arduino-builder/pull/217, which improves caching when compile errors are present by letting the include caching generate .d files (in addition to the compilation process). I never finished that PR properly, since it probably requires changes to recipes/platform.txt, but I never took the time to figure out what is needed exactly and how to do that in a backward compatible way. As I said, it's distinct from this improvement, but if you're interested in also working on that, feel free to work from my code.

shaoyie:
> @shaoyie, can you expand a little bit more on how this changes the include scanning process? The old caching code was intended to ensure that the results are always the same, with or without caching. I haven't dug into your new code deeply yet, but I have the impression that this might end up with different results depending on goroutine timing in some cases (I'm thinking of a case where one library exposes `a.h` and `b.h` and another exposes just `b.h`, inclusion of the latter can depend on whether `a.h` or `b.h` is resolved first).
> 
> In any case, speeding up this process is much welcomed, since it indeed is quite slow currently.
> 
> For a related (but distinctly different) improvement to this process, see [arduino/arduino-builder#217](https://github.com/arduino/arduino-builder/pull/217), which improves caching when compile errors are present by letting the include caching generate .d files (in addition to the compilation process). I never finished that PR properly, since it probably requires changes to recipes/platform.txt, but I never took the time to figure out what is needed exactly and how to do that in a backward compatible way. As I said, it's distinct from this improvement, but if you're interested in also working on that, feel free to work from my code.

Your impression is right that in multiple goroutines situation, the scanning result may look different depends on the goroutine execution order.
My judgement is if the scanning result won't impact the compile result, then it is acceptable. So the new strategy doesn't care the scanning sequence which limited it to single thread, instead it checks the cache's validation against the timestamp of the target file and source file, in which way we can utilize multiple goroutines.
If there's extreme case as you mentioned, I just left it to the ResolveLibrary() method to decide which library will be chosen. Also in such a case, if two files had conflict in the filenames, even in the original behavior, still cannot ensure the right file is happened chosen, still need additional manual work to search out the files with same names handle them very carefully.

shaoyie:
Just like stable sort and unstable sort, the result may vary but legal, depends on what is your expect.

legacy/builder/container_find_includes.go
```diff
@@ -144,16 +153,69 @@ func (s *ContainerFindIncludes) findIncludes(ctx *types.Context) error {
...
+			for source := range queue {
+				job(source)
+				atomic.AddInt32(&unhandled, -1)
+			}
+			wg.Done()
+		}()
+	}
+
+	// Loop until all files are handled
+	for (!sourceFilePaths.Empty() || unhandled != 0 ) {
```
cmaglie, line 195: Isn't this `unhandled` variable redundant here? the WaitGroup should already take care of waiting for all the processes to finish.

legacy/builder/container_find_includes.go
```diff
@@ -144,16 +153,69 @@ func (s *ContainerFindIncludes) findIncludes(ctx *types.Context) error {
 		queueSourceFilesFromFolder(ctx, sourceFilePaths, sketch, srcSubfolderPath, true /* recurse */)
 	}
 
-	for !sourceFilePaths.Empty() {
-		err := findIncludesUntilDone(ctx, cache, sourceFilePaths.Pop())
+	preloadCachedFolder(ctx, cache)
+
+	// The first source file is the main .ino.cpp
+	// handle it first to setup environment for other files
+	findIncludesUntilDone(ctx, cache, sourceFilePaths.Pop())
```
cmaglie, line 160: Why the `main.ino.cpp` is processed separately? If the processing order of the files doesn't matter, this should not be needed.

legacy/builder/container_find_includes.go
```diff
@@ -295,6 +334,46 @@ func writeCache(cache *includeCache, path *paths.Path) error {
...
+					appendIncludeFolderWoCache(ctx, header, library.SourceDir)
+					sourceDirs := library.SourceDirs()
+					for _, sourceDir := range sourceDirs {
+						queueSourceFilesFromFolder(ctx, ctx.CollectedSourceFiles, library, sourceDir.Dir, sourceDir.Recurse)
+					}
+				}
+			}
+		}
+		return true
+	})
```
cmaglie, line 369: If I read this correctly, you're basically reusing all the previous set of include paths, but without keeping the history of how you obtained that set. This won't work in all cases, in particular, if the cache is invalidated you must redo all the work again to be sure to not pick libraries that are not needed anymore.

legacy/builder/container_find_includes.go
shaoyie: The _unhandled_ is not redundant here. During the scanning process, we may find new source files. For example, suppose we have only 4 files to scan in the beginning, but we have 8 cores. Then the sourceFilePaths will be empty after the first goroutines fetch their task and the loop will be ended. But during the scanning  against the 4 files may append more new found files to the sourceFilePaths but they won't be handled. The variable _unhandled_ is to make sure sourceFilePaths is empty only after all goroutines finish the jobs.

legacy/builder/container_find_includes.go
shaoyie: In my original design, I also think the main.ino.cpp has no difference with other files. And it works in most cases.
But for some special cases, e.g. from the commits history of this PR we can see, the test case
assert run_command(["compile", "-b", "adafruit:samd:adafruit_feather_m4", sketch_path])
in test_compile_part_1.py will fail. 
Due to the library found mechanism in arduino-cli, have to set main.ino.cpp as the root of the scanning, then start other scanning from it.

legacy/builder/container_find_includes.go
shaoyie: Yes, in the particular case you mentioned, if need to clear the useless libraries, a clean build is required to do the job.

cmaglie:
Superseded by https://github.com/arduino/arduino-cli/pull/2625.

PR #1753, opened 2022-06-08
adding example usage for command option

TZubiri:
Hello, please take a look at this small documentation change.

Basically I struggled for a while to figure out why the monitor wasn't working, and I had to play with the IDE to understand that I needed to match the baud rates.

I'm adding a really low effort change that might help users work out how to use the `-c` option once they figured out they need to change the baud rate.

Ideally I think the options should be collapsed into command line flags so that the following works:

```text
arduino-cli monitor --baudrate 1000000
```

But for now I think this small change can be pulled, if it can be tested by someone with a working build.

If possible, I would also change the flag usage from

```text
-c, --config strings
```

to 

```text
-c --config ID=Setting
```

in addition to clarifying the syntax, it also points the user towards the `--describe` keywords.

Thank you.
Tom

### DIFF
--- a/i18n/data/en.po
+++ b/i18n/data/en.po
@@ -413,8 +413,8 @@ msgid "Config file written to: %s"
 msgstr "Config file written to: %s"
 
 #: cli/monitor/monitor.go:63
-msgid "Configuration of the port."
-msgstr "Configuration of the port."
+msgid "Configuration of the port, e.g.: baudrate=1000000"
+msgstr "Configuration of the port, e.g.: baudrate=1000000"
 
 #: cli/debug/debug.go:146
 msgid "Configuration options for %s"


per1234 (changes requested):
Thanks for your contribution @TZubiri.

This file is automatically generated so it can not be edited directly. You must instead edit the source string at the path and line indicated by the comment:

https://github.com/arduino/arduino-cli/blob/c52fcf8f7bf00bc0401904a42facaad33e16b896/i18n/data/en.po#L415

here:

https://github.com/arduino/arduino-cli/blob/c52fcf8f7bf00bc0401904a42facaad33e16b896/cli/monitor/monitor.go#L63

Any edits made to this `en.po` file would only be overwritten by the automated system on the next sync.

I have submitted a proposal to more effectively communicate this fact to contributors: https://github.com/arduino/arduino-cli/pull/1837
I welcome any feedback you might have on that proposal, so feel free to review that PR

alessio-perugini:
Closing this as it has been improved by https://github.com/arduino/arduino-cli/pull/2249

PR #1063, opened 2020-11-11
Fix update indexes running when not necessary

silvanocerza:
**Please check if the PR fulfills these requirements**

- [x] The PR has no duplicates (please search among the [Pull Requests](https://github.com/arduino/arduino-cli/pulls)
      before creating one)
- [x] The PR follows
      [our contributing guidelines](https://arduino.github.io/arduino-cli/latest/CONTRIBUTING/#pull-requests)
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

* **What kind of change does this PR introduce?**

An overall enhancement of the CLI behaviour.

* **What is the current behavior?**

If a local index file is missing the CLI tries to download it from the related additional URL whenever it tries to create a new internal instance, if that URL is unreachable for any reason the subsequent command always fails. 
This would happen when executing any command.

* **What is the new behavior?**

Indexes update is now only executed for certain specific commands like `update`, `core update-index`, `lib update-index` and `core search`, or whenever the `library_index.json` or `package_index.json` files are missing.

The initialization of an internal CLI instance still makes the whole command fail if a local index file is missing but the message is clearer now.

Updating indexes now shows a clear error if an URL is unreachable for any reason, we check for `4xx` or `5xx` HTTP status codes so in those cases it fails.

- **Does this PR introduce a breaking change?**

No.

* **Other information**:

None.

This should also fix #1050.

---

See [how to contribute](https://arduino.github.io/arduino-cli/latest/CONTRIBUTING/)

PR #1789, opened 2022-07-01
cli: use cached index with lib search (#1624)

ardnew:

      before creating one)
- [x] The PR follows

Feature

A new copy of `library_index.json` is downloaded every time `lib search` is called (unconditionally).

A new copy of `library_index.json` is downloaded every time `lib search` is called and either one of the following conditions exists: 
  1. The index has never been downloaded
  2. The time since the last index update is greater than one hour

No

* **Other information**:
  - Fixes: #1624
  - Changes based on: #1063
  - **Need assistance adding test cases and documentation for this change**

---

### DIFF
--- a/cli/lib/search.go
+++ b/cli/lib/search.go
@@ -204,7 +204,7 @@ func versionsFromSearchedLibrary(library *rpc.SearchedLibrary) []*semver.Version
 
 // indexNeedsUpdating returns whether library_index.json need updating.
 // A positive duration string must be provided to calculate the time threshold 
-// used to update the index (default: +24 hours).
+// used to update the index.
 // Valid duration units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
 // Use a duration of 0 to always update the index.
 func indexNeedsUpdating(duration string) bool {
--- a/cli/lib/search.go
+++ b/cli/lib/search.go
@@ -21,6 +21,7 @@ import (
 	"os"
 	"sort"
 	"strings"
+	"time"
 
 	"github.com/arduino/arduino-cli/cli/errorcodes"
 	"github.com/arduino/arduino-cli/cli/feedback"
@@ -28,7 +29,9 @@ import (
 	"github.com/arduino/arduino-cli/cli/output"
 	"github.com/arduino/arduino-cli/commands"
 	"github.com/arduino/arduino-cli/commands/lib"
+	"github.com/arduino/arduino-cli/configuration"
 	rpc "github.com/arduino/arduino-cli/rpc/cc/arduino/cli/commands/v1"
+	"github.com/arduino/go-paths-helper"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	semver "go.bug.st/relaxed-semver"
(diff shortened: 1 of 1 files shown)


cli/lib/search.go
```diff
@@ -193,3 +201,38 @@ func versionsFromSearchedLibrary(library *rpc.SearchedLibrary) []*semver.Version
...
+		feedback.Error(tr("Invalid timeout: %s", err))
+		os.Exit(errorcodes.ErrBadArgument)
+	}
+	// The behavior of now.After(T) is confusing if T < 0 and MTIME in the future,
+	// and is probably not what the user intended. Disallow negative T and inform
+	// the user that positive thresholds are expected.
+	if modTimeThreshold < 0 {
+		feedback.Error(tr("Timeout must be non-negative: %dns (%s)", modTimeThreshold, duration))
+		os.Exit(errorcodes.ErrBadArgument)
+	}
```
cmaglie, line 236: I would not bother too much with validating the timeout since it's a hardcoded constant.

```suggestion
```

cli/lib/search.go
```diff
@@ -193,3 +201,38 @@ func versionsFromSearchedLibrary(library *rpc.SearchedLibrary) []*semver.Version
 	sort.Sort(semver.List(res))
 	return res
 }
+
+// indexNeedsUpdating returns whether library_index.json need updating.
+// A positive duration string must be provided to calculate the time threshold
+// used to update the index.
+// Valid duration units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
+// Use a duration of 0 to always update the index.
+func indexNeedsUpdating(duration string) bool {
```
cmaglie, line 210: I'd like to change the `indexUpdateInterval` constant to a `time.Duration` like:
```go
const indexUpdateInterval = 60 * time.Minute
```
so the `indexNeedsUpdating` function may take directly a `time.Duration`

```suggestion
func indexNeedsUpdating(timeout time.Duration) bool {
```

cli/lib/search.go
```diff
@@ -193,3 +201,38 @@ func versionsFromSearchedLibrary(library *rpc.SearchedLibrary) []*semver.Version
...
+		os.Exit(errorcodes.ErrBadArgument)
+	}
+	// The behavior of now.After(T) is confusing if T < 0 and MTIME in the future,
+	// and is probably not what the user intended. Disallow negative T and inform
+	// the user that positive thresholds are expected.
+	if modTimeThreshold < 0 {
+		feedback.Error(tr("Timeout must be non-negative: %dns (%s)", modTimeThreshold, duration))
+		os.Exit(errorcodes.ErrBadArgument)
+	}
+	return modTimeThreshold == 0 || now.After(info.ModTime().Add(modTimeThreshold))
```
cmaglie, line 237: Also this one can be much more simplified, something like

```
return time.Since(info.ModTime()) > timeout
```

should work.

cmaglie (changes requested):
Thanks for the PR.
I see that you have taken the same patch for the `core search` and that's great!

I've made some comments to simplify this patch, IMHO after we validate these changes we can "backport" some of the suggestions to the `core search` command too.

Finally, the integration tests are failing, some of the tests expect the `Upgrading library_index...` message from the CLI, but this is no more happening after the patch. Could you fix the tests too? https://github.com/arduino/arduino-cli/runs/7155610048?check_suite_focus=true#step:10:6806

cmaglie:
I think that just removing these two lines https://github.com/arduino/arduino-cli/blob/f556aff7079ac5b02683f305329c90f8755c50f0/test/test_lib.py#L456-L457 should be enough to fix the integration tests.

PR #1878, opened 2022-09-16
[breaking] Fix behaviour of `lib` commands when a library is installed multiple times

cmaglie:

      before creating one)
- [X] The PR follows

Fix bugs reported in:
  - https://github.com/arduino/arduino-cli/issues/1870
  - https://github.com/arduino/arduino-cli/issues/1871
  - https://github.com/arduino/arduino-cli/issues/932

  1. `lib install` and `lib upgrade` may cause multiple installations of the same library
  2. if you use `lib install/upgrade/uninstall` on a library installed multiple times, the command may act only on one of the multiple copies (chosen pseudo-randomly).
  3. some commands do not use the real name of the library but the installed directory name, i.e. `lib example "Robot_Control"` instead of `lib example "Robot Control"`.

  1. `lib install` and `lib upgrade` will not cause multiple installations anymore
  2. if you use `lib install/upgrade/uninstall` on a library installed multiple times, the command will fail, showing the path of every copy of the library and asking to manually fix the multiple installations.
  3. all commands now use the real name of the library (i.e. the one written inside the `library.properties` file)

Yes. The biggest change is the renaming of the fields in the `Library` structure:
  * `Name` -> `CanonicalName`
  * `RealName` -> `Name`
  This change is quite involved in the go-lang codebase but results in a backward compatible change in the user-facing API. It should make the overall API usage more straightforward since it factors out the legacy (IDE <1.5) way to name the libraries as the containing directory.


internal/integrationtest/lib/lib_test.go
```diff
@@ -93,3 +93,27 @@ func TestLibInstallMultipleSameLibrary(t *testing.T) {
...
+	_, _, err := cli.Run("lib", "install", "ArduinoOTA@1.0.0")
+	require.NoError(t, err)
+	otaLibPath := cli.SketchbookDir().Join("libraries", "ArduinoOTA")
+	err = otaLibPath.CopyDirTo(otaLibPath.Parent().Join("CopyOfArduinoOTA"))
+	require.NoError(t, err)
+	jsonOut, _, err := cli.Run("lib", "list", "--format", "json")
+	require.NoError(t, err)
+	requirejson.Len(t, jsonOut, 2, "Duplicate library install is not detected by the CLI")
+
+	_, _, err = cli.Run("lib", "install", "ArduinoOTA")
```
umbynos, line 129: Wouldn't it be better to also check the error message?

umbynos (approved):

internal/integrationtest/lib/lib_test.go
cmaglie: 👍🏼

umbynos (approved):

kittaakos:
> * `Name` -> `CanonicalName`
> * `RealName` -> `Name`
>   This change is quite involved in the go-lang codebase but results in a backward compatible change in the user-facing API.

Would you consider applying the same change on the gRPC API?

https://github.com/arduino/arduino-cli/blob/5730e2ebbed39452a3a0517bb3571a1b94d864f7/rpc/cc/arduino/cli/commands/v1/lib.proto#L254

https://github.com/arduino/arduino-cli/blob/5730e2ebbed39452a3a0517bb3571a1b94d864f7/rpc/cc/arduino/cli/commands/v1/lib.proto#L283

IDE2 still uses both. `RealName` is available on the UI, and `Name` is the unique ID of the lib.

Should IDE2 apply any changes after closing https://github.com/arduino/arduino-cli/issues/932?

Thank you!

cmaglie:
Now both `Name` and `RealName` gets the same value: the library unique name (as written in the index / library.properties).
I forget to change the description on the field `Name`, will do it in a new PR.
I kept `RealName` for backward compatibility, so there should be no changes required in the IDE.
BTW it's probably better to remove the `RealName` field and cleanup the API, WDYT?

PR #1883, opened 2022-09-20
[skip-changelog] Use a matrix to create parallel builds for each OS

MatteoPologruto:

      before creating one)
- [x] The PR follows

Infrastructure Enhancement

[Release-go-task](https://github.com/arduino/arduino-cli/blob/master/.github/workflows/release-go-task.yml), [publish-go-nightly-task](https://github.com/arduino/arduino-cli/blob/master/.github/workflows/publish-go-nightly-task.yml) and [publish-go-tester-task](https://github.com/arduino/arduino-cli/blob/master/.github/workflows/publish-go-tester-task.yml) are currently using `task dist:all` to build and create the release artifacts. This is done by calling the tasks related to each OS consecutively. As a result, the finishing time of `task dist:all` is the sum of the finishing times of each individual `dist`. Furthermore, the checksums related to the files built are calculated more than once, unnecessarily.

**What is the new behavior?**

- Using a matrix to run each build task greatly improves performance, since they can all start concurrently. The finishing time of the job will be equal to the one of the longer task to build, instead of being the sum of each individual task's finishing time.
- Checksums are only being calculated once, during the final steps of the workflows.
- A check has been introduced to avoid creating the same changelog for each parallel building job.
- `publish-tester-build` now uploads a different artifact for each build during the first job, instead of creating a combined one that needs to be split later.

No

**Other information**:

Release run: https://github.com/MatteoPologruto/arduino-cli/actions/runs/3128175485
Test release: https://github.com/MatteoPologruto/arduino-cli/releases/tag/0.0.3
Nightly build run: https://github.com/MatteoPologruto/arduino-cli/actions/runs/3128099818
Tester build run: https://github.com/MatteoPologruto/arduino-cli/actions/runs/3137219983

[Publish-go-tester-task](https://github.com/arduino/arduino-cli/blob/master/.github/workflows/publish-go-tester-task.yml) was correctly skipped after `run-determination` when I pushed the tag `0.0.2` onto my fork ([here](https://github.com/MatteoPologruto/arduino-cli/actions/runs/3104779344)). Out of curiosity, I triggered the workflow run manually on the same commit and the checksums' calculation was not correct ([here](https://github.com/MatteoPologruto/arduino-cli/actions/runs/3105007409)). Checksums calculated previously and afterwards seem to be correct, even though the workflow has not been modified. I think it might be worth mentioning.

---

### DIFF
--- a/.github/workflows/publish-go-nightly-task.yml
+++ b/.github/workflows/publish-go-nightly-task.yml
@@ -179,6 +179,13 @@ jobs:
           TAG="nightly-$(date -u +"%Y%m%d")"
           sha256sum ${{ env.PROJECT_NAME }}_${TAG}* >> ${TAG}-checksums.txt
 
+      - name: Upload artifacts
+        uses: actions/upload-artifact@v3
+        with:
+          if-no-files-found: error
+          name: ${{ env.ARTIFACT_NAME }}
+          path: ${{ env.DIST_DIR }}
+
       # - name: Upload release files on Arduino downloads servers
       #   uses: docker://plugins/s3
       #   env:
--- a/.github/workflows/publish-go-nightly-task.yml
+++ b/.github/workflows/publish-go-nightly-task.yml
@@ -179,13 +179,6 @@ jobs:
           TAG="nightly-$(date -u +"%Y%m%d")"
           sha256sum ${{ env.PROJECT_NAME }}_${TAG}* >> ${TAG}-checksums.txt
 
-      - name: Upload artifacts
-        uses: actions/upload-artifact@v3
-        with:
-          if-no-files-found: error
-          name: ${{ env.ARTIFACT_NAME }}
-          path: ${{ env.DIST_DIR }}
-
       # - name: Upload release files on Arduino downloads servers
       #   uses: docker://plugins/s3
       #   env:
--- a/.github/workflows/publish-go-nightly-task.yml
+++ b/.github/workflows/publish-go-nightly-task.yml
@@ -22,6 +22,19 @@ jobs:
   create-nightly-artifacts:
     runs-on: ubuntu-latest
 
+    strategy:
+      matrix:
+        os:
+          - Windows_32bit
+          - Windows_64bit
+          - Linux_32bit
+          - Linux_64bit
+          - Linux_ARMv6
+          - Linux_ARMv7
+          - Linux_ARM64
+          - macOS_64bit
+          - macOS_ARM64
+
     steps:
       - name: Checkout repository
         uses: actions/checkout@v3
@@ -35,7 +48,7 @@ jobs:
       - name: Build
         env:
           NIGHTLY: true
-        run: task dist:all
+        run: task dist:${{ matrix.os }}
 
       - name: Upload artifacts
         uses: actions/upload-artifact@v3
--- a/.github/workflows/publish-go-nightly-task.yml
+++ b/.github/workflows/publish-go-nightly-task.yml
@@ -179,31 +179,31 @@ jobs:
           TAG="nightly-$(date -u +"%Y%m%d")"
           sha256sum ${{ env.PROJECT_NAME }}_${TAG}* >> ${TAG}-checksums.txt
 
-      - name: Upload release files on Arduino downloads servers
-        uses: docker://plugins/s3
-        env:
-          PLUGIN_SOURCE: "${{ env.DIST_DIR }}/*"
-          PLUGIN_TARGET: "${{ env.AWS_PLUGIN_TARGET }}nightly"
-          PLUGIN_STRIP_PREFIX: "${{ env.DIST_DIR }}/"
-          PLUGIN_BUCKET: ${{ secrets.DOWNLOADS_BUCKET }}
-          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
-          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+      # - name: Upload release files on Arduino downloads servers
+      #   uses: docker://plugins/s3
+      #   env:
+      #     PLUGIN_SOURCE: "${{ env.DIST_DIR }}/*"
+      #     PLUGIN_TARGET: "${{ env.AWS_PLUGIN_TARGET }}nightly"
+      #     PLUGIN_STRIP_PREFIX: "${{ env.DIST_DIR }}/"
+      #     PLUGIN_BUCKET: ${{ secrets.DOWNLOADS_BUCKET }}
+      #     AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+      #     AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
 
-  report:
-    runs-on: ubuntu-latest
-    needs: publish-nightly
-    if: failure() # Run if publish-nightly or any of its job dependencies failed
+  #  report:
+  #    runs-on: ubuntu-latest
+  #    needs: publish-nightly
+  #    if: failure() # Run if publish-nightly or any of its job dependencies failed
 
-    steps:
-      - name: Report failure
-        uses: masci/datadog@v1
-        with:
-          api-key: ${{ secrets.DD_API_KEY }}
-          events: |
-            - title: "${{ env.PROJECT_NAME }} nightly build failed"
-              text: "Nightly build workflow has failed"
-              alert_type: "error"
-              host: ${{ github.repository }}
-              tags:
-                - "project:${{ env.PROJECT_NAME }}"
-                - "workflow:${{ github.workflow }}"
+  #    steps:
+  #      - name: Report failure
+  #        uses: masci/datadog@v1
+  #        with:
+  #          api-key: ${{ secrets.DD_API_KEY }}
+  #          events: |
+  #            - title: "${{ env.PROJECT_NAME }} nightly build failed"
+  #              text: "Nightly build workflow has failed"
+  #              alert_type: "error"
+  #              host: ${{ github.repository }}
+  #              tags:
+  #                - "project:${{ env.PROJECT_NAME }}"
+  #                - "workflow:${{ github.workflow }}"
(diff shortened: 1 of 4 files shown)


umbynos (changes requested):
Wow! The improvement in time is huge! :bullettrain_front: 
Unfortunately this is not sufficient: we need to find another approach for the checksum generation, [because the generated one is not complete](https://github.com/MatteoPologruto/arduino-cli/releases/download/0.0.0/0.0.0-checksums.txt).
The `task dist:all` was appending the checksums in the same file and then uploading it (since the task were executed sequentially it was possible). Unfortunately this is no longer working.
I suggest removing the [checksum calculation from the tasks](https://github.com/arduino/arduino-cli/blob/master/DistTasks.yml#L51), from the [notarization step](https://github.com/arduino/arduino-cli/blob/master/.github/workflows/release-go-task.yml#L152). And perform it only during the [release creation job](https://github.com/arduino/arduino-cli/blob/a3f6574ab09d19070defadb3be91e5c46fc916b1/.github/workflows/release-go-task.yml#L174-L182)

umbynos:
Also it does not make sense to perform [the changelog generation](https://github.com/arduino/arduino-cli/blob/5b25da356257277734ddab3fd494cb47a274d493/.github/workflows/release-go-task.yml#L28-L34) in all the jobs.. could be an idea to do it only for one of the runs

umbynos:
I've just noticed an unrelated problem with the release CI: https://github.com/arduino/arduino-cli/blob/a3f6574ab09d19070defadb3be91e5c46fc916b1/DistTasks.yml#L214

This should be:
`PLATFORM_DIR: "{{.PROJECT_NAME}}_linux_arm_64"`
It's worth to fix it in this PR IMHO

.github/workflows/publish-go-nightly-task.yml
```diff
@@ -167,13 +177,13 @@ jobs:
           path: ${{ env.DIST_DIR }}
 
       - name: Update checksum
```
umbynos, line 179:
```suggestion
      - name: Output checksum
```
leftover

.github/workflows/publish-go-nightly-task.yml
```diff
@@ -144,9 +157,6 @@ jobs:
           tar -czvf "$PACKAGE_FILENAME" \
           -C "${{ env.PROJECT_NAME }}_osx_${{ matrix.artifact.name }}/" "${{ env.PROJECT_NAME }}" \
           -C ../../ LICENSE.txt
-          CHECKSUM_LINE="$(shasum -a 256 $PACKAGE_FILENAME)"
```
umbynos, line 147: You forgot to remove checksum in the name of this step:
https://github.com/arduino/arduino-cli/pull/1883/files#diff-b01c8ff87205e83918583aff246152ef27d286cccaa14409dddc17bd37104518R143

.github/workflows/publish-go-nightly-task.yml
```diff
@@ -144,9 +157,6 @@ jobs:
...
-          echo "PACKAGE_FILENAME=$PACKAGE_FILENAME" >> $GITHUB_ENV
```
umbynos, line 148: `PACKAGE_FILENAME` was used in the step below for a reason: we used to upload only the notarized binary. This way [we are overwriting all the binaries](https://github.com/MatteoPologruto/arduino-cli/actions/runs/3105467121/jobs/5031149531#step:9:13): which means that either the arm64 or the amd64 one is not notarized anymore

.github/workflows/publish-go-nightly-task.yml
```diff
@@ -144,9 +157,6 @@ jobs:
...
-          echo "::set-output name=checksum-${{ matrix.artifact.name }}::$CHECKSUM_LINE"
 
       - name: Upload artifacts
```
umbynos, line 161:
```suggestion
      - name: Upload artifact
```
typo

.github/workflows/publish-go-nightly-task.yml
umbynos: One way to enhance this further and remove completely this kind of errors would be to only download the artifact we are going to notarize and not all of them [like we do now](https://github.com/arduino/arduino-cli/pull/1883/files#diff-b01c8ff87205e83918583aff246152ef27d286cccaa14409dddc17bd37104518R83-R87). And then upload only the notarized one [here](https://github.com/arduino/arduino-cli/pull/1883/files#diff-b01c8ff87205e83918583aff246152ef27d286cccaa14409dddc17bd37104518R166)

.github/workflows/publish-go-nightly-task.yml
```diff
@@ -167,13 +177,13 @@ jobs:
...
-          for checksum_line in "${checksum_lines[@]}"
+          TAG="nightly-$(date -u +"%Y%m%d")"
+          declare -a artifacts=$(ls | grep "^${{ env.PROJECT_NAME }}_${TAG}")
+          for artifact in "${artifacts[@]}"
           do
-            CHECKSUM=$(echo ${checksum_line} | cut -d " " -f 1)
-            PACKAGE_FILENAME=$(echo ${checksum_line} | cut -d " " -f 2)
-            perl -pi -w -e "s/.*${PACKAGE_FILENAME}/${CHECKSUM}  ${PACKAGE_FILENAME}/g;" ${{ env.DIST_DIR }}/*-checksums.txt
+            sha256sum $artifact >> ${TAG}-checksums.txt
           done
```
umbynos, line 187: I've just discovered that this also works fine:
```
sha256sum ${{ env.PROJECT_NAME }}_${TAG}* >> ${TAG}-checksums.txt
```

.github/workflows/release-go-task.yml
```diff
@@ -132,14 +147,11 @@ jobs:
         run: |
           gon "${{ env.GON_CONFIG_PATH }}"
 
-      - name: Re-package binary and output checksum
+      - name: Re-package binary
         id: re-package
         working-directory: ${{ env.DIST_DIR }}
         # This step performs the following:
         # 1. Repackage the signed binary replaced in place by Gon (ignoring the output zip file)
```
umbynos, line 154:
```suggestion
        # Repackage the signed binary replaced in place by Gon (ignoring the output zip file)
```
It's useless to have a numbered list with only a number

umbynos: This also applies to https://github.com/arduino/arduino-cli/pull/1883/files#diff-b01c8ff87205e83918583aff246152ef27d286cccaa14409dddc17bd37104518R146-R150

.github/workflows/release-go-task.yml
```diff
@@ -149,9 +161,6 @@ jobs:
           tar -czvf "$PACKAGE_FILENAME" \
           -C "${{ env.PROJECT_NAME }}_osx_${{ matrix.artifact.name }}/" "${{ env.PROJECT_NAME }}" \
           -C ../../ LICENSE.txt
-          CHECKSUM_LINE="$(shasum -a 256 $PACKAGE_FILENAME)"
-          echo "PACKAGE_FILENAME=$PACKAGE_FILENAME" >> $GITHUB_ENV
```
umbynos, line 153: same suggestion I did [here](https://github.com/arduino/arduino-cli/pull/1883/files#diff-b01c8ff87205e83918583aff246152ef27d286cccaa14409dddc17bd37104518L148)

.github/workflows/release-go-task.yml
```diff
@@ -149,9 +161,6 @@ jobs:
...
-          echo "::set-output name=checksum-${{ matrix.artifact.name }}::$CHECKSUM_LINE"
 
       - name: Upload artifacts
```
umbynos, line 165:
```suggestion
      - name: Upload artifact
```

.github/workflows/publish-go-tester-task.yml
```diff
@@ -122,6 +135,23 @@ jobs:
...
+          if [ "${{ github.event_name }}" = "pull_request" ]; then
+            PACKAGE_NAME_PREFIX="$PACKAGE_NAME_PREFIX-${{ github.event.number }}"
+          fi
+          PACKAGE_NAME_PREFIX="$PACKAGE_NAME_PREFIX-${{ github.sha }}-"
+          TAG="${PACKAGE_NAME_PREFIX}git-snapshot"
+          declare -a artifacts=$(ls | grep "^arduino-cli_${TAG}")
+          for artifact in "${artifacts[@]}"
+          do
+            sha256sum $artifact >> ${TAG}-checksums.txt
+          done
```
umbynos, line 153: I've just discovered that this also works fine:
```
sha256sum ${{ env.PROJECT_NAME }}_${TAG}* >> ${TAG}-checksums.txt
```

.github/workflows/release-go-task.yml
```diff
@@ -171,14 +180,14 @@ jobs:
...
-          for checksum_line in "${checksum_lines[@]}"
+          TAG="${GITHUB_REF/refs\/tags\//}"
+          declare -a artifacts=$(ls | grep "^${{ env.PROJECT_NAME }}_${TAG}")
+          for artifact in "${artifacts[@]}"
           do
-            CHECKSUM=$(echo ${checksum_line} | cut -d " " -f 1)
-            PACKAGE_FILENAME=$(echo ${checksum_line} | cut -d " " -f 2)
-            perl -pi -w -e "s/.*${PACKAGE_FILENAME}/${CHECKSUM}  ${PACKAGE_FILENAME}/g;" ${{ env.DIST_DIR }}/*-checksums.txt
+            sha256sum $artifact >> ${TAG}-checksums.txt
           done
```
umbynos, line 191: I've just discovered that this also works fine:
```
sha256sum ${{ env.PROJECT_NAME }}_${TAG}* >> ${TAG}-checksums.txt
```

umbynos (changes requested):
Left some suggestions :space_invader:

.github/workflows/publish-go-nightly-task.yml
MatteoPologruto: Unfortunately, it is not possible to download the single artifact to be notarized since only one artifact named `dist` is uploaded in the previous job. One way to do it could be to upload every artifact generated in the building process separately, but it would mean that they'll need to be treated individually when calculating the checksum. I chose to simply reintroduce the above mentioned line.

umbynos (approved):
Great Job Matteo :1st_place_medal:

PR #1887, opened 2022-09-21
Add prioritization to tool selection in absence of explicit dependencies

cmaglie:

      before creating one)
- [X] The PR follows

It removes some ambiguities in compile/upload tools selection when the explicit dependencies set (from `package_index.json`) is missing.

In a situation when different platforms provide the same tool (for example `arm-gcc-none-eabi` from `arduino` and other 3rd parties), and in absence of the `package_index.json`, the selection of the tool in the compile/upload may not be deterministic (AKA: random).

  In the situation above where a tool is provided by multiple packages then the compiler will pick in order of priority:
   * the tool provided by the board's package
   * in absence of it, the tool (from a 3rd party) with the greatest version according to semver
   * in case of a tie, the tool from the packager whose name comes first in alphabetic order
  The last rule is merely a way to make the tool selection fully deterministic even in the most uncommon situation.

No


cmaglie:
I still need to update documentation, I'll do that after validating the changes.

jantje:
>the tool provided by the board's package

Should this not be the boards platform (instead of the board's package)?
Is that the selected platform only or the selected platform and the referenced platform?
And if the selected platform and the referenced platform both have the tool (Which is possible but I have never seen it) which is selected?

>in absence of it, the tool (from a 3rd party) with the greatest version according to semver

FYI
If I remember correctly I had a similar implementation in Sloeber one day and it caused lots of boards to fail build yet they build successfully in Arduino IDE.

cmaglie:
> Should this not be the boards platform (instead of the board's package)?

It's the same since the tools are defined at the package level (and referenced at platform level as `toolsDependecies`).
After re-reading the PR I see that my first explanation was a bit rushed... I made a much better explanation in the documentation changes: https://github.com/arduino/arduino-cli/pull/1887/files#diff-29b8b99faa5765ef9cc15dbe6010806716089af4475d8483766b02c237f33753R274-R288
here is the extract:

> When the IDE needs a tool, it downloads the corresponding archive file and unpacks the content into a private folder
that can be referenced from `platform.txt` using one of the following properties:
>
> - `{runtime.tools.TOOLNAME-VERSION.path}`
> - `{runtime.tools.TOOLNAME.path}`
> 
> For example, to obtain the avr-gcc 4.8.1 folder we can use `{runtime.tools.avr-gcc-4.8.1.path}` or
`{runtime.tools.avr-gcc.path}`.
>
> In general the same tool may be provided by different platforms (for example the Arduino AVR platform may provide an
`arduino:avr-gcc` and another 3rd party platform may provide their own `3rdparty:avr-gcc`). The rules to disambiguate
are as follows:
>
> - The property `{runtime.tools.TOOLNAME.path}` points, in order of priority, to:
>
>  1. the tool, version and packager specified via `toolsDependencies` in the `package_index.json`
>  1. the highest version of the tool provided by the packager of the current platform
>  1. the highest version of the tool provided by any other packager (in case of tie, the first packager in alphabetical
>     order wins)
>
> - The property `{runtime.tools.TOOLNAME-VERSION.path}` points, in order of priority, to:
>   1. the tool and version provided by the packager of the current platform
>   1. the tool and version provided by any other packager (in case of tie, the first packager in alphabetical order wins)

I should change the sentence:

> In general the same tool may be provided by different platforms (for example the Arduino AVR platform may provide an
`arduino:avr-gcc` and another 3rd party platform may provide their own `3rdparty:avr-gcc`). The rules to disambiguate
are as follows:

to

> In general the same tool may be provided by different packagers (for example the Arduino packager may provide an
`arduino:avr-gcc` and another 3rd party platform may provide their own `3rdparty:avr-gcc`). The rules to disambiguate
are as follows:

In practice the idea is that when the `toolsDependencies` is not defined or not available, we first try to select the tools from the same packager or, if not found, from another packager.

---

About the other question:

> Is that the selected platform only or the selected platform and the referenced platform?

That is a really good point, I didn't consider the referenced platform, and I must add that in this PR.

> And if the selected platform and the referenced platform both have the tool (Which is possible but I have never seen it) which is selected?

The selected platform of course.

Thanks for the feedback @jantje

umbynos (approved):
Code LGTM

matthijskooijman:
I just tested git master and this also solves an issue I had with a git checkout of the stm32duino core that had been annoying for quite some time (but I could fix it by just retrying and it would randomly work or not work, so wasn't annoying enough to investigate ;-p). Thanks!

PR #1931, opened 2022-10-19
Add task to collect proto files and add steps to release them with tester, nightly and stable builds

### LINKED ISSUE #588: [proposal] Release the `proto` files with the CLI
### Current behavior

As a gRPC consumer of the CLI, I have to generate the JS/TS modules from the `proto` files. It would be nice to ship the releases with the `proto` files so that one can make sure the CLI is compatible with the gRPC API.

As a possible workaround for this limitation:
 - one can download the CLI,
 - run `./arduino-cli version`,
 - parse the output:
```
./arduino-cli version
arduino-cli Version: 0.8.0 Commit: 64c501d
```
 - clone arduino/arduino-cli and checkout the 64c501d commit,
 - run the  TS/JS generation from the  proto files

### Environment

- CLI version (output of `arduino-cli version`):
- OS and platform:

silvanocerza:

      before creating one)
- [x] The PR follows

Adds a new task and changes the tester, nightly and stable build release processes.

Currently there is no easy way for developers to download only `.proto` files to work on projects that depend on them. 
Those files are necessary to generate other languages files to handle gRPC communication.
The only way as of now is to download the whole `arduino-cli` repo and copy the files.

## What is the new behavior?

A new command has been added to zip all the `.proto` files and save them in `dist`, the task is called `protoc:collect`.
The output zip will be called `arduino-cli_<version>_proto.zip`, `<version>` as defined [here](https://github.com/arduino/arduino-cli/blob/master/Taskfile.yml#L351). The command can be run locally and on CI.

This zip will also be included in the released files for tester, nightly and stable builds.

None.

## Other information

Closes #588.

### DIFF
--- a/.github/workflows/publish-go-tester-task.yml
+++ b/.github/workflows/publish-go-tester-task.yml
@@ -132,6 +132,32 @@ jobs:
           path: ${{ env.DIST_DIR }}/${{ matrix.os.path }}
           name: ${{ matrix.os.name }}
 
+  proto-files:
+    needs:
+      - package-name-prefix
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Install Task
+        uses: arduino/setup-task@v1
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          version: 3.x
+
+      - name: Collect proto files
+        run: task protoc:collect
+
+      - name: Upload proto files zip
+        uses: actions/upload-artifact@v3
+        with:
+          path: ${{ env.DIST_DIR }}/*.zip
+          name: rpc-protocol-files
+
   checksums:
     needs:
       - build
--- a/.github/workflows/publish-go-tester-task.yml
+++ b/.github/workflows/publish-go-tester-task.yml
@@ -71,41 +71,44 @@ jobs:
 
           echo "::set-output name=prefix::$PACKAGE_NAME_PREFIX"
 
-  build:
+  create-artifacts:
     needs: package-name-prefix
-    name: Build ${{ matrix.os.name }}
+    name: Create artifact ${{ matrix.artifact.name }}
     runs-on: ubuntu-latest
 
     strategy:
       matrix:
-        os:
-          - task: Windows_32bit
+        artifact:
+          - task: dist:Windows_32bit
             path: "*Windows_32bit.zip"
             name: Windows_X86-32
-          - task: Windows_64bit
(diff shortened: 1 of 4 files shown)


silvanocerza:
Tester build with proto files: https://github.com/arduino/arduino-cli/actions/runs/3282804966

image

.github/workflows/publish-go-tester-task.yml
```diff
@@ -159,3 +159,35 @@ jobs:
         with:
           path: ./*checksums.txt
           name: checksums
+
+  proto-files:
+    # Run this after checksums or the uploaded proto files zip
+    # will make the checksum job always fail
+    needs:
+      - checksums
+      - package-name-prefix
```
cmaglie, line 168: Why `checksums` is failing? shouldn't just add the checksum of the protoc archive to the checksums file together with the other archives?

silvanocerza: I didn't think about that, it would make sense since it's a file part of the release. :thinking: 
I'd have to edit a bit the checksum process I think, I'll take a look at it.

.github/workflows/publish-go-tester-task.yml
```diff
@@ -132,9 +132,39 @@ jobs:
...
+          PACKAGE_NAME_PREFIX=${{ needs.package-name-prefix.outputs.prefix }}
+          export PACKAGE_NAME_PREFIX
+          task protoc:collect
+
+      - name: Upload proto files zip
+        uses: actions/upload-artifact@v3
+        with:
+          path: ${{ env.DIST_DIR }}/*.zip
+          name: rpc-protocol-files
+
```
umbynos, line 163: what do you think about adding this as an additional step in the matrix of `build` job? It seems like almost all the steps are the same, we could avoid some code duplication.

silvanocerza: I don't think it's semantically correct, this job doesn't build anything in itself. 
It also would require some changes to that job to make it more generic if we want to go toward that direction too.

umbynos: I wouldn't duplicate the code that way. There are some things that could be made in order to achieve that:
- If we move the task `protoc:collect` to `DistTask.yml` the workflow do not require any big change.
- another idea would be to simply rename the task as `dist:protoc-collect`
- or maybe just move `dist:` in the `matrix.os.task`

silvanocerza: Wasn't `DistTask.yml`shared in other projects too? I thought of putting it there but didn't cause of that.

Second and third might work but we'd still need to reword the `build` job and `os` matrix variable to be generic.

Issue #340, opened 2019-08-09
Requesting ARM6 build for Raspberry Pi Zero


Tatwi:
Your install.sh script indicated the following: 

"Sorry, we dont have a dist for your system: Linux armv6
You can request one here: https://github.com/Arduino/arduino-cli/issues"

As such, I would like to request a build for ARM6 that is compatible with the Raspberry Pi Zero (and Zero W). 

Thanks! :)

per1234:
Thanks for reporting this!

To tide you over until the installation script is fixed, there is a Raspberry Pi compatible version of arduino-cli available at the "arduino-cli_0.4.0_Linux_32bit.tar.gz" link here:
https://github.com/arduino/arduino-cli/releases/tag/0.4.0

masci:
Hi @Tatwi, at this stage we only distribute binaries for ARMv7 and later, this is something that might change if users ask for it but I'm not sure about implications right now. As a workaround, have you tried building the CLI yourself? The [contributing page](https://github.com/arduino/arduino-cli/blob/master/CONTRIBUTING.md#prerequisites) provides some guidance but feel free to reach out should you have any issue with the build.

Let's use this issue to collect feature requests, if anybody else needs an ARMv6 distribution please leave your feedback or upvote OP.

Tatwi:
Thanks masci. I'll build it myself when I have time to put together the toolchain (I don't use the software that you use, so I don't have it setup).

My thoughts on why having an ARM6 build would be beneficial are thus:

1. The Raspberry Pi Zero / Zero W will be manufactured and available until at least 2026.

2. Of all the Raspberry Pi products, the Pi Zero best synergies with Arduino products, being that it is often used in embedded systems, robotics, etc. It's less than half the size of an Arduino UNO and it integrates nicely with Arduino Nano devices.

3. Of all the Raspberry Pi products, the ARM6 based Pi Zero is the one that benefits the most from using the command line version of the Arduino compiler, because often the Pi Zero will be used either headless or command line only.

Personally I am just slapping my UNO together with my Zero for the sake of doing so, but I'm sure there are others out there who are enjoying the combination of computing power, small size, low power requirements, and ability to program multiple Arduinos on the fly ("Smart" robots for instance). I think it's worth doing, at least until the Pi Zero is no longer being sold as an ARM6 device.

facchinm:
[...]

PR #1937, opened 2022-10-21
[skip-changelog] Add ARMv6 download links to install docs

mik13ST:
As a newcomer, this made me assume ARMv6 is not supported so I searched the issues, found https://github.com/arduino/arduino-cli/issues/340, which suggested building from source so I did, only to discover an ARMv6 binary in the release files.

As for the table layout, it's not ideal. It would also make sense to have this in another column.


cmaglie (approved):

cmaglie:
Thank you for the PR!

> As for the table layout, it's not ideal. It would also make sense to have this in another column.

You're right, but it works for me, even if not 100.00% correct.

PR #1953, opened 2022-10-27
[skip-changelog] Run integration tests concurrently

MatteoPologruto:

      before creating one)
- [x] The PR follows

Infrastructure enhancement

Integration tests are run consecutively and each test has to wait for the previous ones to end before starting, making the whole process slower any time a new test is added.
Unit tests and integration tests are executed by different jobs of the workflow.

## What is the new behavior?
Using a matrix to run each package test concurrently reduces the waiting time. `GO_TEST_PACKAGE` is set to the test package name to identify the single package to run.

No


Taskfile.yml
```diff
@@ -91,8 +91,6 @@ tasks:
   # Source: https://github.com/arduino/tooling-project-assets/blob/main/workflow-templates/assets/test-go-task/Taskfile.yml
   go:test:
     desc: Run unit tests and the Go-based integration tests
-    deps:
-      - task: go:build
```
umbynos, line 95: why this is required?

.github/workflows/test-go-task.yml
```diff
@@ -92,8 +104,17 @@ jobs:
...
+
       - name: Run tests
-        run: task go:test
+        shell: bash
+        run: |
+          GO_TEST_REGEX=[^${{ matrix.tests }}]
+          export GO_TEST_REGEX
+          GO_MODULE_PATH=$(dirname $(ls ./internal/integrationtest/*/${{ matrix.tests }}))
+          export GO_MODULE_PATH
+          task go:test
```
umbynos, line 117:
```suggestion
        shell: bash
        run: |
          export GO_TEST_REGEX=[^${{ matrix.tests }}]
          export GO_MODULE_PATH=$(dirname $(ls ./internal/integrationtest/*/${{ matrix.tests }}))
          task go:test
```
this should work

umbynos (changes requested):

Taskfile.yml
MatteoPologruto: I had to remove it because, if the parent directory of the test file is specified as `GO_MODULE_PATH`, `task go:build` returns this error:

> go build github.com/arduino/arduino-cli/internal/integrationtest/compile: no non-test Go files in /Users/runner/work/arduino-cli/arduino-cli/internal/integrationtest/compile

You can check this workflow run: https://github.com/MatteoPologruto/arduino-cli/actions/runs/3329851322/jobs/5507606209
I tried to solve this issue building the cli in a workflow step before executing `task go:test`.

MatteoPologruto:
I made some changes to avoid specifying a different go path. It is no longer needed to build the cli in a specific step of the workflow.

.github/tools/get_go_integration_tests.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+if __name__ == "__main__":
+    import sys
+
+    tests_path = sys.argv[1]
+
+    test_files = [str(f).split("/")[-2] for f in Path(tests_path).glob("*/*_test.go")]
+    test_files = list(dict.fromkeys(test_files))
+    print(json.dumps(test_files))
```
cmaglie, line 12: Can't we run `go list | grep /integrationtest/` instead of another python script?

Taskfile.yml
```diff
@@ -99,10 +99,11 @@ tasks:
         go test \
           -v \
           -short \
+          {{ .GO_TEST_PACKAGE }} \
           -run '{{default ".*" .GO_TEST_REGEX}}' \
           {{default "-timeout 20m -coverpkg=./... -covermode=atomic" .GO_TEST_FLAGS}} \
           -coverprofile=coverage_unit.txt \
-          {{default .DEFAULT_GO_PACKAGES .GO_PACKAGES}} \
+          {{default .DEFAULT_INTEGRATIONTEST_GO_PACKAGES .GO_PACKAGES}} \
```
cmaglie, line 106: Doing so removes the unit-tests from the test suite.

Before `DEFAULT_GO_PACKAGES` had both unit and integration tests, after the split `DEFAULT_GO_PACKAGES` now has the unit tests and `DEFAULT_INTEGRATIONTEST_GO_PACKAGES` has the integration tests. BTW this change is not reflected in the TaskFile targets, I'd expect to have a `go:test` target to run unit-tests and a `go:integration-test` to run the unit tests.

The same separation should be done in the github workflows, the integration-test should be run under the os+package matrix, while the unit test should be under the os matrix.

cmaglie (changes requested):

.github/workflows/test-go-task.yml
```diff
@@ -58,13 +58,61 @@ jobs:
...
+          version: 3.x
+
+      - name: Run tests
+        shell: bash
+        run: |
+          export GO_TEST_PACKAGE="github.com/arduino/arduino-cli/internal/integrationtest/${{ matrix.tests }}"
+          task go:integration-test
+
+  test:
+    needs: test-integration
```
cmaglie, line 114: Why is needed `test-integration`?

.github/workflows/test-go-task.yml
```diff
@@ -93,7 +141,8 @@ jobs:
           version: 3.x
 
       - name: Run tests
-        run: task go:test
+        run: |
+          task go:test
```
cmaglie, line 145: remove the diff

Taskfile.yml
```diff
@@ -105,6 +105,23 @@ tasks:
           {{default .DEFAULT_GO_PACKAGES .GO_PACKAGES}} \
           {{.TEST_LDFLAGS}}
 
+  go:integration-test:
+    desc: Run unit tests and the Go-based integration tests
```
cmaglie, line 109:
```suggestion
    desc: Run the Go-based integration tests
```

cmaglie (changes requested):
Just a couple of nitpicks but I think we are there. 👍🏼

cmaglie (approved):

umbynos (approved):

PR #1982, opened 2022-11-17
Improve `board list` detection via cloud API

### LINKED ISSUE #1963: Provide an `--offline` directive to prevent errors in portable mode
### Describe the request

I have been testing a setup for portable mode of Arduino CLI.
I am happy to know that (at least on Mac OS) it works beautifully, but when I have an unrecognised device attached to the USB port, it tries to resolve the `fqbn` using the Arduino online builder.
This generates an error which is unpleasant.
I am not sure what other calls to services are made besides updating index for libraries and platforms, but in theory a user might decide to manually copy the platforms and related index files into a portable directory and run `arduino-cli` from there

with the following file structure, `arduino-cli` can be run locally and off removable storage.
The requirements are that any platform and its supporting package index can be found by the CLI.
A local `arduino-cli.yaml` is required and must be properly customised (follows)

CleanShot 2022-11-04 at 09 49 03@2x

I have manually copied what was needed from my `Arduino15` folder and defined user and data paths for the configuration `.yaml` file

**arduino-cli.yaml**
```
board_manager:
  additional_urls:
  - https://dl.espressif.com/dl/package_esp32_index.json
daemon:
  port: "50051"
directories:
  data: ./Arduino15
  downloads: ./Arduino15/staging
  user: ./Arduino
library:
  enable_unsafe_install: false
logging:
  file: ""
  format: text
  level: info
metrics:
  addr: :9090
  enabled: true
output:
  no_color: false
sketch:
  always_export_binaries: true
updater:
  enable_notification: true
```

### Arduino CLI version

test-1845-3fd2864ff8d77d476d1e73500f8a0c54d03354b3-git-snapshot

### Operating system

macOS

### Operating system version

12.6

### Additional context

_No response_

### Issue checklist

- [X] I verified the feature was still missing when using the [nightly build](https://arduino.github.io/arduino-cli/dev/installation/#nightly-builds)

cmaglie:

      before creating one)
- [X] The PR follows

* calls to builder-api are cached 24h
* do not error on `board list` if the network is not available (fix #1963)

`board list` terminate with an error if the network is not available and the board is not "known" (there are no platform installed for that board).

## What is the new behavior?

The `board list` command always complete even if the network is not available

No

### DIFF
--- a/commands/board/list.go
+++ b/commands/board/list.go
@@ -32,6 +32,7 @@ import (
 	"github.com/arduino/arduino-cli/arduino/discovery"
 	"github.com/arduino/arduino-cli/arduino/httpclient"
 	"github.com/arduino/arduino-cli/commands"
+	"github.com/arduino/arduino-cli/inventory"
 	rpc "github.com/arduino/arduino-cli/rpc/cc/arduino/cli/commands/v1"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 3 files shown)


cmaglie:
Actually, it doesn't strictly fix #1963, but if I read correctly between the lines, this is what @ubidefeo was asking for.

commands/board/list.go
```diff
@@ -72,39 +71,41 @@ func apiByVidPid(vid, pid string) ([]*rpc.BoardListItem, error) {
...
-
-		var dat map[string]interface{}
-		err = json.Unmarshal(body, &dat)
-		if err != nil {
-			return nil, errors.Wrap(err, tr("error processing response from server"))
+	res, err := httpClient.Do(req)
+	if err != nil {
+		return nil, errors.Wrap(err, tr("error querying Arduino Cloud Api"))
+	}
+	if res.StatusCode >= 400 {
```
umbynos, line 112: Is this check really required?

commands/board/list.go
```diff
@@ -50,6 +51,39 @@ var (
...
+func cachedApiByVidPid(vid, pid string) ([]*rpc.BoardListItem, error) {
+	var resp []*rpc.BoardListItem
+
+	cacheKey := fmt.Sprintf("cache.builder-api.v3/boards/byvid/pid/%s/%s", vid, pid)
+	if cachedResp := inventory.Store.GetString(cacheKey + ".data"); cachedResp != "" {
+		ts := inventory.Store.GetTime(cacheKey + ".ts")
+		if time.Since(ts) < time.Hour*24 {
+			// Use cached response
+			if cachedResp == "ErrNotFound" {
+				return nil, ErrNotFound
```
umbynos, line 63: is it correct to return this?

cmaglie: Yes, the web API is ill-defined, it returns a 404 (HTTP Not Found) if the vid/pid is not in the database.
It would have been better a 200 (HTTP OK) with an empty response, but I think it's late to change it now.

commands/board/list.go
cmaglie: Yes I guess? the `httpclient.Do` returns an error so why not check it?

BTW the error check was there also before I just moved from the bottom of a very big `if {...}` after inverting the condition from `err == nil` to `err != nil`.  Removing the whitespaces from the diff may help to see it.

commands/board/list.go
cmaglie: or maybe we can mask this behavior inside `apiByVidPid`? this looks like a good idea, can make the whole thing a bit nicer, let me try it, I'll add a new commit to this PR

umbynos (approved):

PR #2026, opened 2023-01-04
chore: update to golang 1.19

Bikappa:

      before creating one)
- [x] The PR follows

It updates the golang language version used in the source files. The version is bumped from go 1.17 to go 1.19.
The change will provide the latest security patches and bug fixes in the standard library and it will allow developers to write code with modern syntax.
In this regard the main language feature brought in are [generics](https://gobyexample.com/generics), that enable writing of simpler and more concise interfaces in addition to the possibility to use such interfaces from third party modules.
See golang [release notes](https://go.dev/doc/devel/release) for more details.

It also update `go.sum` files after running `go mod tidy`

Not on the interface behaviours, but it's worth mentioning that go 1.19 is now required for compilation.


i18n/convert.go
```diff
@@ -26,7 +26,7 @@ var javaFormatPlaceholderRegexp = regexp.MustCompile(`{(\d)}`)
 // FromJavaToGoSyntax convert a translation string made for Java to a one suitable for golang (printf-style).
 // The conversion transforms java placeholders like "{0}","{1}","{2}",etc... with the equivalent for golang
 // "%[1]v","%[2]v","%[3]v",etc...
-// The double single-quote "''" is translated into a single single-quote "'".
+// A pair of adjacent single-quote characters "'" (U+0027) is replaced with one only.
```
Bikappa, line 29: required rewriting for the formatter to not replace "''" with "”" and making checks angry.
See https://tip.golang.org/doc/comment#paragraphs

.github/workflows/i18n-weekly-pull.yaml
```diff
@@ -1,5 +1,10 @@
 name: i18n-weekly-pull
 
+env:
+  # See: https://github.com/actions/setup-go/tree/main#supported-version-syntax
+  GO_VERSION: ">=1.18"
```
cmaglie, line 5: Why 1.18 and not 1.19?
Should we specify something like `^1.19.4` to get the latest 1.19?

DistTasks.yml
```diff
@@ -19,7 +19,7 @@ version: "3"
 
 vars:
   CONTAINER: "docker.elastic.co/beats-dev/golang-crossbuild"
-  GO_VERSION: "1.17.8"
+  GO_VERSION: "1.19"
```
cmaglie, line 22: Currently, the latest go release is `1.19.4`, I think the containers require an exact version.

.github/workflows/i18n-weekly-pull.yaml
Bikappa: Good catch on the `1.18` 👍 
For the specific version, I first went that way but then I saw that the `setup-go` action [already picks the latest minor release](https://github.com/arduino/arduino-cli/actions/runs/3846152753/jobs/6551134726#step:4:15). So I steered towards keeping the same pattern as before

DistTasks.yml
Bikappa: crossbuild images don't have a `1.19.4` tag yet, but updated to `1.19.3`

cmaglie (approved):

PR #2034, opened 2023-01-17
[skip-changelog] Add query parameters to urls generated by lib commands

MatteoPologruto:

      before creating one)
- [x] The PR follows

Code enhancement

## What is the new behavior?
Three different query parameters have been added to the urls generated by `lib` commands like `lib install` and `lib upgrade`. The query parameters are:
- `install`: which is selected when a library is installed through `lib install`;
- `upgrade`: which is selected when a library is upgraded through `lib upgrade`;
- `depends`: which is selected when a library dependency is installed.

No

### DIFF
--- a/arduino/httpclient/httpclient.go
+++ b/arduino/httpclient/httpclient.go
@@ -33,6 +33,7 @@ var tr = i18n.Tr
 
 // DownloadFile downloads a file from a URL into the specified path. An optional config and options may be passed (or nil to use the defaults).
 // A DownloadProgressCB callback function must be passed to monitor download progress.
+// If a not empty queryParameter is passed, it is appended to the URL for analysis purposes.
 func DownloadFile(path *paths.Path, URL string, queryParameter string, label string, downloadCB rpc.DownloadProgressCB, config *downloader.Config, options ...downloader.DownloadOptions) (returnedError error) {
 	if queryParameter != "" {
 		URL = URL + "?query=" + queryParameter
--- a/arduino/httpclient/httpclient.go
+++ b/arduino/httpclient/httpclient.go
@@ -25,6 +25,7 @@ import (
 	"github.com/arduino/arduino-cli/i18n"
 	rpc "github.com/arduino/arduino-cli/rpc/cc/arduino/cli/commands/v1"
 	"github.com/arduino/go-paths-helper"
+	"github.com/sirupsen/logrus"
 	"go.bug.st/downloader/v2"
 )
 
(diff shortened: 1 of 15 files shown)


arduino/httpclient/httpclient.go
```diff
@@ -25,14 +25,19 @@ import (
...
 	rpc "github.com/arduino/arduino-cli/rpc/cc/arduino/cli/commands/v1"
 	"github.com/arduino/go-paths-helper"
+	"github.com/sirupsen/logrus"
 	"go.bug.st/downloader/v2"
 )
 
 var tr = i18n.Tr
 
 // DownloadFile downloads a file from a URL into the specified path. An optional config and options may be passed (or nil to use the defaults).
 // A DownloadProgressCB callback function must be passed to monitor download progress.
```
umbynos, line 35: Could you please change this comment and include info regarding the new parameter?

arduino/resources/download.go
```diff
@@ -27,7 +27,7 @@ import (
 
 // Download performs a download loop using the provided downloader.Config.
 // Messages are passed back to the DownloadProgressCB using label as text for the File field.
```
umbynos, line 29: same as above

commands/lib/download.go
```diff
@@ -31,7 +31,7 @@ import (
 var tr = i18n.Tr
 
 // LibraryDownload FIXMEDOC
```
umbynos, line 33: This could be the chance to fix this

commands/lib/install.go
```diff
@@ -31,7 +31,7 @@ import (
 )
 
 // LibraryInstall FIXMEDOC
```
umbynos, line 33: same as above

commands/lib/install.go
```diff
@@ -96,11 +96,20 @@ func LibraryInstall(ctx context.Context, req *rpc.LibraryInstallRequest, downloa
 	}
 
 	for libRelease, installTask := range libReleasesToInstall {
-		if err := downloadLibrary(lm, libRelease, downloadCB, taskCB); err != nil {
-			return err
-		}
-		if err := installLibrary(lm, libRelease, installTask, taskCB); err != nil {
-			return err
+		if libRelease.GetName() == req.Name {
```
umbynos, line 99: why is this requires? It's not so clear

internal/cli/lib/download.go
```diff
@@ -60,7 +60,7 @@ func runDownloadCommand(cmd *cobra.Command, args []string) {
 			Name:     library.Name,
 			Version:  library.Version,
 		}
-		_, err := lib.LibraryDownload(context.Background(), libraryDownloadRequest, feedback.ProgressBar())
+		_, err := lib.LibraryDownload(context.Background(), libraryDownloadRequest, feedback.ProgressBar(), "")
```
umbynos, line 63: I don't know if this has to be left empty

umbynos (changes requested):
left some suggestions

umbynos (approved):
LGTM :checkered_flag:

cmaglie:
I see a couple of problems with this PR:
* it changes the signature of the function `commands.LibraryDownload`, this is a breaking change that is not documented
* all public functions in package `commands` are made to strictly follow the gRPC API. If we need to add a parameter in any of the public functions of `command.*` we must first change the gRPC API and rebuild it otherwise, we allow the CLI to bypass the gRPC API. In fact, this implementation works only in the CLI, there is no way to access this feature from the gRPC API (so it will not work in the Arduino IDE or any other gRPC client).

PR #2038, opened 2023-01-19
refactoring: Fixing legacy pathnames

cmaglie:

      before creating one)
- [x] The PR follows

Some changes and cleanup on the `legacy` wrt build paths.
There is a lot of redundancy in the `legacy` builder, this PR aims to remove all code duplication.

No changes in behavior, this PR should be equivalent code change.

No breaking changes


commands/compile/compile.go
```diff
@@ -155,11 +153,16 @@ func Compile(ctx context.Context, req *rpc.CompileRequest, outStream, errStream
...
 		if err != nil {
 			return nil, &arduino.PermissionDeniedError{Message: tr("Cannot create build cache directory"), Cause: err}
 		}
+		if err := buildCachePath.MkdirAll(); err != nil {
+			return nil, &arduino.PermissionDeniedError{Message: tr("Cannot create build cache directory"), Cause: err}
+		}
+		builderCtx.CoreBuildCachePath = buildCachePath.Join("core")
+	} else {
+		builderCtx.CoreBuildCachePath = paths.TempDir().Join("arduino", "core-cache")
 	}
```
Bikappa, line 166: Nit: I'd swap the if with the else. I find it easier to read if the short cases come first because I don't need to remember what the condition was when encountering an else

arduino/sketch/sketch.go
```diff
@@ -293,14 +291,15 @@ func CheckForPdeFiles(sketch *paths.Path) []*paths.Path {
 	return files
 }
 
-// GenBuildPath generates a suitable name for the build folder.
-// The sketchPath, if not nil, is also used to furhter differentiate build paths.
-func GenBuildPath(sketchPath *paths.Path) *paths.Path {
```
Bikappa, line 298: wasn't a different hash created in the case a sketchPath was given? We don't need it anymore?

cmaglie: No, actually the hash depends only on the given `sketchPath` (a.k.a. `sketch.FullPath`).

I removed the static function `GenBuildPath` and put it as a method of `Sketch` -> `sketch.DefaultBuildPath(...)` which is much more expressive.

Also, I've made the `sketch.Hash()` function that gives back only the "hash" piece so we can use it in #2033 as a key for a directory cache.

Bikappa (approved):

PR #2082, opened 2023-02-21
Refactored build-properties creation (and removed some legacy code)

### LINKED ISSUE #1012: Uploader prefs miss  {build.core.path}, {build.system.path} and {build.variant.path}
The keys `{build.core.path}`, `{build.system.path}` and `{build.variant.path}` are not evaluated for platform.txt tools recipes (`tools.xy.upload.pattern`).

Then the tools recipes must use `runtime.platform.path`. But then there are problems to use the platform as referenced core, because `runtime.platform.path` is the referencing board platform path.

For example in the SAMD package the `upload.pattern` contains `{runtime.platform.path}/variants/{build.variant}/{build.openocdscript}` instead of `{build.variant.path}/{build.openocdscript}`. A board in referencing board platform can use in board.txt an existing variant of the referenced core platform in `xy.build.variant` for compilation, but the upload tool will not find the script for the variant.

I try to make the esp8266 package to be used as as referenced core. Now it uses `{runtime.platform.path}/tools` all over platform.txt and `{runtime.platform.path}` will evaluate to referencing board platform, not to  referenced core platform. They could use `{build.core.path}/../../tools` or rename the folder to `system` and use `{build.system.path}` variable. It works for building, but it doesn't evaluate in `uploader.pattern`.

I debugged the Arduino IDE 1 and found out that compiler recipes use prefs evaluated by arduino-builder, while Uploaders use the preferences map of the IDE. So it is not a simple fix. I think, the best way for fixing this would be to pass the prefs evaluated by `Compiler` over `SketchController` from build() to upload(). Same in class `processing.app.Base`.

cmaglie:

      before creating one)
- [X] The PR follows

It moves most of the build-properties creation out of the `legacy` package. This has some advantages:
* It has centralized the place where these build properties are made, so we removed some duplicated logic
* This is particularly true for the `runtime.tools.*` properties generation
* It fixes #1012: some of the `build.*` properties are now available on upload actions

Build properties are missing for upload recipes.

## What is the new behavior?

Build properties are now available in upload recipes.

No breaking changes.


arduino/cores/packagemanager/package_manager.go
```diff
@@ -288,36 +293,112 @@ func (pme *Explorer) ResolveFQBN(fqbn *cores.FQBN) (
...
+		referredPlatformRelease = pme.GetInstalledPlatformRelease(referredPlatform)
+		if referredPlatformRelease == nil {
+			return targetPackage, platformRelease, board, nil, nil,
+				fmt.Errorf(tr("missing platform release %[1]s:%[2]s referenced by board %[3]s"), referredPackageName, fqbn.PlatformArch, fqbn)
 		}
-		buildPlatform := buildPackage.Platforms[fqbn.PlatformArch]
-		if buildPlatform == nil {
-			return targetPackage, platformRelease, board, buildProperties, nil,
-				fmt.Errorf(tr("missing platform %[1]s:%[2]s referenced by board %[3]s"), referredPackage, fqbn.PlatformArch, fqbn)
+	}
```
Bikappa, line 340: Can we make a function of these lines? The block is very wordy

cmaglie: ok it should be much cleaner now!

Bikappa (approved):
lgtm

PR #2083, opened 2023-02-22
[skip-changelog] Exclude libraries with a missing `.h` file from `lib list`

MatteoPologruto:

      before creating one)
- [x] The PR follows

Code imperfection fix

Empty folders under `<directories.user>/libraries` show up as `is_legacy: true` libraries.

## What is the new behavior?
Folders that do not contain a `library.properties` or a `.h` file are ignored when `lib list` is executed.

No

### DIFF
--- a/arduino/libraries/loader.go
+++ b/arduino/libraries/loader.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/arduino/arduino-cli/arduino/globals"
 	"github.com/arduino/arduino-cli/arduino/sketch"
 	"github.com/arduino/go-paths-helper"
 	properties "github.com/arduino/go-properties-orderedmap"
@@ -126,6 +127,11 @@ func makeNewLibrary(libraryDir *paths.Path, location LibraryLocation) (*Library,
 }
 
 func makeLegacyLibrary(path *paths.Path, location LibraryLocation) (*Library, error) {
+	if foundHeader, err := containsHeaderFile(path); err != nil {
+		return nil, err
+	} else if !foundHeader {
+		return nil, errors.Errorf(tr("invalid library: no header files found"))
+	}
 	library := &Library{
 		InstallDir:    path.Canonical(),
 		Location:      location,
@@ -185,4 +191,20 @@ func addExamplesToPathList(examplesPath *paths.Path, list *paths.PathList) error
 		}
 	}
 	return nil
+}
+
+// containsHeaderFile returns true if the directory contains a ".h" file.
+// Returns false otherwise
+func containsHeaderFile(d *paths.Path) (bool, error) {
+	dirContent, err := d.ReadDir()
+	if err != nil {
+		return false, fmt.Errorf(tr("reading directory %[1]s content: %[2]w", d, err))
+	}
+	dirContent.FilterOutDirs()
+	headerExtensions := []string{}
+	for k := range globals.HeaderFilesValidExtensions {
+		headerExtensions = append(headerExtensions, k)
+	}
+	dirContent.FilterSuffix(headerExtensions...)
+	return len(dirContent) > 0, nil
 }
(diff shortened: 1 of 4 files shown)


internal/integrationtest/lib/lib_test.go
```diff
@@ -1535,3 +1535,20 @@ func TestLibQueryParameters(t *testing.T) {
...
+
+	// Create a library that does not have neither a library.properties nor a ".h" fil
+	emptyLib := cli.SketchbookDir().Join("libraries", "empty")
+	require.NoError(t, emptyLib.MkdirAll())
+
+	// Check the output of lib list is empty
+	stdout, _, err := cli.Run("lib", "list", "--format", "json")
+	require.NoError(t, err)
+	requirejson.Empty(t, stdout)
+}
```
umbynos, line 1554: I would enhance this test by creating a lib with a `.h` file inside it too

cmaglie (changes requested):

commands/lib/list.go
```diff
@@ -135,6 +135,9 @@ func listLibraries(lm *librariesmanager.LibrariesManager, updatable bool, all bo
 			if updatable && available == nil {
 				continue
 			}
+			if !containsDotH(lib) {
+				continue
+			}
```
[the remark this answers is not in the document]
umbynos: :man_facepalming: I did not thought about this but you are completely right. In this case adding a test for the other libs command could help

arduino/libraries/loader.go
```diff
@@ -186,3 +192,23 @@ func addExamplesToPathList(examplesPath *paths.Path, list *paths.PathList) error
...
 	return nil
 }
+
+// containsHeaderFile returns true if the directory contains a ".h" file.
+// Returns false otherwise
+func containsHeaderFile(d *paths.Path) (bool, error) {
+	if d.NotExist() {
+		// A directory that doesn't exist can't obviously contain any header file
+		return false, nil
+	}
```
cmaglie, line 202: I'd remove this check because:
* we should never reach this point with `d` nonexistent
* if it will ever happen, it's better to fail with an error on `ReadDir` later

arduino/libraries/loader.go
```diff
@@ -186,3 +192,23 @@ func addExamplesToPathList(examplesPath *paths.Path, list *paths.PathList) error
...
+	dirContent, err := d.ReadDir()
+	if err != nil {
+		return false, fmt.Errorf(tr("reading directory %s content: %w", d, err))
```
cmaglie, line 205:
```suggestion
		return false, fmt.Errorf(tr("reading directory %[1]s content: %[2]w", d, err))
```

internal/integrationtest/lib/lib_test.go
```diff
@@ -1587,3 +1587,40 @@ func TestLibBundlesWhenLibWithTheSameNameIsInstalledGlobally(t *testing.T) {
 		requirejson.Parse(t, stdout).Query(`.[].library.examples[1]`).MustContain(`"OTALeds"`)
 	}
 }
+
+func TestLibListDoesNotIncludeEmptyLibraries(t *testing.T) {
```
cmaglie, line 1591: In addition to the integration test could you make a unit test to check the `Load` function behavior here: https://github.com/arduino/arduino-cli/blob/master/arduino/libraries/libraries_test.go
After adding the unit-test you should be able to immediately see code coverage in the IDE.

arduino/libraries/loader.go
```diff
@@ -126,6 +127,11 @@ func makeNewLibrary(libraryDir *paths.Path, location LibraryLocation) (*Library,
 }
 
 func makeLegacyLibrary(path *paths.Path, location LibraryLocation) (*Library, error) {
+	if foundHeader, err := containsHeaderFile(path); err != nil {
+		return nil, err
+	} else if !foundHeader {
+		return nil, errors.Errorf(tr("library not valid"))
```
cmaglie, line 133: Let's make it a bit more verbose on why it has been rejected.
```suggestion
		return nil, errors.Errorf(tr("invalid library: no header files found"))
```

cmaglie (changes requested):

umbynos (approved):

PR #2085, opened 2023-02-23
Fix/remove tree config lookup

### LINKED ISSUE #758: Loading arduino-cli.yaml from current directory might be unexpected or unsafe
First off: even though I consider this a security problem, I do not think this is very easy to exploit (unnoticed). I've reported this issue tot the Arduino security staff privately first, and we agreed that this would be acceptable for public discussion. Hence this issue.

I noticed (after reading https://github.com/arduino/arduino-cli/issues/699#issuecomment-633567379) that `arduino-cli` allows loading its `arduino-cli.yaml` config file from the current directory, or any of the parent directories (e.g. see [here](https://github.com/arduino/arduino-cli/blob/8220b32d158723143ff942eb57d21534dfe91d78/cli/cli.go#L154-L156) and [here](https://github.com/arduino/arduino-cli/blob/8220b32d158723143ff942eb57d21534dfe91d78/cli/cli.go#L126-L144)).

However, I'm a bit worried about the (security) implications of this. It seems that this would make it very easy to mess up arduino-cli when a `arduino-cli.yaml` file is present in an unexpected location.

From a security standpoint, this means that `arduino-cli` can *never* be ran from a untrusted directory, which I think users will not expect.

For example, consider a sketch downloaded from elsewhere. Normally, compiling an untrusted sketch is not problematic, since the files contained in the sketch are only passed to the compiler, not actually executed on the compiling machine. Uploading a sketch with untrusted code to an Arduino can of course be problematic, but Arduinos are usually sufficient sandboxed, and a user can expect that untrusted code will be ran in this case.

Now, consider that this untrusted sketch also contains an `arduino-cli.yaml` file, that sets up the `user` directory to a directory *inside* the sketch. If it then also contains a `hardware` directory with a custom core, then running `arduino-cli compile` will execute arbitrary commands from the recipes in the untrusted `platform.txt` file. These could then do anything, including removing files, installing a backdoor or stealing data.

As for fixing this, here's some thoughts:

 1. You can think of an approach where not all configuration directives will be accepted in these potentially untrusted "project" config files (i.e. divide all config directives into two groups, one needing a trusted config file and the other allowed from any config file).  However, this is often fragile, when new options are added or refactored, it is easy to forgot to realize that an option can cause a security problem.

 2. Hence, it might be better to consider an approach where a project config file must be explicitly allowed (e.g. when an `arduino-cli.yaml` exists in the current directory, `arduino-cli` should ask the user whether it is trusted, maybe with a short explanation of the risks involved). Then, for each file (maybe along with a checksum), the answer (allow or ignore) should be stored. This is an approach also used by tools like `direnv` or `githooks`.

 3. Or maybe disable these project files by default, and enable them in the global `arduino-cli.yaml` config file.

Note that all of the above suggestions require that the e.g.  `~/.arduino15/arduino-cli.yaml` global file should *always* be loaded (either to get values for the trusted-only directives, to store the trusted project config files, or to store the "allow project config" setting). That also opens up the door to loading the global file for defaults, and letting a project config file only override some values, as needed, which reduces the chances of a project config file from accidentally changing settings it does not really need to change.

The essence of this issue has been considered before, which is (I think) one of the reason why per-sketch compilation options were never implemented in the Java IDE (e.g. see https://github.com/arduino/Arduino/issues/421#issuecomment-326992418).

Bikappa:

      before creating one)
- [ ] The PR follows

## Other information

Fixes #758

### DIFF
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -136,9 +136,9 @@ func GetDefaultBuiltinLibrariesDir() string {
 	return filepath.Join(getDefaultArduinoDataDir(), "libraries")
 }
 
-// FindConfigFileInArgsOrWorkingDirectory returns the config file path using the
+// FindConfigFileInArgs returns the config file path using the
 // argument '--config-file' (if specified) or looking in the current working dir
-func FindConfigFileInArgsOrWorkingDirectory(args []string) string {
+func FindConfigFileInArgs(args []string) string {
 	// Look for '--config-file' argument
 	for i, arg := range args {
 		if arg == "--config-file" {
@@ -147,23 +147,5 @@ func FindConfigFileInArgsOrWorkingDirectory(args []string) string {
 			}
 		}
 	}
-
-	// Look into current working directory
-	if cwd, err := paths.Getwd(); err != nil {
-		return ""
-	} else if configFile := searchConfigTree(cwd); configFile != nil {
-		return configFile.Join("arduino-cli.yaml").String()
-	}
 	return ""
-}
-
-func searchConfigTree(cwd *paths.Path) *paths.Path {
-	// go back up to root and search for the config file
-	for _, path := range cwd.Parents() {
-		if path.Join("arduino-cli.yaml").Exist() {
-			return path
-		}
-	}
-
-	return nil
 }
--- a/configuration/configuration_test.go
+++ b/configuration/configuration_test.go
@@ -21,7 +21,6 @@ import (
 	"path/filepath"
 	"testing"
 
-	paths "github.com/arduino/go-paths-helper"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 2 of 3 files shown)


matthijskooijman:
Shouldn't this be marked as breaking compatibility? And set to close #758?

Bikappa:
> Shouldn't this be marked as breaking compatibility? And set to close #758?

@matthijskooijman yes, that's correct. It will be marked as such once it leaves the draft mode, at the latest.

internal/integrationtest/compile_1/compile_test.go
```diff
@@ -526,7 +526,7 @@ func compileWithExportBinariesConfig(t *testing.T, env *integrationtest.Environm
 	defer cli.WorkingDir().Join("arduino-cli.yaml").Remove()
 
 	// Test if arduino-cli config file written in the previous run has the `always_export_binaries` flag set.
-	stdout, _, err := cli.Run("config", "dump", "--format", "json")
+	stdout, _, err := cli.Run("config", "dump", "--format", "json", "--config-file", "arduino-cli.yaml")
```
cmaglie, line 529: All these integration tests are not intended to test the `--config-file` option, it's just boilerplate code to make the test work.

I'd better change `cli.Run(..)` method to transparently add the `--config-file arduino-cli.yaml` flag to the command line (unless specified otherwise) so we don't need to repeat it repeatedly.

cmaglie (changes requested):

cmaglie (approved):

PR #2103, opened 2023-03-09
Extend code-coverage to integration test

### LINKED ISSUE #1829: Proposal to investigate: Increase test coverage using integration tests
### Describe the request

Currently, the code coverage is calculated only on unit-test because go-lang offers this feature via `go test` command. For example, in our [`go:test` Taskfile script](https://github.com/arduino/arduino-cli/blob/869c971189d94f6786ebe83f2f5f085c5a71e100/Taskfile.yml#L99-L106) we usa a command line flag that can be summarized as:

```
$ go test -coverpkg=./... -covermode=atomic -coverprofile=coverage_unit.txt ...
```
These flags produce a `coverage_unit.txt` file that is later sent to the codecov.

In the integration tests, instead, we run directly a build of the `arduino-cli`, using the `executils` library. in this case, just adding the `-cover*` flags in the `go test` commands actually would not add any coverage (except maybe for the `executils` library).

BTW a way to workaround this limitation exists, even if a bit convoluted, here's how:

1. the `go test` command offers a `-c` flag that "compiles" the test instead of running it:
   ```
   -c
       Compile the test binary to pkg.test but do not run it
       (where pkg is the last element of the package's import path).
       The file name can be changed with the -o flag.
   ```
   This means that we can create and executable that when run performs the tests exactly as `go test` will do.
2. We can make a special "unit-test" that just runs the CLI with the given arguments, something along the line of:
   ```go
   func RunArduinoCLITest(t *testing.T) {
      // ...mangle os.Args to fix some paths or arguments...
      cli.Main(os.Args)
   }
   ```
3. We can create with `go test ./internal/cli-runner/ --run=RunArduinoCLITest -coverpkg=./... -covermode=atomic -coverprofile=coverage_unit.txt  -c -o arduino-cli-withcoverage` an executable called `arduino-cli-withcoverage` that will run the `arduino-cli` as it was in a test but with the parameters we pass to it.
4. We can use this `arduino-cli-withcoverage` to run integration tests instead of `arduino-cli` and get the coverage from the integration tests!

Now this is very theoretical, I can already see some obstacles:

1. We need to refactor the arduino-cli so it has a callable `Main` method (easy)
2. We need to mangle os.Args in the `RunArduinoCLITest` to make it digestible from the arduino-cli (medium)
3. The test coverage output is written if the process terminates "cleanly": this means that we should intercept panics (medium) or `os.Exit` calls (hard)
4. From the integration-test we may need to kill the arduino-cli-withcoverage process, but killing it would not produce coverage test, so we probably need to set up a communication channel to exit the process cleanly (hard)
5. Each run of `arduino-cli-withcoverage` will produce a different `coverage.txt`, all the `coverage.txt` should be collected and merged together (medium)
6. Something else that I'm surely not considering here (hard)

Anyway, given the difficulties above, IMHO it's still something worth considering if possible.

No test coverage is reported from integration-tests.

### Arduino CLI version

N/A

### Operating system

N/A

### Operating system version

N/A

### Additional context

_No response_

### Issue checklist

- [X] I verified the feature was still missing when using the [nightly build](https://arduino.github.io/arduino-cli/dev/installation/#nightly-builds)

cmaglie:

      before creating one)
- [X] The PR follows

It's a tentative implementation of the tooling described here: https://go.dev/blog/integration-test-coverage
It requires to upgrade golang version to 1.20

Code-coverage is calculated only for unit tests.

## What is the new behavior?

Code-coverage is calculated on unit test and integration tests.

No

## Other information

Fix #1829


Taskfile.yml
```diff
@@ -110,10 +110,14 @@ tasks:
...
     deps:
       - task: go:build
+        vars:
+          EXTRA_FLAGS: "-covermode=atomic"
     dir: '{{default "./" .GO_MODULE_PATH}}'
     cmds:
       - |
-        go test \
+        rm -fr coverage_data
+        mkdir coverage_data
```
umbynos, line 119: is this required?

Taskfile.yml
```diff
@@ -110,10 +110,14 @@ tasks:
...
+        INTEGRATION_GOCOVERDIR={{ .ROOT_DIR }}/coverage_data go test \
```
umbynos, line 120: Shouldn't be possible to avoid this and use directly `GOCOVERDIR` env var instead? Without modifying the `arduino-cli.go`. See `Running a coverage-instrumented binary` section [here](Running a coverage-instrumented binary)

cmaglie: No, I tried it, as I've explained in the PR comment:
>  1. To produce the coverage output, the instrumented executable must be run with a GOCOVERDIR environment variable set to an existing directory path where the coverage data will be written. To accomplish this, I've added to the integration test the support for another env var INTEGRATION_GOCOVERDIR, the content of this variable will be passed to the arduino-cli run as GOCOVERDIR. **We can not use GOCOVERDIR directly because it is overwritten by the `go test ...` command.**

Taskfile.yml
cmaglie: Yes, the directory must exist.
Also every time you run the executable a new file is added to the directory, so if you restart the test you must remove the old data otherwise you get the union of the old coverage with the new one.

umbynos (approved):

PR #2119, opened 2023-03-23
Automatically download indexes, if missing, in gRPC `Init` call

### LINKED ISSUE #1529: `GetPlatform` called via gRPC does not force to update the indices
### Current behavior

Call `GetPlatform` with `all=true` on a clean system does not force to update the indices.

### Expected behavior

It forces the update (exactly as it's done if calling via command-line `core list --all`

### Environment

- CLI version (output of `arduino-cli version`): 0.19.2 commit [6a6f50ae](https://github.com/arduino/arduino-cli/commit/6a6f50ae)
- OS and platform: Alpine linux

### Additional context

Looking into the source code i can't see any difference except missing `instance.CreateAndInit()` [call](https://github.com/arduino/arduino-cli/blob/e0d983a8840f4d2c1024a50a3909f21796458710/cli/core/list.go#L52).

In my case i do call create + init requests before requesting the subscriptions.

PS. Using custom `data` path in config file (not sure if it makes a difference)

cmaglie:

      before creating one)
- [X] The PR follows
- [X] `UPGRADING.md` has been updated with a migration guide (for breaking changes)

This PR changes the behavior of the `Init` function of the gRPC API.

Previously if the main indexes `package_index.json` or `library_index.json` were missing (for example after installing on a clean system) the gRPC client had to call `UpdateIndex` or `UpdateLibrariesIndex` to download the indexes and re-`Init` the gRPC instance.

## What is the new behavior?

The indexes download is performed automatically.

It changes some golang API calls as described in the `UPGRADING.md`.
The gRPC is backward compatible, but the old clients may not be optimized and may force the download of the index twice.

## Other information

Fix #1529

### DIFF
--- a/arduino/resources/index.go
+++ b/arduino/resources/index.go
@@ -37,6 +37,15 @@ type IndexResource struct {
 	SignatureURL *url.URL
 }
 
+// IndexFileName returns the index file name as it is saved in data dir (package_xxx_index.json).
+func (res *IndexResource) IndexFileName() string {
+	filename := path.Base(res.URL.Path) // == package_index.json[.gz] || packacge_index.tar.bz2
+	if i := strings.Index(filename, "."); i != -1 {
+		filename = filename[:i]
+	}
+	return filename + ".json"
+}
+
 // Download will download the index and possibly check the signature using the Arduino's public key.
 // If the file is in .gz format it will be unpacked first.
 func (res *IndexResource) Download(destDir *paths.Path, downloadCB rpc.DownloadProgressCB) error {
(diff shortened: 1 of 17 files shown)


kittaakos:
This is looking very promising. 😍 The sister PR in the IDE2 repo is here: https://github.com/arduino/arduino-ide/pull/2023.


A few remarks:
 - IDE2 would like to recover the 3rd party package indexes on startup. Is it possible? It does not work currently: https://github.com/arduino/arduino-ide/issues/2021.
 - IDE2 does a periodic (every 4 hours) index update. To correctly calculate the four hours timeframe, IDE2 must know if the index has been recovered during the app startup. With this PR, it's impossible, as IDE2 does not know if the index has been updated.
 - _TBD (I am going to update this message with additional items...)_

cmaglie:
I'm still not fully convinced this is the right thing to do, BTW I've implemented this one:

> IDE2 would like to recover the 3rd party package indexes on startup. Is it possible? It does not work currently: https://github.com/arduino/arduino-ide/issues/2021.

so we can try it.

The visible change to the user is that now, as soon as you add a 3rd party package_index URL to the config, the CLI will automatically download the index, if not already present, at the first use (you don't need to do the **first** `core update-index` anymore).
BTW you still need to do `update-index` if you want to update it.

> IDE2 does a periodic (every 4 hours) index update. To correctly calculate the four hours timeframe, IDE2 must know if the index has been recovered during the app startup. With this PR, it's impossible, as IDE2 does not know if the index has been updated.

The extra auto-update should not happen (only at the first startup maybe? or if the user explicitly deletes an index from the data dir... how often this happens?).
BTW In the worst-case scenario, the only visible effect would be the two index updates in a row, not a big deal IMHO.

kittaakos:
> so we can try it.

Great. I will update the IDE2 branch soon.

> how often this happens?).
> the only visible effect would be the two index updates in a row, not a big deal IMHO.

I completely agree. Let's not overcomplicate the index update strategies.

alessio-perugini (approved):

PR #2129, opened 2023-03-27
Added codecov token to solve code-coverage upload issues

cmaglie:

      before creating one)
- [X] The PR follows

Added codecov token to solve code-coverage upload issues

## Other information

See https://github.com/codecov/codecov-action/issues/557

### DIFF
--- a/.github/workflows/test-go-task.yml
+++ b/.github/workflows/test-go-task.yml
@@ -197,6 +197,7 @@ jobs:
       - name: Send unit tests coverage to Codecov
         uses: codecov/codecov-action@v3
         with:
+          token: ${{ secrets.CODECOV_TOKEN }}
           files: ./coverage.txt
           flags: unit
           fail_ci_if_error: ${{ github.repository == 'arduino/arduino-cli' }}


umbynos (approved):

per1234 (commented):
It should be noted that PRs from forks do not have access to repository secrets, so this fix only applies to PRs from branches of the `arduino/arduino-cli` repository. The same intermittent spurious workflow run failures will continue to occur for PRs from forks (such as this PR).

The alternative solution would be to add the token in plaintext directly in the workflow. The security implications of that approach are described here:

https://community.codecov.com/t/upload-issues-unable-to-locate-build-via-github-actions-api/3954

> Public repositories that rely on PRs via forks will find that they cannot effectively use Codecov if the token is stored as a GitHub secret. The scope of the Codecov token is only to confirm that the coverage uploaded comes from a specific repository, not to pull down source code or make any code changes.
>
> For this reason, we recommend that teams with public repositories that rely on PRs via forks consider the security ramifications of making the Codecov token available as opposed to being in a secret.
>
> A malicious actor would be able to upload incorrect or misleading coverage reports to a specific repository if they have access to your upload token, but would not be able to pull down source code or make any code changes.

.github/workflows/test-go-task.yml
```diff
@@ -197,6 +197,7 @@ jobs:
       - name: Send unit tests coverage to Codecov
         uses: codecov/codecov-action@v3
         with:
+          token: db21daa9-939e-4742-be68-a9db33b8bdcb
```
per1234, line 200:
```suggestion
          token: ${{ vars.CODECOV_TOKEN }}
```

I gave it more thought and I don't think it is a good idea to hardcode into the workflow in this way after all.

I'm not too concerned about malicious use of the token because an attacker can't do anything of value with it. However, I think there is the potential for inadvertent harm by legitimate contributors and other community members. With the token hardcoded into the workflow, the coverage data from runs in forks will be uploaded to Arduino's codecov account. That isn't much of a problem if the run is in a feature branch, but it is fairly common for contributors to stage PRs in the default branch, and also any development work in forks would also likely be done in the default branch. This will cause irrelevant data to be added to Arduino's Codecov account.

For example, here is some data that was uploaded from a run triggered by a change I made in the `master` branch in my fork:

image

There is a recently added "repository variables" feature that allows variables to be defined in the repository settings. These are not secrets, so they are available to workflow runs triggered by a PR from a fork:

https://docs.github.com/en/actions/learn-github-actions/variables#defining-configuration-variables-for-multiple-workflows

The variables are not transferred when someone forks a repository, so the token will only be defined when the workflow runs in this repository (the forker could always define their own variable with the token from their Codecov account if they want their own coverage data).

per1234 (changes requested):

.github/workflows/test-go-task.yml
cmaglie: > With the token hardcoded into the workflow, the coverage data from runs in forks will be uploaded to Arduino's codecov account.

That was a case I'd liked to try before merging, but you beat me, thanks @per1234 :-)

I'll regenerate the token and convert it to a repository variable.

per1234: >I'll regenerate the token and convert it to a repository variable.

To accompany my review suggestion, I already added a variable for the existing token value:

image


But you're welcome to adjust that however you like, on this page:

https://github.com/arduino/arduino-cli/settings/variables/actions

Note that GitHub wrote some incorrect information on that page:

>Anyone with collaborator access to this repository can use these secrets and variables for actions. They are not passed to workflows that are triggered by a pull request from a fork.

This is correct for secrets, but incorrect for variables. Variables are passed to workflows triggered by PRs from forks.

cmaglie: I've updated the token since the old one was published.

.github/workflows/test-go-task.yml
```diff
@@ -197,6 +197,7 @@ jobs:
       - name: Send unit tests coverage to Codecov
         uses: codecov/codecov-action@v3
         with:
+          token: ${{ secrets.CODECOV_TOKEN }}
```
per1234, line 200:
```suggestion
          token: ${{ vars.CODECOV_TOKEN }}
```

Reference:
https://docs.github.com/en/actions/learn-github-actions/variables#using-the-vars-context-to-access-configuration-variable-values

per1234 (changes requested):

.github/workflows/test-go-task.yml
```diff
@@ -197,6 +197,7 @@ jobs:
       - name: Send unit tests coverage to Codecov
         uses: codecov/codecov-action@v3
         with:
+          token: ${{ vars.CODECOV_TOKEN }}
```
per1234, line 200:
```suggestion

      # A token is used to avoid intermittent spurious job failures caused by rate limiting.
      - name: Set up Codecov upload token
        run: |
          if [[ "${{ github.repository }}" == "arduino/arduino-ide" ]]; then
            # In order to avoid uploads of data from forks, only use the token for runs in the arduino/arduino-ide repo.
            # Token is intentionally exposed.
            # See: https://community.codecov.com/t/upload-issues-unable-to-locate-build-via-github-actions-api/3954
            CODECOV_TOKEN="18b9885b-cbf2-4166-bfbb-39617323daf0"
          else
            # codecov/codecov-action does unauthenticated upload if empty string is passed via the `token` input.
            CODECOV_TOKEN=""
          fi

          echo "CODECOV_TOKEN=$CODECOV_TOKEN" >> "$GITHUB_ENV"

      - name: Send unit tests coverage to Codecov
        uses: codecov/codecov-action@v3
        with:
          token: ${{ env.CODECOV_TOKEN }}
```

There was an unexpected result of the last workflow run:

I investigated and found I lied in my previous comment:

> Note that GitHub wrote some incorrect information on that page:
>
>> Anyone with collaborator access to this repository can use these secrets and variables for actions. They are not passed to workflows that are triggered by a pull request from a fork.
>
> This is correct for secrets, but incorrect for variables. Variables are passed to workflows triggered by PRs from forks.

That was based on an experiment I ran, but I must have made a mistake in the experiment or my interpretation of the results because when I try it now I see that, indeed variables are not passed to runs triggered by PRs from forks. It doesn't make any sense that GitHub would impose that restriction because the variables are printed in the logs so anyone can see them so there is no reason to treat them the same as secrets. But that's how it works. I apologize for leading this PR down a dead end road.

---

So the only solution is to go back to hardcoding the token in the workflow, but configure the workflow to only use that token for runs in this repository.

Although I don't like that the new approach adds complexity to the workflow, the advantage is that the whole system is accessible to all contributors rather than relying on a repository variable that can only be set by an administrator.

per1234 (changes requested):

.github/workflows/test-go-task.yml
cmaglie: > `if [[ "${{ github.repository }}" == "arduino/arduino-ide" ]]; then`

I guess you intended `arduino/arduino-cli` here.

per1234: > I guess you intended arduino/arduino-cli here.

Oops, yeah copy/paste error. 🤦

per1234 (approved):
Working as intended now:

https://github.com/arduino/arduino-cli/actions/runs/4551550129/jobs/8026020631#step:8:29

```text
[2023-03-29T08:05:58.070Z] ['info'] ->  Token found by environment variables
```

Thanks Cristian!

PR #2159, opened 2023-04-27
Respect the order of menu, when computing the config options

alessio-perugini:

      before creating one)
- [x] The PR follows

This will change the output sorting when calling the boards details.

When launching the `arduino-cli boards details -b esp8266:esp8266:generic --format json` the order of the
config options doesn't follow the order of the menu declared in the first lines of `boards.txt` file.

## What is the new behavior?

We now ensure that the order of the config options are following the order of the menus fields declared at the beginning of `boards.txt`

Nope

### DIFF
--- a/arduino/cores/board.go
+++ b/arduino/cores/board.go
@@ -79,15 +79,13 @@ func (b *Board) buildConfigOptionsStructures() {
 
 	b.configOptions = properties.NewMap()
 	allConfigs := b.Properties.SubTree("menu")
+	allConfigOptions := allConfigs.FirstLevelOf()
 
 	// Used to show the config options in the same order as the menu, defined at the begging of platform.txt
 	if b.PlatformRelease.Menus != nil {
 		for _, menuOption := range b.PlatformRelease.Menus.FirstLevelKeys() {
-			for _, option := range allConfigs.FirstLevelKeys() {
-				if menuOption == option {
-					b.configOptions.Set(option, b.PlatformRelease.Menus.Get(menuOption))
-					break
-				}
+			if _, ok := allConfigOptions[menuOption]; ok {
+				b.configOptions.Set(menuOption, b.PlatformRelease.Menus.Get(menuOption))
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


arduino/cores/board.go
```diff
@@ -79,8 +79,17 @@ func (b *Board) buildConfigOptionsStructures() {
 
 	b.configOptions = properties.NewMap()
 	allConfigs := b.Properties.SubTree("menu")
-	for _, option := range allConfigs.FirstLevelKeys() {
-		b.configOptions.Set(option, b.PlatformRelease.Menus.Get(option))
+
+	// Used to show the config options in the same order as the menu, defined at the begging of platform.txt
+	if b.PlatformRelease.Menus != nil {
+		for _, menuOption := range b.PlatformRelease.Menus.FirstLevelKeys() {
+			for _, option := range allConfigs.FirstLevelKeys() {
```
cmaglie, line 86: `allConfigs.FirstLevelKeys()` can be moved out of the loop? AFAICS is constant so we can compute it just once.

cmaglie: BTW thinking about it a bit more...

Instead of `allConfigs.FirstLevelKeys()`, you can use `allConfigs.FirstLevelOff()` that returns a `map[string]*properties.Map`. Since you are testing the existence of `menuOption` in `allConfig` the check will be transformed into a map lookup instead of a loop.

You can also reuse the result of `allConfigs.FirstLevelOff()` later in the next loop, so basically there is no added complexity.

```go
    ...
	for option, optionProps := range allConfigs.FirstLevelOf() {   <---
		b.configOptionValues[option] = properties.NewMap()
		values := optionProps.FirstLevelKeys()
        ...
```

cmaglie (changes requested):
Nice idea! I think we can improve it a bit, I left some comments.

IMHO this PR can be removed from the draft state.

arduino/cores/board.go
alessio-perugini: I totally missed that function while implementing this :facepalm: Nice catch!

PR #2174, opened 2023-05-09
Identify managed platforms not tracked by a package index

alessio-perugini:

      before creating one)
- [x] The PR follows

We're introducing a warning message in the cli, when the user runs the upgrade of a platform that's no longer present in the additional-urls config. 
We also modify a bit some proto messages.

Right now when a platform is installed but the URL tracking the latest update for that platform gets removed from the config we're no longer to track the newest versions, and the user is not notified about it.

## What is the new behavior?

When a platform is installed but the additional URL is no longer present In the config, we send a warning message in the stdout when the user tries to perform a core upgrade for that platform.

Protobuff changes:
- `message Platform`:
  - added  bool indexed = 13;
  - added  bool missing_metadata = 14;
  
 - `message PlatformUpgradeResponse`:
   - added  Platform platform = 3;

The rpc call `PlatformUpgrade` now also returns the Platform object which FE can use the  `indexed` and `missing_metadata` to display some warnings. Like:
- indexed = false -> warning
- indexed = false AND missing_metdata = true -> warning 

_bonus_:
- missing_metadata = true -> we can decide to show a warning because the platform could be installed with the Arduino IDE 1.8

nope

## Other information

I think we should discuss if we need a system to track down various warning messages.


rpc/cc/arduino/cli/commands/v1/common.proto
```diff
@@ -102,6 +102,10 @@ message Platform {
   // A URL provided by the author of the platform's package, intended to point
   // to their online help service.
   HelpResources help = 12;
+  // If true the platform is indexed
+  bool indexed = 13;
+  // This field is true when the platform is installed with the Arduino IDE 1.8
```
cmaglie, line 107:
```suggestion
  // This field is true when the platform is installed with the Arduino IDE 1.8.
  // If the platform is also not indexed it may fail to work correctly in some circumstances,
  // and it may need to be re-installed.
```

rpc/cc/arduino/cli/commands/v1/core.proto
```diff
@@ -97,6 +97,8 @@ message PlatformUpgradeResponse {
   DownloadProgress progress = 1;
   // Description of the current stage of the upgrade.
   TaskProgress task_progress = 2;
+  // Platform to be upgraded.
```
cmaglie, line 100:
```suggestion
  // The upgraded platform.
```

internal/integrationtest/daemon/daemon_test.go
```diff
@@ -413,3 +400,158 @@ func TestDaemonLibrariesRescanOnInstall(t *testing.T) {
...
+		t.Run("and install.json is missing", func(t *testing.T) {
+			env, cli := createEnvForDaemon(t)
+			defer env.CleanUp()
+
+			grpcInst := cli.Create()
+			updateIndexAndInstallPlatform(cli, grpcInst, "3.1.0")
+
+			// remove installed.json{
+			x := env.RootDir().Join("A/packages/esp8266/hardware/esp8266/3.1.0/installed.json")
+			require.NoError(t, os.Remove(x.String()))
```
cmaglie, line 462:
```suggestion
			// remove installed.json
			x := env.RootDir().Join("A/packages/esp8266/hardware/esp8266/3.1.0/installed.json")
			require.NoError(t, x.Remove())
```

internal/integrationtest/daemon/daemon_test.go
```diff
@@ -413,3 +400,158 @@ func TestDaemonLibrariesRescanOnInstall(t *testing.T) {
...
+			grpcInst := cli.Create()
+			updateIndexAndInstallPlatform(cli, grpcInst, "3.1.0")
+
+			// remove additional urls and installed.json
+			{
+				require.NoError(t, cli.SetValue("board_manager.additional_urls", `[]`))
+				refreshInstance(t, grpcInst)
+
+				x := env.RootDir().Join("A/packages/esp8266/hardware/esp8266/3.1.0/installed.json")
+				require.NoError(t, os.Remove(x.String()))
```
cmaglie, line 510:
```suggestion
				require.NoError(t, x.Remove())
```

cmaglie (changes requested):
Some nitpicks, the PR looks good overall.

cmaglie (approved):

PR #2177, opened 2023-05-15
[skip changelog] Automate license headers checks in CI

alessio-perugini:

      before creating one)
- [x] The PR follows

This adds a new CI check for new go files missing the license headers. 
We have to put in the root of the folder a template called: `license_header.tpl`. This file will be used by `addlicense` to add any missing license header in the go file.

If we create a new go file and forget to add a license header we can spot it only by a manual CR which might be error-prone if the PR is large.

## What is the new behavior?

Every time a new golang file is pushed with a missing license header the CI will fail and warn the user.

## Other information

Any suggestion are welcomed

### DIFF
--- a/.github/workflows/check-go-task.yml
+++ b/.github/workflows/check-go-task.yml
@@ -219,6 +219,41 @@ jobs:
       - name: Check whether any tidying was needed
         run: git diff --color --exit-code
 
+  check-license-headers:
+    name: check-license-headers (${{ matrix.module.path }})
+    needs: run-determination
+    if: needs.run-determination.outputs.result == 'true'
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+
+      matrix:
+        module:
+          - path: ./
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Install Go
+        uses: actions/setup-go@v4
+        with:
+          go-version: ${{ env.GO_VERSION }}
+
+      - name: Install Task
+        uses: arduino/setup-task@v1
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          version: 3.x
+
+      - name: Check for missing license headers
+        env:
+          GO_MODULE_PATH: ${{ matrix.module.path }}
+        run: |
+          task go:add-license-headers
+          git diff --color --exit-code
+
   # Do a simple "smoke test" build for the modules with no other form of validation
   build:
     name: build (${{ matrix.module.path }})
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -142,6 +142,12 @@ tasks:
       - easyjson arduino/cores/packageindex/index.go
       - easyjson arduino/libraries/librariesindex/json.go
 
+  go:add-license-headers:
+    desc: Add missing go license headers
+    cmds:
+      - go install github.com/google/addlicense@v1.1.1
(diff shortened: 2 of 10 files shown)


cmaglie:
I tried to change the license in `main.go`, but running the check would not detect it. I guess it only generates the license header when completely missing?

Can this be changed to detect also if a license is not the correct one?

docsgen/main.go
```diff
@@ -1,8 +1,9 @@
-// Source: https://github.com/arduino/tooling-project-assets/blob/main/workflow-templates/assets/cobra/docsgen/main.go
```
cmaglie, line 1: It's a pity to lose this reference, it should be copied after the license

.github/workflows/check-go-task.yml
```diff
@@ -219,6 +219,55 @@ jobs:
...
+              -type d -name 'testdata' -prune -o \
+              -type d -name 'legacy' -prune -o \
+              -type f ! -name '*_grpc.pb.go' \
+              -type f ! -name 'json_easyjson.go' \
+              -type f ! -name 'index_easyjson.go' \
+              -regex ".*\.go$" -print
+          ); do
+          # We read only the license header, removing all the comments symbols and changing the date to be 2000. This will
+          # help us understand if the other part of the license are exactly the same as the template.
+          HEADER=$(head -n$N_LINES $file | sed -r 's| [0-9]{4} | 2000 |' | sed 's|// ||; s|^//||' )
```
cmaglie, line 263: We must keep the year range `2020-2023`. I did some research and found that:

1. the starting year must be kept to indicate when the copyright started (otherwise any previous work may not be considered under copyright)
2. the ending year must be kept to indicate when the copyright expires (it happens after the last year + some decades actually...)
3. we cannot use `2020-present`...
4. old headers should be updated every time a file is touched (for example if a file made in 2020 is modified in 2023 then the ending year in the header must be updated with `2020-2023`)

At this point, maybe the simplest solution is to just put the year range in the template file and update the template  + all source files every year?

.github/workflows/check-go-task.yml
```diff
@@ -219,6 +219,55 @@ jobs:
...
+          if [ "$LICENSE" != "$HEADER" ]; then
+            STATUS=1
+            echo "$file"
+          fi
+          done
+          exit $STATUS
```
cmaglie, line 269: I don't like this check because it just tells you that a file is wrong but didn't tell you what's wrong, and we don't have any way to re-run it locally. Also running `task go:add-license-headers` won't fix the wrong file.

Ideally, the action should just run `task go:add-license-header` and check if there are changes with `git diff`.
(more or less the same as the check for protoc files https://github.com/arduino/arduino-cli/blob/master/.github/workflows/check-protobuf-task.yml#L130).

In other words `task go:add-license-header` should write the files as they should be before pushing the PR.

cmaglie (changes requested):

alessio-perugini:
Ideally, we'd like to have a tool that can recognize and fix unwanted changes in the license header without being only limited to checking the presence of a license header. We tried to implement something quick with bash, but it's more tricky than expected. For example, do we need to provide a date range for our license headers? Or it's okay to keep the year when the file was created? In case we have to opt for a year range, should we always put 2020-CURRENT_YEAR?

We agreed that the most common mistake is to forget to add the license for a newly created file. Without overcomplicating this, the `addlicense` should be enough for now.

cmaglie (approved):

PR #2207, opened 2023-06-09
Return exit code 0 when a core is already up to date

alessio-perugini:

      before creating one)
- [x] The PR follows

Return exit code 0 if a platform is already at the latest version

Returns exit code 1.

## What is the new behavior?

Returns exit code 0.

### DIFF
--- a/internal/cli/core/upgrade.go
+++ b/internal/cli/core/upgrade.go
@@ -110,7 +110,7 @@ func Upgrade(inst *rpc.Instance, args []string, skipPostInstall bool) {
 		warningMissingIndex(response)
 		if err != nil {
 			if _, ok := err.(*arduino.PlatformAlreadyAtTheLatestVersionError); ok {
-				feedback.Fatal(err.Error(), feedback.Success)
+				feedback.Warning(err.Error())
 				continue
 			}
 
@@ -121,4 +121,19 @@ func Upgrade(inst *rpc.Instance, args []string, skipPostInstall bool) {
 	if hasBadArguments {
 		feedback.Fatal(tr("Some upgrades failed, please check the output for details."), feedback.ErrBadArgument)
 	}
+
+	feedback.PrintResult(&platformUpgradeResult{})
+}
+
+// This is needed so we can print warning messages in case users use --format json
+type platformUpgradeResult struct{}
+
+// Data implements feedback.Result.
+func (r *platformUpgradeResult) Data() interface{} {
+	return r
+}
+
+// String implements feedback.Result.
+func (r *platformUpgradeResult) String() string {
+	return ""
 }
--- a/internal/cli/core/upgrade.go
+++ b/internal/cli/core/upgrade.go
@@ -17,7 +17,6 @@ package core
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"os"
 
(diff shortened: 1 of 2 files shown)


internal/cli/core/upgrade.go
```diff
@@ -110,8 +109,8 @@ func Upgrade(inst *rpc.Instance, args []string, skipPostInstall bool) {
 		response, err := core.PlatformUpgrade(context.Background(), r, feedback.ProgressBar(), feedback.TaskProgress())
 		warningMissingIndex(response)
 		if err != nil {
-			if errors.Is(err, &arduino.PlatformAlreadyAtTheLatestVersionError{}) {
-				feedback.Print(err.Error())
+			if _, ok := err.(*arduino.PlatformAlreadyAtTheLatestVersionError); ok {
+				feedback.Fatal(err.Error(), feedback.Success)
```
alessio-perugini, line 113: It feels a bit hackish. We should refactor a bit the `feedback` package. But I think it's out of scope to do it here as I suspect it would require us to do quite a bit of work.

cmaglie: What about `feedback.Warning(...)`?

cmaglie: I'd also keep `errors.Is(...)` because:
* it's clearer
* it checks for nested errors too

cmaglie: Also, `feedback.Fatal` will immediately exit so the loop could not continue to process more arguments.

alessio-perugini: > What about feedback.Warning(...)?

If an user call the cli with --format json it won't output anything. Right now if we don't Fatal it won't flush the buffer as a json output.

> I'd also keep errors.Is(...) because:

We cannot use errors.Is because it checks that the pointers of the 2 errors are actually the same. This never happens and we have 2 options:
1. Do something like this:
```go
func (e *PlatformAlreadyAtTheLatestVersionError) Is(t error) bool {
	if _, ok := t.(*PlatformAlreadyAtTheLatestVersionError); ok {
		return true
	}
	return false
}
```
2. Use errors.As
```go
	var alreadyAtTheLatestVersionErr *arduino.PlatformAlreadyAtTheLatestVersionError
	if errors.As(err, &alreadyAtTheLatestVersionErr) {
		feedback.Warning(err.Error())
		continue
	}
```
> Also, feedback.Fatal will immediately exit so the loop could not continue to process more arguments.

Agree. (Right now it actually never entered in the errors.Is, so we're maintaining the same behavior, still incorrect tho)

Open questions:
1. We have to decide, in the case of only warning messages, if we want to show the json output. (This requires a refactoring of feedback package)

cmaglie: > We cannot use errors.Is because it checks that the pointers of the 2 errors are actually the same. This never happens and we have 2 options:

ah that's unfortunate :-/, ok let's keep your change then.

> If an user call the cli with --format json it won't output anything. Right now if we don't Fatal it won't flush the buffer as a json output.

In JSON mode, the warnings that were given via `feedback.Warning(...)` are accumulated and injected into the JSON response when `feedback.PrintResult` is called. BTW, I see that in this command we do not print any result, so nothing is printed, in fact, if you run it with `--format json` the output is empty, which is somewhat "wrong" since I'm expecting a JSON.

IMHO we should create a `coreUpgradeResult` struct that implements `feedback.Result` and print that (basically it's an empty result like `{ }`, or maybe a bare minimum object like `{ "success" : true }`).

cmaglie (approved):

PR #2212, opened 2023-06-13
Add Delete command to gRPC

MatteoPologruto:

      before creating one)
- [ ] The PR follows

Code enhancement

It is not possible to entirely remove a key from the config file using the gRPC interface. Setting it to an empty object using either `Merge` or `SetValue` would not work because viper would set it back to its default value, which is the one written in the config when it was loaded.
The only workaround would be to set the nested key to empty objects using the above mentioned commands:
```
network:
  proxy: ""
  user_agent_ext: ""
```

## What is the new behavior?
The key can be delete using the `Delete` command. Since it works in the same way as `config delete [key]`, the environment needs to be restarted to actually have the key removed from the settings.

No


commands/daemon/settings.go
```diff
@@ -144,3 +145,35 @@ func (s *SettingsService) Write(ctx context.Context, req *rpc.WriteRequest) (*rp
...
+	}
+
+	updatedSettings := viper.New()
+	for _, k := range keys {
+		updatedSettings.Set(k, configuration.Settings.Get(k))
+	}
+
+	if err := updatedSettings.WriteConfigAs(req.GetFilePath()); err != nil {
+		return nil, err
+	}
```
cmaglie, line 177: We have `SettingsService.Write` to do this operation, usually the other operations perform an update of the running environment (that is lost when the daemon terminates if not saved).

Could it be possible to replace the `configuration.Settings` with the `updatedSettings` at runtime?

commands/daemon/settings.go
```diff
@@ -144,3 +145,35 @@ func (s *SettingsService) Write(ctx context.Context, req *rpc.WriteRequest) (*rp
 	}
 	return &rpc.WriteResponse{}, nil
 }
+
+// Delete removes a key from the config file
+func (s *SettingsService) Delete(ctx context.Context, req *rpc.DeleteRequest) (*rpc.DeleteResponse, error) {
```
cmaglie, line 150: There is a very similar function implemented in `internal/cli/config/delete.go`: https://github.com/arduino/arduino-cli/blob/master/internal/cli/config/delete.go#L46-L72, it may be replaced with a function call to this one?

commands/daemon/settings.go
MatteoPologruto: It is not possible to delete a key from `configuration.Settings` at runtime, but we could set them as `""` as a workaround. Maybe something like this:
```
for _, k := range configuration.Settings.AllKeys() {
		if !strings.HasPrefix(k, toDelete) {
			keys = append(keys, k)
			continue
		}
		keyExists = true
		configuration.Settings.Set(k, "")
	}
```
I'm not using `SettingsService.Write` because it relies on `configuration.Settings` and the key would not be deleted as intended.

MatteoPologruto: Overwriting it is 👍

PR #2221, opened 2023-06-20
Improve cli exit codes

alessio-perugini:

      before creating one)
- [x] The PR follows

It adds different cmd exit codes.

Right now almost every error exit code is identified with the number 1

## What is the new behavior?

The new behavior expands the exit code numbers to distinguish better what kind of error we had. This is especially useful in commands like `compile` that perform various operations and it's not clear if it is something related to the compilation itself or other components it's using like sketches file problems or upload issues.

## Other information

I've experimented a bit at trying to throw very precise exit codes for every error we had, but it becomes way too unmanageable (plus we have limited exit code numbers available in UNIX systems) 
The tradeoff is having one exit code that expresses an error happening in a specific component. For example, if the compile commands use sketch, upload, and compile components and one of them fails we can return their associated exit codes.

Open points:
- Do we like this kind of handling?
- Should we return these enhanced exit codes only for commands that use different components, and it's not straightforward to understand what caused the problem? I see no value in creating a specific exit code for commands like `core install`, `core upgrade`, or `board details` as their logic uses only their dedicated component. 

I'll leave this PR open for feedback, especially from @kittaakos

### DIFF
--- a/internal/cli/compile/compile.go
+++ b/internal/cli/compile/compile.go
@@ -166,7 +166,7 @@ func runCompileCommand(cmd *cobra.Command, args []string) {
 
 	sk, err := sketch.LoadSketch(context.Background(), &rpc.LoadSketchRequest{SketchPath: sketchPath.String()})
 	if err != nil {
-		feedback.FatalError(err, feedback.ErrGeneric)
+		feedback.FatalError(err, feedback.ErrSketchError)
 	}
 	fqbn, port := arguments.CalculateFQBNAndPort(&portArgs, &fqbnArg, inst, sk.GetDefaultFqbn(), sk.GetDefaultPort(), sk.GetDefaultProtocol())
 
@@ -269,7 +269,7 @@ func runCompileCommand(cmd *cobra.Command, args []string) {
 		}
 
 		if err := upload.Upload(context.Background(), uploadRequest, stdOut, stdErr); err != nil {
-			feedback.Fatal(tr("Error during Upload: %v", err), feedback.ErrGeneric)
+			feedback.Fatal(tr("Error during Upload: %v", err), feedback.ErrBoardUpload)
 		}
 	}
 
@@ -366,7 +366,7 @@ func runCompileCommand(cmd *cobra.Command, args []string) {
 				}
 			}
 		}
-		feedback.FatalResult(res, feedback.ErrGeneric)
+		feedback.FatalResult(res, feedback.ErrCompilation)
 	}
 	feedback.PrintResult(res)
 }
--- a/internal/cli/feedback/errorcodes.go
+++ b/internal/cli/feedback/errorcodes.go
@@ -42,4 +42,13 @@ const (
 
 	// ErrBadArgument is returned when the arguments are not valid (7)
 	ErrBadArgument
+
+	// ErrSketchError is return when the occurs some error in the sketch package (8)
+	ErrSketchError
+
+	// ErrCompilation is returned when the compiler throw some errors (9)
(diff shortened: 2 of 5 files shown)


kittaakos:
Thank you for working on this. When the compilation fails, I see the expected non-zero exit code ✅ from the CLI, but the console output❓ is strange. Do you know if this is expected?

```
./arduino-cli version
arduino-cli  Version: git-snapshot Commit: d20e576d Date: 2023-06-28T06:54:50Z
```

```
cat ./test_sketches/broken/broken.ino 
trash
```

```
./arduino-cli compile -b arduino:avr:uno ./test_sketches/broken
/Users/a.kitta/dev/git/arduino-cli/test_sketches/broken/broken.ino:1:1: error: 'trash' does not name a type; did you mean 'tanh'?
 trash
 ^~~~~
 tanh


Used platform Version Path                                                                
arduino:avr   1.8.6   /Users/a.kitta/Library/Arduino15/packages/arduino/hardware/avr/1.8.6
```

```
echo $?
9
```

`Error during build: exit status 1` vs. `echo $? 9`. I assume exit status 1 comes from `avr-g++`, and CLI maps it to 9`. Altogether it is working as expected, but the console output is a bit misleading.

Thank you!

_(I continue with the review...)_

kittaakos:
I get the same exit code 9 when the platform is not installed, and the sketch has compiler errors:

```
./arduino-cli version 
arduino-cli  Version: git-snapshot Commit: d20e576d Date: 2023-06-28T06:54:50Z
```

```
./arduino-cli compile -b arduino:mbed:nanorp2040connect ./test_sketches/ok
Error during build: Platform 'arduino:mbed' not found: platform not installed
Try running `arduino-cli core install arduino:mbed`
```

```
echo $?                                                                   
9
```

```
./arduino-cli compile -b arduino:avr:uno ./test_sketches/broken           
/Users/a.kitta/dev/git/arduino-cli/test_sketches/broken/broken.ino:1:1: error: 'trash' does not name a type; did you mean 'tanh'?
 trash
 ^~~~~
 tanh


Used platform Version Path                                                                
arduino:avr   1.8.6   /Users/a.kitta/Library/Arduino15/packages/arduino/hardware/avr/1.8.6
```

```
echo $?                                                        
9
```

kittaakos:
I do not know if this is expected, but when I run the `compile` command with flawed arguments, I get back exit code 1, although there is a dedicated error for illegal arguments:

https://github.com/arduino/arduino-cli/blob/d20e576d2eebe0f8ac25c25d03b6b34416e8080e/internal/cli/feedback/errorcodes.go#L43-L44

```
./arduino-cli compile ./test_sketches/ok 
Missing FQBN (Fully Qualified Board Name)
```

```
echo $?                                  
1
```

alessio-perugini:
> I do not know if this is expected, but when I run the compile command with flawed arguments, I get back exit code 1, although there is a dedicated error for illegal arguments:

Yes, it's expected because some logic that seems easy to distinguish comes from another component that handles the compile action in this case.

We centralized under the `commands` pkg all the logic used by the gRPC interface and the CLI. For this reason, we cannot distinguish clearly some errors. The only way would be to wrap every error coming from that pkg and make a type switch to understand which error code to give. 
I've brought this to the team, and we cannot find a good enough approach to make this feature consistent. Since error codes are part of the API we cannot guarantee a solid way to always put errors in the right "category". We risk implementing a feature that becomes flaky and untrustable.

At the moment, we prefer to not implement this. We can re-evaluate this feature in the future depending on how many requests come from the community.

PR #2234, opened 2023-06-30
Fix 1200-bps touch DTR handling (Windows)

cmaglie:

      before creating one)
- [X] The PR follows

This PR disables the step that sets DTR low on the 1200-bps touch subroutine.

This reason why it was originally introduced:
https://github.com/arduino/Arduino/commit/a6909bdb49d99253b4e684365e72e5dce31a49a7

Why we are removing it now?
* Windows does preserve the state of the RTS/DTR bits on the next opening of the serial port.
* The serial library used in the Arduino IDE 1.8.x has a bug when trying to set DTR=false, on successive opening of the port the DTR line is set back high by the USB serial driver. This works differently from the serial library we use in the Arduino CLI, which sets DTR=false for good and this change is preserved on the next opening of the port.
* Having the serial port left in a state with DTR=false may cause problems with tools uploading later.

It may probably be applied to all OS but, for now, it will be disabled only for Windows to reduce the testing surface

Upload may fail on some boards.

## What is the new behavior?

Uploads should be successful...

No, in theory. A lot of testing is required to be sure.

### DIFF
--- a/arduino/serialutils/serialutils.go
+++ b/arduino/serialutils/serialutils.go
@@ -17,6 +17,7 @@ package serialutils
 
 import (
 	"fmt"
+	"runtime"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


umbynos (approved):

cmaglie:
Failing checks are OK ✅. They are happening because this PR is against an old release branch.
It will be eventually cherry-picked on `master`.

PR #2285, opened 2023-08-28
[BREAKING] Support only compatible platform versions

alessio-perugini:

      before creating one)
- [ ] The PR follows

- We added the `latest_compatible` field in the `cc.arduino.cli.commands.v1.Platform` gRPC message.
- We now always show the latest installable aka compatible version of a core.
- The upgrade and install commands now will always pick the latest compatible version. A user can force the installation of an incompatible/potentially-incompatible version if they provide a specific version like: `arduino-cli core install arduino:renesas_uno@1.0.2`

We always try to install (and show) the latest version even if it contains tools not available for the current os.

## What is the new behavior?

We show, install, and upgrade a core only if we're sure that all the tools are available for the current OS. Is possible to install a potentially incompatible version providing a specific version of a core.

yup


arduino/cores/cores.go
```diff
@@ -247,6 +258,18 @@ func (platform *Platform) GetAllReleasesVersions() []*semver.Version {
...
+func (platform *Platform) GetAllCompatibleReleasesVersions() []*semver.Version {
+	versions := []*semver.Version{}
+	for _, release := range platform.Releases {
+		if release.Incompatible {
+			continue
+		}
+		versions = append(versions, release.Version)
+	}
+	return versions
+}
```
cmaglie, line 271: `*Platform.GetAllCompatibleReleasesVersions() []*semver.Version` is used only in
`*Platform.latestCompatibleReleaseVersion() *semver.Version` that is used only in
`*Platform.GetLatestCompatibleRelease() *PlatformRelease`.

At this point, I'd like to drop the first two methods and implement directly the last one (we can cycle directly through the PlatformReleases and avoid a call to `platform.FindReleaseWithVersion(latestVersion)` to retrieve the PlatformRelease from the version.

arduino/cores/packageindex/index.go
```diff
@@ -147,6 +147,40 @@ func (index Index) MergeIntoPackages(outPackages cores.Packages) {
...
+						if tr.GetCompatibleFlavour() == nil {
+							return true
+						}
+					}
+					return false
+				}
+				pr.Incompatible = platformHasIncompatibleTools()
+			}
+		}
+	}
```
cmaglie, line 183: This loop is executed multiple times (after each package is loaded). This is redundant, ideally, it should be executed only once after all platforms have been loaded. Maybe a good place could be the `Build()` method before returning the package `Explorer`?

cmaglie (changes requested):

rpc/cc/arduino/cli/commands/v1/common.proto
```diff
@@ -108,6 +108,9 @@ message Platform {
   // If the platform is also not indexed it may fail to work correctly in some
   // circumstances, and it may need to be re-installed.
   bool missing_metadata = 14;
+  // Newest available version of the platform containing containing all
```
kittaakos, line 111:
```suggestion
  // Newest available version of the platform containing all
```

rpc/cc/arduino/cli/commands/v1/common.proto
```diff
@@ -108,6 +108,11 @@ message Platform {
   // If the platform is also not indexed it may fail to work correctly in some
   // circumstances, and it may need to be re-installed.
   bool missing_metadata = 14;
+  // Newest available version of the platform containing all compatible
+  // dependencies.
+  string latest_compatible = 15;
+  // True if the platform contains an incompatible dependency.
+  bool incompatible = 16;
```
kittaakos, line 115: Hi, why not `compatible`? Is there a reason to invert the logic and introduce a  `latest_compatible` property together with the `incompatible`?

---

Please review the `Platform` type. The new field (no matter if we name as `compatible` or `incompatible`) will provide sufficient info to clients to determine if a historical revision of a platform retrieved by `core search --all` is compatible. But storing all this info on the same level is very strange. Attaching the `website`, `email`, `boards`, etc. properties to the historical revision of a platform is reasonable. But `installed`, `latest`, `latest_compatible`, etc., are unrelated to a revision and describe the currently managed state by the CLI.

arduino/cores/packageindex/index.go
alessio-perugini: Nice catch! I've moved this piece of code in the commit function of the builder.

rpc/cc/arduino/cli/commands/v1/common.proto
alessio-perugini: >Hi, why not compatible

I've used the `incompatible` as we can benefit from the default zero value of protobuf. Since most platform versions are compatible, that field will be false, and gRPC doesn't need to send that information to the wire. This was my rationale. But TBH I'd pick a better name since the performance of the payload transfer is negligible here. If you have some suggestions about it please share. (Some ideas: `installable`, `compatible`)

> Please review the Platform type

Thank you for bringing this up! I agree with you. I've discussed with @cmaglie some ideas. 
Modifying the Platform type means that we have to change also: `PlatformListResponse`.
Below are the proposed API changes.

```diff
diff --git a/rpc/cc/arduino/cli/commands/v1/common.proto b/rpc/cc/arduino/cli/commands/v1/common.proto
index cf452120..9f3cafe2 100644
--- a/rpc/cc/arduino/cli/commands/v1/common.proto
+++ b/rpc/cc/arduino/cli/commands/v1/common.proto
@@ -74,10 +74,10 @@ message Programmer {
 message Platform {
   // Platform ID (e.g., `arduino:avr`).
   string id = 1;
-  // Version of the platform.
+  // Installed version of the platform
   string installed = 2;
-  // Newest available version of the platform.
-  string latest = 3;
+  // Current version of the platform
+  string version = 3;
   // Name used to identify the platform to humans (e.g., "Arduino AVR Boards").
   string name = 4;
   // Maintainer of the platform's package.
@@ -108,11 +108,8 @@ message Platform {
   // If the platform is also not indexed it may fail to work correctly in some
   // circumstances, and it may need to be re-installed.
   bool missing_metadata = 14;
-  // Newest available version of the platform containing all compatible
-  // dependencies.
-  string latest_compatible = 15;
   // True if the platform contains an incompatible dependency.
-  bool incompatible = 16;
+  bool incompatible = 15;
 }

 message InstalledPlatformReference {
diff --git a/rpc/cc/arduino/cli/commands/v1/core.proto b/rpc/cc/arduino/cli/commands/v1/core.proto
index 70ec4d46..46aa6d1a 100644
--- a/rpc/cc/arduino/cli/commands/v1/core.proto
+++ b/rpc/cc/arduino/cli/commands/v1/core.proto
@@ -138,6 +138,12 @@ message PlatformListRequest {
 }

 message PlatformListResponse {
-  // The installed platforms.
-  repeated Platform installed_platforms = 1;
+  // List of platforms.
+  repeated PlatformListItem platforms = 1;
+}
+
+message PlatformListItem {
+  Platform platform = 1;
+  string latest = 2;
+  string latest_compatible = 3;
 }

```

cmaglie: Actually, I would remove also the `installed` field from `Platform` (and maybe replicate it in `PlatformListItem`).

kittaakos: I did not forget about this. Thank you!

> platform retrieved by `core search --all`

Could you please make a diff for `core search`?

alessio-perugini: >Could you please make a diff for core search?

I'm not sure what I should diff against.
The diff I posted a couple of messages above (https://github.com/arduino/arduino-cli/pull/2285#discussion_r1334086979) 
is the only thing that changes from the latest commit I pushed and the proposed changes. All the rest of the API remains the same.

Here how it would look like the protobuf if we accept the proposed changes:
```protobuf
message PlatformListResponse {
  // List of platforms.
  repeated PlatformListItem platforms = 1;
}

message PlatformListItem {
  Platform platform = 1;
  string latest = 2;
  string latest_compatible = 3;
}

message Platform {
  // Platform ID (e.g., `arduino:avr`).
  string id = 1;
  // Installed version of the platform
  string installed = 2;
  // Current version of the platform
  string version = 3;
  // Name used to identify the platform to humans (e.g., "Arduino AVR Boards").
  string name = 4;
  // Maintainer of the platform's package.
  string maintainer = 5;
  // A URL provided by the author of the platform's package, intended to point
  // to their website.
  string website = 6;
  // Email of the maintainer of the platform's package.
  string email = 7;
  // List of boards provided by the platform. If the platform is installed,
  // this is the boards listed in the platform's boards.txt. If the platform is
  // not installed, this is an arbitrary list of board names provided by the
  // platform author for display and may not match boards.txt.
  repeated Board boards = 8;
  // If true this Platform has been installed manually in the user' sketchbook
  // hardware folder
  bool manually_installed = 9;
  // If true this Platform has been deprecated
  bool deprecated = 10;
  // Type of the platform.
  repeated string type = 11;
  // A URL provided by the author of the platform's package, intended to point
  // to their online help service.
  HelpResources help = 12;
  // If true the platform is indexed
  bool indexed = 13;
  // This field is true when the platform is installed with the Arduino IDE 1.8.
  // If the platform is also not indexed it may fail to work correctly in some
  // circumstances, and it may need to be re-installed.
  bool missing_metadata = 14;
  // True if the platform contains an incompatible dependency.
  bool incompatible = 15;
}
 ```

Comment (kittaakos) [reply rpc/cc/arduino/cli/commands/v1/common.proto:115]: Without assuming anything go, viper, etc. specific, this is my model change [proposal](https://github.com/arduino/arduino-cli/compare/master...kittaakos:arduino-cli:platform-grpc-api):
 - it pulls up the installation/versions/index/metadata props to a summary model,
 - a summary instance can have 1..* platform revisions.

Let me know what you think.

_(Note: I changed the model in a text editor. Hence, it might not be valid, but I hope it helps to get the idea.)_

Comment (kittaakos) [reply rpc/cc/arduino/cli/commands/v1/common.proto]: > I've used the `incompatible` as we can benefit from the default zero value of protobuf. Since most platform versions are compatible, that field will be false, and gRPC doesn't need to send that information to the wire.

+1 for a better API and using the `compatible` property instead of the `incompatible` one with this insignificant optimization.

**Review by umbynos — APPROVED** [review]
LGTM

File: arduino/cores/cores.go [hunk 5]
```diff
@@ -266,6 +303,23 @@ func (platform *Platform) latestReleaseVersion() *semver.Version {
...
+		return nil
+	}
+	max := versions[0]
+
+	for i := 1; i < len(versions); i++ {
+		if versions[i].GreaterThan(max) {
+			max = versions[i]
+		}
+	}
+
```
cmaglie, line 319:
```suggestion
	var max *semver.Version
	for v := range versions {
		if max == nil || v.GreaterThan(max) {
			max = v
		}
	}
```

arduino/cores/cores.go
```diff
@@ -75,6 +75,7 @@ type PlatformRelease struct {
 	PluggableDiscoveryAware bool                          `json:"-"` // true if the Platform supports pluggable discovery (no compatibility layer required)
 	Monitors                map[string]*MonitorDependency `json:"-"`
 	MonitorsDevRecipes      map[string]string             `json:"-"`
+	Compatible              bool                          `json:"-"` // true if at all ToolDependencies are available for the current OS/ARCH.
```
cmaglie, line 78: Unless necessary, I'd keep new fields private by default (we were already bitten by this practice...).
I see that you also put an `IsCompatible` method, that's 👍🏼

arduino/cores/packagemanager/package_manager_test.go
```diff
@@ -744,6 +744,7 @@ func TestFindToolsRequiredFromPlatformRelease(t *testing.T) {
 func TestFindPlatformReleaseDependencies(t *testing.T) {
 	pmb := NewBuilder(nil, nil, nil, nil, "test")
 	pmb.LoadPackageIndexFromFile(paths.New("testdata", "package_tooltest_index.json"))
+	pmb.calculateCompatibleReleases()
```
cmaglie, line 747: shouldn't this be called automatically inside `pmb.Build()`?

internal/cli/arguments/completion.go
```diff
@@ -112,8 +112,11 @@ func GetInstallableCores() []string {
 	})
 	var res []string
 	// transform the data structure for the completion
-	for _, i := range platforms.SearchOutput {
-		res = append(res, i.GetMetadata().GetId()+"\t"+i.GetReleases()[i.GetLatestVersion()].GetName())
+	for _, i := range platforms.GetSearchOutput() {
+		if i.GetLatestCompatibleVersion() == "" {
+			continue
+		}
+		res = append(res, i.GetMetadata().GetId()+"\t"+i.GetLatestCompatibleRelease().GetName())
```
cmaglie, line 119: We can spare the call to `i.GetLatestCompatibleRelease()` if we reuse the result of `i.GetLatestCompatibleRelease()` 

```suggestion
	for _, i := range platforms.GetSearchOutput() {
		if latest := i.GetLatestCompatibleVersion(); latest != nil {
			res = append(res, i.GetMetadata().GetId()+"\t"+latest.GetName())
		}
```

internal/cli/core/search.go
```diff
@@ -108,15 +111,24 @@ func (sr searchResults) String() string {
...
-			if platform.Deprecated {
-				name = fmt.Sprintf("[%s] %s", tr("DEPRECATED"), name)
+			// When allVersions is not requested we only show the latest compatible version
+			if !sr.allVersions {
+				if latestCompatible := platform.GetLatestCompatibleRelease(); latestCompatible != nil {
+					name := latestCompatible.Name
+					if latestCompatible.Deprecated {
+						name = fmt.Sprintf("[%s] %s", tr("DEPRECATED"), latestCompatible.Name)
+					}
+					t.AddRow(platform.Id, latestCompatible.Version, name)
```
cmaglie, line 121: make an anonymous function to add a PlatformRelease in a table row (to be reused a few lines below)

rpc/cc/arduino/cli/commands/v1/common.proto
```diff
@@ -91,6 +91,9 @@ message PlatformSummary {
   string installed_version = 3;
   // The latest available version of the platform, or empty if none available
   string latest_version = 4;
+  // The latest available version of the platform containing all compatible
+  // dependencies, or empty if none available
+  string latest_compatible_version = 5;
```
cmaglie, line 96: I'm not sure about this field, I'm for removing it (the latest non-compatible version, if really needed, could be determined from the `releases` map when the request is done with `all_version==true` set).

arduino/cores/cores.go
alessio-perugini: Unfortunately, we cannot make it private as is used by the `packagemanager` pkg in the commit func.

arduino/cores/packagemanager/package_manager_test.go
alessio-perugini: Nope, it's called by the commit function returned by the `NewBuilder` method receiver.

rpc/cc/arduino/cli/commands/v1/common.proto
alessio-perugini: >I'm not sure about this field, I'm for removing it (the latest non-compatible version, if really needed, could be determined from the releases map when the request is done with all_version==true set).

@kittaakos WDYT? :point_up:

kittaakos: No problem. Please remove it if the info can be inferred.

The `latest_version` can be inferred too 🙃

rpc/cc/arduino/cli/commands/v1/common.proto
```diff
@@ -141,6 +144,9 @@ message PlatformRelease {
   bool missing_metadata = 7;
   // True this release is deprecated
   bool deprecated = 8;
+  // True if the platform dependencies are available for the current OS. This
+  // also means that the platform is installable.
```
kittaakos, line 148: Besides the OS, please mention the arch in the public API. It's nicely documented in the code. Thanks

> // true if at all ToolDependencies are available for the current OS/ARCH.

cmaglie (approved):
Just a small nit pick.
Incredible work! 👍🏼

anttiryt:
Do you have a solution for https://github.com/arduino/ctags/issues/19 ?

PR #2307, opened 2023-09-10
add "config get" command to print settings values

ardnew:

      before creating one)
- [x] The PR follows
- [ ] ~`UPGRADING.md` has been updated with a migration guide (for breaking changes)~
- [ ] ~`configuration.schema.json` updated if new parameters are added.~

Feature addition

There are `config` commands to `add`/`set` individual settings, but retrieving settings requires the user to `dump` all settings and parse the output with either JSON/YAML to retrieve an individual.

## What is the new behavior?

With the addition of a `config get` command, the user can retrieve individual configuration settings without having to parse JSON/YAML.

Not a breaking change

## Other information

None

### DIFF
--- a/internal/cli/config/get.go
+++ b/internal/cli/config/get.go
@@ -17,7 +17,6 @@ package config
 
 import (
 	"encoding/json"
-	"fmt"
 	"os"
 
 	"github.com/arduino/arduino-cli/commands/daemon"
@@ -26,6 +25,7 @@ import (
 	rpc "github.com/arduino/arduino-cli/rpc/cc/arduino/cli/commands/v1"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
+	"gopkg.in/yaml.v3"
 )
 
 func initGetCommand() *cobra.Command {
@@ -77,5 +77,10 @@ func (gr getResult) Data() interface{} {
 }
 
 func (gr getResult) String() string {
-	return fmt.Sprintf("%v", gr.resp)
+	gs, err := yaml.Marshal(gr.resp)
+	if err != nil {
+		// Should never happen
+		panic(tr("unable to marshal config to YAML: %v", err))
+	}
+	return string(gs)
 }
--- a/internal/cli/config/get.go
+++ b/internal/cli/config/get.go
@@ -29,12 +29,11 @@ import (
 func initGetCommand() *cobra.Command {
 	getCommand := &cobra.Command{
 		Use:   "get",
-		Short: tr("Gets a setting value."),
-		Long:  tr("Gets a setting value."),
+		Short: tr("Gets settings key values."),
+		Long:  tr("Gets settings key values."),
 		Example: "" +
-			"  " + os.Args[0] + " config get logging.level\n" +
-			"  " + os.Args[0] + " config get logging.file\n" +
-			"  " + os.Args[0] + " config get sketch.always_export_binaries\n" +
+			"  " + os.Args[0] + " config get logging\n" +
+			"  " + os.Args[0] + " config get logging.level logging.file\n" +
 			"  " + os.Args[0] + " config get board_manager.additional_urls",
 		Args: cobra.MinimumNArgs(1),
 		Run:  runGetCommand,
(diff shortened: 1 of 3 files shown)


MatteoPologruto:
Hello @ardnew! Thanks for your contribution.
Is this PR ready to be reviewed? If so, can you please rebase your branch onto `arduino:master` to include the latest commits and resolve any conflicts?
If you have any questions, please feel free to ask!

ardnew:
@MatteoPologruto Will try to rebase this PR very soon to fix the conflicts.

Also, there are 2 items in the PR template that have not been completed, which is why this was originally posted as a draft. Not sure if they are essential or if I can submit the PR without them (after resolving conflicts)?

internal/cli/config/get.go
```diff
@@ -0,0 +1,88 @@
...
+
+	var value interface{}
+	switch kind {
+	case reflect.Slice:
+		value = configuration.Settings.GetStringSlice(key)
+	case reflect.String:
+		value = configuration.Settings.GetString(key)
+	case reflect.Bool:
+		value = configuration.Settings.GetBool(key)
+	}
```
MatteoPologruto, line 65: We should replace this with a call to the function [SettingsGetValue](https://github.com/arduino/arduino-cli/blob/master/commands/daemon/settings.go#L93-L121). The reason is that we want to use the CLI as "wrapper" for the gRPC interface. Another command that works this way is the `config delete` command, which was modified in #2212. Since, in this case, the gRPC function has already been implemented, you can skip directly to [this commit](https://github.com/arduino/arduino-cli/pull/2212/commits/9ae946190358d57380f184311cfe16f9bf39f796).

MatteoPologruto (changes requested):

internal/cli/config/get.go
```diff
@@ -53,27 +55,26 @@ func runGetCommand(cmd *cobra.Command, args []string) {
...
 }
 
 func (gr getResult) String() string {
-	gs, ok := gr.data.(string)
-	if !ok {
-		// Should never happen
-		panic(tr("Cannot get key %s value as string: %v", gr.key, gr.data))
-	}
-	return gs
+	return fmt.Sprintf("%v", gr.resp)
```
MatteoPologruto, line 79: Printing the result this way is not optimal. When I try to get configs using a key that returns a map, this is the output:
```
$ ./arduino-cli.exe config get metrics
map[addr::9090 enabled:true]
```
It is, of course, correct, but we should display it more clearly. I think that changing this `String()` function to match the one from [config dump](https://github.com/arduino/arduino-cli/blob/master/internal/cli/config/dump.go#L55-L62). Trying it locally, this is the output I get:
```
$ ./arduino-cli.exe config get metrics
addr: :9090
enabled: true

$ ./arduino-cli.exe config get metrics.addr
:9090
```

MatteoPologruto (changes requested):

internal/cli/config/get.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+func runGetCommand(cmd *cobra.Command, args []string) {
+	logrus.Info("Executing `arduino-cli config get`")
+
+	svc := daemon.ArduinoCoreServerImpl{}
+	for _, toGet := range args {
+		resp, err := svc.SettingsGetValue(cmd.Context(), &rpc.SettingsGetValueRequest{Key: toGet})
+		if err != nil {
+			feedback.Fatal(tr("Cannot get the key %[1]s: %[2]v", toGet, err), feedback.ErrGeneric)
```
cmaglie, line 56:
```suggestion
			feedback.Fatal(tr("Cannot get the configuration key %[1]s: %[2]v", toGet, err), feedback.ErrGeneric)
```

internal/cli/config/get.go
```diff
@@ -0,0 +1,80 @@
...
+		}
+		var result getResult
+		err = json.Unmarshal([]byte(resp.GetJsonData()), &result.resp)
+		if err != nil {
+			feedback.Fatal(tr("Cannot parse JSON for key %[1]s: %[2]v", toGet, err), feedback.ErrGeneric)
```
cmaglie, line 61: JsonData comes directly from a `json.Marshal` so it should always be valid JSON.

```suggestion
			// Should never happen...
			panic(fmt.Sprintf("Cannot parse JSON for key %[1]s: %[2]v", toGet, err))
```

internal/cli/config/get.go
ardnew: I was expecting the user to provide `--format` flag if they wished to have a particular format:

```
$ arduino-cli config get metrics --format yaml
addr: :9090
enabled: true

$ arduino-cli config get metrics.addr --format yaml
:9090
```

Otherwise, this `text`  format is impossible to generate, which I find useful as an unambiguous syntax reference.

On the other hand, the subcommands of `config` (`dump`, `get`) should probably generate the same format by default, and people probably already depend on the format of `dump`.

So I'll concede and accept the change :)

MatteoPologruto:
@ardnew code LGTM 🎉 
You only need to update the tests according to the changes you made. After that, I'd say we are ready to merge!

MatteoPologruto (approved):
Great job! Thanks for your contribution @ardnew 🎉

PR #2350, opened 2023-10-02
Update deps

alessio-perugini:

      before creating one)
- [ ] The PR follows

Each commit bumps the version of a dependency.
The critical ones are:
- viper
- cobra
- yaml v3
We can cherry-pick some of them and put them in the next RC.


go.mod
```diff
@@ -16,6 +16,7 @@ require (
 	github.com/djherbis/buffer v1.1.0
 	github.com/djherbis/nio/v3 v3.0.1
 	github.com/fatih/color v1.13.0
+	github.com/go-git/go-git/v5 v5.4.2
```
umbynos, line 19: I would use the latest version https://github.com/go-git/go-git/releases/tag/v5.9.0

umbynos (commented):
Left some comments

go.mod
alessio-perugini: Unfortunately version after `v5.4.2` is causing some errors

arduino/sketch/profiles.go
```diff
@@ -26,21 +26,23 @@ import (
...
 
 // Project represents the sketch project file
 type Project struct {
-	Profiles        Profiles `yaml:"profiles"`
-	DefaultProfile  string   `yaml:"default_profile"`
-	DefaultFqbn     string   `yaml:"default_fqbn"`
-	DefaultPort     string   `yaml:"default_port,omitempty"`
-	DefaultProtocol string   `yaml:"default_protocol,omitempty"`
+	ProfilesRaw     yaml.Node  `yaml:"profiles"`
+	Profiles        []*Profile `yaml:"-"`
```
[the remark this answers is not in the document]
alessio-perugini: In the new yamlv3 package, they removed the `MapSlice` which would parse the map and respect the order of the key. 
So to achieve the same thing we have to use the `yaml.Node` type and do by hand what essentially the `MapSlice` was doing. The only downside is that we have to bring the `ProfilesRaw` just for parsing purposes. I thought that bringing a new struct for parsing purposes and copying the result back to the original struct was a bit too much in this case.

arduino/sketch/profiles.go
```diff
@@ -260,5 +270,7 @@ func LoadProjectFile(file *paths.Path) (*Project, error) {
 	if err := yaml.Unmarshal(data, &res); err != nil {
 		return nil, err
 	}
+	res.Profiles = res.getProfiles()
```
cmaglie, line 273: Can't we do this inside the `YamlUnmarshal`? We may forget to produce the profiles for example we are trying to do another `yaml.Unmarshal` here: https://github.com/arduino/arduino-cli/blob/f561da0de04593e3eb24acace22eacd77cb084c4/arduino/sketch/yaml.go#L77

arduino/sketch/profiles.go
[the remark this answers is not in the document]
cmaglie: > I thought that bringing a new struct for parsing purposes and copying the result back to the original struct was a bit too much in this case.

IMHO it's better to use a support structure, I don't like much this `ProfilesRaw` field, it's a publicly visible field that is confusing (should I look inside ProfilesRaw or Profiles? Should they be kept synchronized?). I see also that we cannot even make it private because we must unmarshal.
Another advantage is that the support structure may be discarded after being used.

arduino/sketch/profiles.go
```diff
@@ -71,18 +90,9 @@ func (p *Profiles) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
-	for _, profileData := range profilesData {
-		profileName, ok := profileData.Key.(string)
-		if !ok {
-			return fmt.Errorf("invalid profile name: %v", profileData.Key)
-		}
-		profile := unmarshaledProfiles[profileName]
-		profile.Name = profileName
+	for k, v := range unmarshaledProfiles {
+		profile := v
+		profile.Name = k
```
cmaglie, line 95: If you're not using the type `type Profiles []*Profile` anymore it can be removed completely. BTW I'm not sure, it may be helpful to automate the unmarshaling (see my other comments)

umbynos (approved):

cmaglie (approved):

PR #2373, opened 2023-10-17
Added search using qualifier[:=]value syntax

zvonler:
      before creating one)
- [x] The PR follows

More powerful search using a syntax of `qualifier[:=]value` that allows for partial or exact match (still case-insensitive) against specific library metadata fields.

The behavior was discussed in https://github.com/arduino/arduino-cli/issues/1535

All `lib search` commands compare the provided query terms against a string constructed from several of each library's metadata fields.

## What is the new behavior?

All queries that do not contain either a ':' or '=' character are handled exactly as before.

Queries that do contain one of the above characters are parsed using new logic that allows for double-quoted strings to contain spaces, and allows for backslash-escaped literal double-quote characters. The resulting tokens are examined for case-insensitive match with the known qualifier names, which are these:

- Architecture
- Author
- Category
- Dependencies
- License
- Maintainer
- Name
- Paragraph
- Provides
- Sentence
- Types
- Versions
- Website

A query token that is one of these names followed by a colon character (':') indicates to search for the following `value` anywhere in the named metadata field. A query token that is one of these names followed by an equals sign ('=') indicates to return only libraries with the named metadata field that matches exactly (case-insensitive) the following `value`.

Query tokens that do not match either of the two previous tests are handled using the original search algorithm, so that it is still possible to search for, e.g. "sentence" and get the same results as before.

The double-quote handling means one can search for, e.g. `name="Arduino Low Power"`, and the backslash-escaping means one can search for literal double-quotes in specific fields, e.g. `sentence:\"`.

A query that mixes the `qualifier[:=]value` syntax with non-matching tokens will return only libraries that successfully match all of the tokens, consistent with previous behavior.

It does not

## Other information

~~The tests added here use "live" data in the sense that the query the actual library manager index. That approach is used by the existing tests, so I tried to target the same libraries as they do so as not to introduce additional hidden dependency on the index contents, but there is a chance the tests added here will break due to future library manager index changes. I would add more unit tests if there were a framework for populating a mock index to run tests against. I did perform several tests by hand on the command-line of the escaping and quoting.~~

~~Where should this new behavior be documented?~~

### DIFF
--- a/commands/lib/search.go
+++ b/commands/lib/search.go
@@ -38,24 +38,141 @@ func LibrarySearch(ctx context.Context, req *rpc.LibrarySearchRequest) (*rpc.Lib
 	return searchLibrary(req, lm), nil
 }
 
+// MatcherTokensFromQueryString parses the query string into tokens of interest
+// for the qualifier-value pattern matching.
+func MatcherTokensFromQueryString(query string) []string {
+	escaped := false
+	quoted := false
+	tokens := []string{}
+	sb := &strings.Builder{}
+
+	for _, r := range query {
+		// Short circuit the loop on backslash so that all other paths can clear
+		// the escaped flag.
+		if !escaped && r == '\\' {
+			escaped = true
+			continue
+		}
+
+		if r == '"' {
+			if !escaped {
+				quoted = !quoted
+			} else {
+				sb.WriteRune(r)
+			}
+		} else if !quoted && r == ' ' {
+			tokens = append(tokens, strings.ToLower(sb.String()))
+			sb.Reset()
+		} else {
+			sb.WriteRune(r)
+		}
+		escaped = false
+	}
+	if sb.Len() > 0 {
+		tokens = append(tokens, strings.ToLower(sb.String()))
+	}
+
+	return tokens
+}
+
+// DefaulLibraryMatchExtractor returns a string describing the library that
+// is used for the simple search.
+func DefaultLibraryMatchExtractor(lib *librariesindex.Library) string {
+	res := lib.Name + " " +
+		lib.Latest.Paragraph + " " +
+		lib.Latest.Sentence + " " +
+		lib.Latest.Author + " "
+	for _, include := range lib.Latest.ProvidesIncludes {
+		res += include + " "
+	}
+	return res
+}
+
+// MatcherFromQueryString returns a closure that takes a library as a
+// parameter and returns true if the library matches the query.
+func MatcherFromQueryString(query string) func(*librariesindex.Library) bool {
+	// A qv-query is one using <qualifier>[:=]<value> syntax.
+	qvQuery := strings.Contains(query, ":") || strings.Contains(query, "=")
+
+	if !qvQuery {
+		queryTerms := utils.SearchTermsFromQueryString(query)
+		return func(lib *librariesindex.Library) bool {
+			return utils.Match(DefaultLibraryMatchExtractor(lib), queryTerms)
+		}
+	}
+
+	joinedStrings := func(strs []string) string {
+		return strings.Join(strs, " ")
+	}
+
+	qualifiers := []struct {
+		key       string
+		extractor func(*librariesindex.Library) string
+	}{
+		// The library name comes from the Library object.
+		{"name", func(lib *librariesindex.Library) string { return lib.Name }},
+
+		// All other values come from the latest Release.
+		{"architectures", func(lib *librariesindex.Library) string { return joinedStrings(lib.Latest.Architectures) }},
+		{"author", func(lib *librariesindex.Library) string { return lib.Latest.Author }},
+		{"category", func(lib *librariesindex.Library) string { return lib.Latest.Category }},
+		{"dependencies", func(lib *librariesindex.Library) string {
+			names := []string{}
+			for _, dep := range lib.Latest.Dependencies {
+				names = append(names, dep.GetName())
+			}
+			return joinedStrings(names)
+		}},
+		{"maintainer", func(lib *librariesindex.Library) string { return lib.Latest.Maintainer }},
+		{"paragraph", func(lib *librariesindex.Library) string { return lib.Latest.Paragraph }},
+		{"sentence", func(lib *librariesindex.Library) string { return lib.Latest.Sentence }},
+		{"types", func(lib *librariesindex.Library) string { return joinedStrings(lib.Latest.Types) }},
+		{"version", func(lib *librariesindex.Library) string { return lib.Latest.Version.String() }},
+		{"website", func(lib *librariesindex.Library) string { return lib.Latest.Website }},
+	}
+
+	queryTerms := MatcherTokensFromQueryString(query)
+
+	return func(lib *librariesindex.Library) bool {
+		matched := true
+		for _, term := range queryTerms {
+
+			// Flag indicating whether the search term matched a known qualifier
+			knownQualifier := false
+
+			for _, q := range qualifiers {
+				if strings.HasPrefix(term, q.key+":") {
+					target := strings.TrimPrefix(term, q.key+":")
+					matched = (matched && utils.Match(q.extractor(lib), []string{target}))
+					knownQualifier = true
+					break
+				} else if strings.HasPrefix(term, q.key+"=") {
+					target := strings.TrimPrefix(term, q.key+"=")
+					matched = (matched && strings.ToLower(q.extractor(lib)) == target)
+					knownQualifier = true
+					break
+				}
+			}
+
+			if !knownQualifier {
+				matched = (matched && utils.Match(DefaultLibraryMatchExtractor(lib), []string{term}))
+			}
+		}
+		return matched
+	}
+}
+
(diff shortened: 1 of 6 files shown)


zvonler:
PR addresses https://github.com/arduino/arduino-cli/issues/1535

alessio-perugini:
Thank you for the PR! The review will take some time bear with us :pray:

zvonler:
> Thank you for the PR! The review will take some time bear with us 🙏

Thanks. I'm still just learning Go, so very interested in the review.

commands/lib/search.go
```diff
@@ -38,24 +38,141 @@ func LibrarySearch(ctx context.Context, req *rpc.LibrarySearchRequest) (*rpc.Lib
...
+			}
+			return joinedStrings(names)
+		}},
+		{"maintainer", func(lib *librariesindex.Library) string { return lib.Latest.Maintainer }},
+		{"paragraph", func(lib *librariesindex.Library) string { return lib.Latest.Paragraph }},
+		{"sentence", func(lib *librariesindex.Library) string { return lib.Latest.Sentence }},
+		{"types", func(lib *librariesindex.Library) string { return joinedStrings(lib.Latest.Types) }},
+		{"version", func(lib *librariesindex.Library) string { return lib.Latest.Version.String() }},
+		{"website", func(lib *librariesindex.Library) string { return lib.Latest.Website }},
+	}
```
alessio-perugini, line 132:
```suggestion
var qualifiers map[string]func(lib *librariesindex.Library) string = map[string]func(lib *librariesindex.Library) string{
	"name":          func(lib *librariesindex.Library) string { return lib.Name },
	"architectures": func(lib *librariesindex.Library) string { return strings.Join(lib.Latest.Architectures, " ") },
	"author":        func(lib *librariesindex.Library) string { return lib.Latest.Author },
	"category":      func(lib *librariesindex.Library) string { return lib.Latest.Category },
	"dependencies": func(lib *librariesindex.Library) string {
		names := make([]string, len(lib.Latest.Dependencies))
		for i, dep := range lib.Latest.Dependencies {
			names[i] = dep.GetName()
		}
		return strings.Join(names, " ")
	},
	"maintainer": func(lib *librariesindex.Library) string { return lib.Latest.Maintainer },
	"paragraph":  func(lib *librariesindex.Library) string { return lib.Latest.Paragraph },
	"sentence":   func(lib *librariesindex.Library) string { return lib.Latest.Sentence },
	"types":      func(lib *librariesindex.Library) string { return strings.Join(lib.Latest.Types, " ") },
	"version":    func(lib *librariesindex.Library) string { return lib.Latest.Version.String() },
	"website":    func(lib *librariesindex.Library) string { return lib.Latest.Website },
}
```
Those qualifiers are always the same so I'd move them outside the function and use a map instead.

commands/lib/search.go
```diff
@@ -38,24 +38,141 @@ func LibrarySearch(ctx context.Context, req *rpc.LibrarySearchRequest) (*rpc.Lib
...
+					break
+				} else if strings.HasPrefix(term, q.key+"=") {
+					target := strings.TrimPrefix(term, q.key+"=")
+					matched = (matched && strings.ToLower(q.extractor(lib)) == target)
+					knownQualifier = true
+					break
+				}
+			}
+
+			if !knownQualifier {
```
alessio-perugini, line 157: With `qualifiers` being a map we can avoid this for loop. Changing how we extract the qualifier key from `term`: if the key exists in the map we take the associated extractor, otherwise, we fallback to the default one.

commands/lib/search_test.go
alessio-perugini, line 1: > I would add more unit tests if there were a framework for populating a mock index to run tests against

The library manager is loaded from `fullIndexPath` (var fullIndexPath = paths.New("testdata", "full")) and points to a folder containing a snapshot of a library_index.json. (it acts like a mock)
If you would like to create a new `library_index.json` or multiple ones feel free to do so under `commands/lib/testdata/*`
(each new `library_index.json` must be inside a dedicated folder)

commands/lib/search_test.go
```diff
@@ -94,3 +94,27 @@ func TestSearchLibraryFields(t *testing.T) {
...
+	res := query("name:FlashStorage")
+	require.Len(t, res, 7)
+
+	res = query("name=FlashStorage")
+	require.Len(t, res, 1)
+	require.Equal(t, "FlashStorage", res[0])
+
+	res = query("name=\"Painless Mesh\"")
+	require.Len(t, res, 1)
+	require.Equal(t, "Painless Mesh", res[0])
```
alessio-perugini, line 119: I would like to add some more tests regarding:
- the usage of multiple qualifiers
- multiple qualifiers and a query (`author:arduino architecture:avr led`) 
- mixing different multiple qualifiers (`author:arduino architecture=avr`)
- providing some "invalid" inputs like wrong escaping, missing closed `"` etc (example: `name=\"mesh "`

alessio-perugini (changes requested):
The implementation side looks good, a bit too functional to my taste. I tried to see if we could shape it differently, but to be honest I didn't come up with something that didn't feel over-engineered. 

Would be cool to add some basic integration tests in `internal/integrationtest/lib/lib_test.go`, it might highlight some edge cases coming from the cli side. Not super necessarily as we already have unit tests. If you want to skip them it's fine.

Moving all the matcher-related stuff into a new file called `search_matcher.go` probably will bring more clarity to the codebase.

>Where should this new behavior be documented?

Most likely:
- expanding the description in the cli-side (we take cobra related description and autogenerate the doc). 
- and adding a dedicated page

But @per1234 is the best person to answer this :mage: 

---

We're very happy with the quality we're seeing in this PR, you're doing great :muscle:

commands/lib/search_test.go
zvonler: Ah, apologies for not chasing down where that index was coming from, I assumed it was running against the real one. I'm adding a new JSON file for the qualified search tests now.

commands/lib/search_test.go
zvonler: Regarding the "invalid" inputs - since there is no mechanism to indicate syntax errors, should it try to be as permissive as possible?

A specific example is `name:"First Second` (no closing double-quote). The current code will return everything after the opening quote as a single token, meaning that this search returns the same results as `name:"First Second"`. I think that's reasonable behavior and there wouldn't be benefit in rejecting the former query as malformed.

zvonler:
Sorry for the churn, I did the force push because it was telling me there was a merge conflict with the trunk, and I didn't understand I wasn't running integration tests as part of `task go:test` so the integration test was broken initially.

I think those changes address all the issues raised except documentation. I can take a swing at that if it makes sense, otherwise happy to coordinate with someone else writing the prose.

commands/lib/search_test.go
alessio-perugini: @zvonler yeah, it's totally fine to be permissive. I quoted "invalid" because we can handle such cases (just wanted to make it obvious on the test side :nerd_face: ). I quickly looked at the tests, and it's looking great.
I'll be traveling this weekend so you'll have to be patient till Monday:pray:

commands/lib/search_matcher.go
```diff
@@ -0,0 +1,137 @@
...
+					// Unknown qualifier names revert to basic search terms.
+					matched = (matched && utils.Match(defaultLibraryMatchExtractor(lib), []string{term}))
+				}
+			} else {
+				// Terms that do not use qv-syntax are handled as usual.
+				matched = (matched && utils.Match(defaultLibraryMatchExtractor(lib), []string{term}))
+			}
+		}
+		return matched
+	}
```
alessio-perugini, line 136:
```suggestion
	return func(lib *librariesindex.Library) bool {
		matched := true
		for _, term := range queryTerms {
			if sepIdx := strings.IndexAny(term, ":="); sepIdx != -1 {
				qualifier, separator, target := term[:sepIdx], term[sepIdx], term[sepIdx+1:]
				if extractor, ok := qualifiers[qualifier]; ok {
					switch separator {
					case ':':
						matched = (matched && utils.Match(extractor(lib), []string{target}))
						continue
					case '=':
						matched = (matched && strings.ToLower(extractor(lib)) == target)
						continue
					}
				}
			}
			// We perform the usual match in the following cases:
			// 1. Unknown qualifier names revert to basic search terms.
			// 2. Terms that do not use qv-syntax.
			matched = (matched && utils.Match(defaultLibraryMatchExtractor(lib), []string{term}))
		}
		return matched
	}
```
Little refactor to bring more clarity :nerd_face:

alessio-perugini (changes requested):

commands/lib/search_matcher.go
zvonler: Nice

alessio-perugini:
@zvonler We're almost there :smiley:,  for the docs it's okay to write them directly in cobra command fields. (`internal/cli/lib/search.go`)
In the `Long` property you can explain the new feature: which qualifiers are allowed, the operators (`:`,`=`), and how they work.
In the `Example` property, you can add a few examples using the new filters.

zvonler:
Added some docs, and in doing noticed that two index fields weren't searchable yet so I added them to the set. I hope the command documentation is appropriate, I tried to keep it as short as possible but answer the questions users would have.

alessio-perugini (approved):
LGTM. Outstanding work! :star_struck:

PR #2417, opened 2023-11-13
Make yaml output consistent with the json one

alessio-perugini:

      before creating one)
- [ ] The PR follows

Instead of using go struct to produce YAML output, we're now generating it starting from the JSON result using `sigs.k8s.io/yaml`.
pros:
- We don't have to implement custom Marshaling for YAML object
- We are guaranteed that the same JSON output matches 1:1 to YAML one
- We don't risk mismatching key names between JSON and YAML.
- We don't have to maintain 2 different output formats in many places.

cons:
- slightly less performant as we're doing a double marshaling

Currently, some values in the JSON output are missing when using the YAML format.

## What is the new behavior?

The yaml output is now consistent with the JSON one.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -45,6 +45,7 @@ require (
 	google.golang.org/grpc v1.59.0
 	google.golang.org/protobuf v1.31.0
 	gopkg.in/yaml.v3 v3.0.1
+	sigs.k8s.io/yaml v1.4.0
 )
 
 require (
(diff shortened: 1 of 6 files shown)


go.mod
```diff
@@ -45,6 +45,7 @@ require (
 	google.golang.org/grpc v1.59.0
 	google.golang.org/protobuf v1.31.0
 	gopkg.in/yaml.v3 v3.0.1
```
umbynos, line 47: Is this still used somewhere?

umbynos (approved):
Wow I didn't know this existed! I like the approach to the problem

go.mod
alessio-perugini: yep, in the profiles part.

PR #2515, opened 2024-01-29
Add diagnostics in the preprocessor

alessio-perugini:

      before creating one)
- [x] The PR follows

We're adding the diagnostics in the preprocessing part.

Currently, the diagnostics are considered only for the sketch compilation.

## What is the new behavior?

The diagnostics field will report errors from the preprocessor and the compiler.

## Other information

I'm proposing to use a struct that Stores the information of the diagnostics instead of using a callback approach. We can always revert that.


internal/arduino/builder/internal/preprocessor/ctags.go
```diff
@@ -54,38 +57,38 @@ func PreprocessSketchWithCtags(sketch *sketch.Sketch, buildPath *paths.Path, inc
...
-	verboseOutput.Write(gccStderr)
-	normalOutput.Write(gccStderr)
+	result, err := GCC(sourceFile, ctagsTarget, includes, buildProperties)
+	verboseOutput.Write(result.Stdout())
+	verboseOutput.Write(result.Stderr())
+	normalOutput.Write(result.Stderr())
 	if err != nil {
 		if !onlyUpdateCompilationDatabase {
-			return normalOutput.Bytes(), verboseOutput.Bytes(), err
+			return Result{args: result.Args(), stdout: verboseOutput.Bytes(), stderr: normalOutput.Bytes()}, err
```
cmaglie, line 66: I didn't understand this passage previously these functions returned `verboseOutput` and `normalOutput` where:
- `verboseOuput` had GCC stdout+stderr
- `normalOuput` had GCC stderr

now we have `stdout` and `stderr` where:
- `stdout` has GCC stdout+stderr
- `stderr` has GCC stderr

that works, but I don't like the improper assignment.

Maybe one solution could be to pass the verbose flag to the function, so:
1. we don't have to keep "verbose" and "non-verbose" output separated
2. we may return stdout and stderr unmixed

internal/arduino/builder/internal/preprocessor/result.go
```diff
@@ -0,0 +1,19 @@
+package preprocessor
```
cmaglie, line 1: missing header && comments

cmaglie (approved):

PR #2564, opened 2024-03-10
Add "permission denied" FAQ entry

wilyJ80:
Given the 50+ results for this "permission denied" error when uploading a sketch to the board in the forum, this entry was added to the FAQ. This might be desirable to have in an easily accessible page given it's a necessary step to take to get started with the microcontroller.

      before creating one)
- [ ] The PR follows

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -41,6 +41,14 @@ when using a serial port, the monitor baud rate can be set to 4800 with the foll
 
 `$ arduino-cli monitor -p <port> --config baudrate=4800`
 
+## "Permission denied" error in sketch upload
+
+This can be solved by setting up serial port permission. This can be achieved through the following command:
+
+`$ sudo usermod -a -G dialout <username>`
+
+Replace `<username>` with your user name. Logging out and logging in again is necessary for the changes to take effect.
+
 ## Additional assistance
 
 If your question wasn't answered, feel free to ask on [Arduino CLI's forum board][1].
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -43,12 +43,12 @@ when using a serial port, the monitor baud rate can be set to 4800 with the foll
 
 ## "Permission denied" error in sketch upload
 
-This problem might happen on some Linux systems, and can be solved by setting up serial port permissions. First, search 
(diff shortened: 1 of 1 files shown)


matthijskooijman:
Looks good to document this. I would suggest:
 - Making it a bit clearer that this works on Linux
 - Clarify that this command works on *most* Linux distributions, but the exact command (group name) might be different on differnet distros (maybe also add an `ls -l /dev/ttyACM0` command or so to help people figure out the command they need.
 - Maybe adding a command `id | grep dialout` or `cat /dev/ttyACM0` they can use to verify if this is actually the problem and this fix is appropriate.

I realize that adding all these things might result in hairy and hard-to-read FAQ entry, so maybe not everything should be added (or maybe you can find a wording and structure that balances readability and completeness...).

wilyJ80:
Pushed some changes. Tried to keep things concise but still reliable as suggested. Thanks for the suggestions, I'm learning more and more about the software side of Arduino.

alessio-perugini:
@wilyJ80 Thank you for opening this. I see that it was committed the compiled binary in https://github.com/arduino/arduino-cli/pull/2564/commits/b41effc1ffb637334b865012bd4db10ffe1e1cfa do you mind removing it?

wilyJ80:
Removed the binary directory. Is it OK now? Thank you for the patience. I'm starting out with contributing with this community. :smile:

alessio-perugini (approved):

alessio-perugini:
@wilyJ80 Yup, thank you for bearing with us! :smile:

matthijskooijman:
Hm, I'm a bit late to the party, but the documentation now just says "your group name", without any reference to how to figure out this group name, which is probably not very actionable for most users. I would suggest to use `dialout` in the example commandline and then say something like "The group name must be the owning group of the serial port (e.g. `/dev/ttyACM0`), on most Linux distributions it is `dialout`"

alessio-perugini:
@matthijskooijman Oops, I missed that it was later changed into `<group>` somehow I managed to still read it with `dialout`. 
@wilyJ80 if you don't mind I'd open a quick PR adding the suggestion of @matthijskooijman .
Sorry about that :pray: , I guess I have to change my :eyeglasses:

PR #2619, opened 2024-05-29
install.sh to work with wget

MilosSubotic:
I run install.sh on freshly installed Ubuntu so did not have curl installed as most people do, so it fallback to wget, but there was a bug in getFile(). Here is correction.


cmaglie:
@MilosSubotic thanks for your submission, I tested it and it works.

Before I can merge it I need to ask you to sign the CLA.

cmaglie (approved):

PR #2730, opened 2024-10-12
Set installer tmp path to env var

lucasjrt:

      before creating one)
- [X] The PR follows

This is an improvement in the `install.sh` script to support more platforms. For example, Termux is an Android terminal emulator which doesn't follow conventional UNIX file system, which means `/tmp` folder does not exist. Although, `$TMPDIR` is a variable that can be used to reference the `/tmp` equivalent in Termux.

The current behavior of the `install.sh` script is that it crashes when running from Termux, as the installer tries to write the installation files in a folder that doesn't exist (`/tmp`).

## What is the new behavior?

A new variable `$TEMPDIR` is defined based on some default variables that holds those paths, and if none of those are found, it defaults to previous `/tmp`.

No

## Other information

Android environments have the potential to be a development environment, so it would be nice that tools like arduino-cli could support it.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -10,6 +10,7 @@ PROJECT_NAME="arduino-cli"
 # BINDIR represents the local bin location, defaults to ./bin.
 EFFECTIVE_BINDIR=""
 DEFAULT_BINDIR="$PWD/bin"
+TEMPDIR="${TMPDIR:-${TEMP:-${TMP:-/tmp}}}"
 
 fail() {
   echo "$1"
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -10,6 +10,7 @@ PROJECT_NAME="arduino-cli"
 # BINDIR represents the local bin location, defaults to ./bin.
 EFFECTIVE_BINDIR=""
 DEFAULT_BINDIR="$PWD/bin"
+TEMPDIR="${TMPDIR:-${TEMP:-${TMP:-/tmp}}}"
```
alessio-perugini, line 13:
```suggestion
TMPDIR="${TMPDIR:-${TEMP:-${TMP:-/tmp}}}"
```
You're declaring the variable as `TEMPDIR` but using it as `TMPDIR`, without the letter `e`.

This breaks the script, I'll let you decide if you prefer `TEMPDIR` or `TMPDIR`. 
Using the correct variable name, your feature works as expected.

Looking forward to the quick patch. Thank you for your contribution! :nerd_face:

alessio-perugini (changes requested):

install.sh
lucasjrt: Hello, thank you for noticing that! My system uses the TMPDIR variable, so it worked even with the wrong name, sorry. I've just amended the commit, so it should be fine now.

alessio-perugini (approved):

PR #2744, opened 2024-11-07
Avoid `signature expired` error on index update, if the clock time is not set correctly.

cmaglie:

      before creating one)
- [X] The PR follows

Some users reported spurious "expired signature" errors. After some investigation, we found that all of them had the wrong system date set on their machine, with a date set in the past. Even if the error says that the signature is "expired", it's a signature that is not yet valid (it will be in the future).
Since we cannot trust the system clock, we recheck the signature with a future date so we may avoid displaying a difficult-to-understand error to the user.

This PR will suppress the, apparently unrelated, `signature expired` error in such cases.

An index update will trigger a `signature expired` error if the clock is set to a past date.

## What is the new behavior?

If the clock is set to a past date, an index update will no longer trigger a `signature expired` error.

No


cmaglie:
Superseded by #2750

PR #2788, opened 2024-12-17
compile: fix export when artifacts contains a folder

alessio-perugini:

      before creating one)
- [x] The PR follows

Fixes the export-dir, when it tries to copy a folder it crashes.

The copy mechanism fails to copy the dir because it's treated as a file, and the compile exits with error code 1

## What is the new behavior?

We filter out the directories and copy only exported binaries files.

### DIFF
--- a/internal/integrationtest/compile_1/compile_test.go
+++ b/internal/integrationtest/compile_1/compile_test.go
@@ -1321,3 +1321,46 @@ func buildWithCustomBuildPathAndOUtputDirFlag(t *testing.T, env *integrationtest
 		require.NotEmpty(t, content)
 	}
 }
+
+func TestCompileWithOutputDirFlagIgnoringFoldersCreatedByEsp32RainMakerLib(t *testing.T) {
+	env, cli := integrationtest.CreateArduinoCLIWithEnvironment(t)
+	defer env.CleanUp()
+
+	_, _, err := cli.Run("update")
+	require.NoError(t, err)
+
+	// Update index with esp32 core and install it
+	url := "https://raw.githubusercontent.com/espressif/arduino-esp32/gh-pages/package_esp32_index.json"
+	_, _, err = cli.Run("core", "update-index", "--additional-urls="+url)
+	require.NoError(t, err)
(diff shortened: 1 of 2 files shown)


internal/integrationtest/compile_1/compile_test.go
```diff
@@ -1321,3 +1321,46 @@ func buildWithCustomBuildPathAndOUtputDirFlag(t *testing.T, env *integrationtest
...
+	// When importing the `RMaker` library, the esp32 core produces in output a folder,
+	// containing all the binaries exported. The name of the folder matches the sketch name.
+	// In out case:
+	// cache-dir/
+	//  |- RainMakerSketch.ino/ <--- This should be ignored
+	//  |- RainMakerSketch.ino.uf2
+	//  |- RainMakerSketch.ino.bin
+	//  |- ...
+	err = sketchPath.Join(sketchName + ".ino").WriteFile([]byte(`
+		#include "RMaker.h"
```
alessio-perugini, line 1353: I was wondering if it would be better if we simulate the outcome of this test without relying on this library.

If we manually create the `RainMakerSketch.ino` folder inside the build path, it would be more future-proof. We don't risk breaking this test in case espressif changes something to that library. (Without downloading new core the test would be faster)

@cmaglie any preference on this?

cmaglie: > If we manually create the RainMakerSketch.ino folder inside the build path, it would be more future-proof.

The problem is that you must create the folder after the compile but before the export and those actions are quickly performed one after the other on the same "compile" command.
Maybe a better way to test is to extract an `exportSketchFiles` function from the compile command and unit-test it.

BTW I'm fine also with this test.

cmaglie (approved):

internal/integrationtest/compile_1/compile_test.go
alessio-perugini: @cmaglie since we're using the build path we have 2 options:
- precreate the folder -> compile -> assertions
- compile -> create the folder -> compile -> assertions

PR #2789, opened 2024-12-18
Add metadata retrieved from the `context` to the user agent when a new HTTP client is created

MatteoPologruto:

      before creating one)
- [x] The PR follows

Bug fix

The user agent does not specify if the CLI runs in daemon mode. It is always something like `arduino-cli/git-snapshot (amd64; windows; go1.23.2) Commit:84fc413a`.

## What is the new behavior?
The user agent obtained from the context metadata is propagated to the`network.user_agent_ext` if it's empty and the CLI runs in daemon mode. The complete user agent is `arduino-cli/git-snapshot arduino-ide/2.3.4 grpc-node-js/1.9.5 daemon (amd64; windows; go1.23.4) Commit:f3dc127e`.

### DIFF
--- a/internal/integrationtest/daemon/daemon_test.go
+++ b/internal/integrationtest/daemon/daemon_test.go
@@ -555,6 +555,34 @@ func TestDaemonCoreUpgradePlatform(t *testing.T) {
 	})
 }
 
+func TestDaemonUserAgent(t *testing.T) {
+	env, cli := integrationtest.CreateEnvForDaemon(t)
+	defer env.CleanUp()
+
+	// Set up an http server to serve our custom index file
+	// The user-agent is tested inside the HTTPServeFile function
+	test_index := paths.New("..", "testdata", "test_index.json")
+	url := env.HTTPServeFile(8000, test_index, true)
+
+	grpcInst := cli.Create()
+	require.NoError(t, grpcInst.Init("", "", func(ir *commands.InitResponse) {
+		fmt.Printf("INIT> %v\n", ir.GetMessage())
+	}))
+
+	// Set extra indexes
+	err := cli.SetValue("board_manager.additional_urls", `["http://127.0.0.1:8000/test_index.json"]`)
+	require.NoError(t, err)
+
+	{
+		cl, err := grpcInst.UpdateIndex(context.Background(), false)
+		require.NoError(t, err)
+		res, err := analyzeUpdateIndexClient(t, cl)
+		require.NoError(t, err)
+		require.Len(t, res, 2)
+		require.True(t, res[url.String()].GetSuccess())
+	}
+}
+
 func analyzeUpdateIndexClient(t *testing.T, cl commands.ArduinoCoreService_UpdateIndexClient) (map[string]*commands.DownloadProgressEnd, error) {
 	analyzer := NewDownloadProgressAnalyzer(t)
 	for {
--- a/internal/integrationtest/daemon/daemon_test.go
+++ b/internal/integrationtest/daemon/daemon_test.go
@@ -20,6 +20,10 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"maps"
+	"net/http"
+	"net/http/httptest"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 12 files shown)


internal/cli/daemon/daemon.go
```diff
@@ -66,6 +66,17 @@ func NewCommand(srv rpc.ArduinoCoreServiceServer, settings *rpc.Configuration) *
 			if maxGRPCRecvMsgSize < 1024 {
 				feedback.Fatal(i18n.Tr("%s must be >= 1024", "--max-grpc-recv-message-size"), feedback.ErrBadArgument)
 			}
+
+			// The user agent should include "daemon" for analytics purposes
+			_, err := srv.SettingsSetValue(cmd.Context(), &rpc.SettingsSetValueRequest{
+				Key:          "network.user_agent_ext",
+				ValueFormat:  "cli",
+				EncodedValue: "daemon",
+			})
```
alessio-perugini, line 75: What should we do in case this value is not empty? 
1. Should we always override with the `daemon` value
2. should we append to the previous value
3. should we honor the user setting?

/cc @cmaglie

alessio-perugini (commented):
Left some comments/
It would be cool if we're able to test it.

Maybe we could use a proxy HTTP, and inspect the call that arrives are with the expected User agent. Something similar to: https://github.com/arduino/arduino-cli/blob/master/internal/cli/configuration/network_test.go#L29-L51

internal/cli/daemon/daemon.go
MatteoPologruto: I think we should go with the third option and honor the user setting. If it's okay with you, I'm going to make this change.

alessio-perugini: @MatteoPologruto I agree. Let's ping also @cmaglie and @Xayton as I don't know if they might have better insight

internal/cli/daemon/daemon.go
```diff
@@ -58,14 +58,25 @@ func NewCommand(srv rpc.ArduinoCoreServiceServer, settings *rpc.Configuration) *
 				})
 				if err != nil {
 					// Should never happen...
-					panic("Failed to set default value for directories.builtin.libraries: " + err.Error())
+					feedback.Fatal(i18n.Tr("Cannot set the configuration key directories.builtin.libraries: %v", err), feedback.ErrGeneric)
```
cmaglie, line 61: Why this change? it looks unrelated and adds a string that should never happen to the translations.

internal/cli/daemon/daemon.go
```diff
@@ -58,14 +58,25 @@ func NewCommand(srv rpc.ArduinoCoreServiceServer, settings *rpc.Configuration) *
...
 				}
 			}
 
 			// Validate the maxGRPCRecvMsgSize flag
 			if maxGRPCRecvMsgSize < 1024 {
 				feedback.Fatal(i18n.Tr("%s must be >= 1024", "--max-grpc-recv-message-size"), feedback.ErrBadArgument)
 			}
+
+			// The user agent should include "daemon" for analytics purposes
+			_, err := srv.SettingsSetValue(cmd.Context(), &rpc.SettingsSetValueRequest{
```
cmaglie, line 71: > The user agent should include "daemon" for analytics purposes

Actually no, it should infer the user agent from the gRPC client:

https://github.com/arduino/arduino-cli/blob/6d9c930c79b6e8609915da0b774680588c5cc363/commands/instances.go#L67-L70

I think the Arduino IDE is passing something like `arduino-ide/x.y.z` into this string.
We should understand why this string is not propagated.

cmaglie (changes requested):

cmaglie (approved):

MatteoPologruto:
1hsiii

@cmaglie

internal/integrationtest/daemon/daemon_test.go
```diff
@@ -555,6 +555,34 @@ func TestDaemonCoreUpgradePlatform(t *testing.T) {
...
+	{
+		cl, err := grpcInst.UpdateIndex(context.Background(), false)
+		require.NoError(t, err)
+		res, err := analyzeUpdateIndexClient(t, cl)
+		require.NoError(t, err)
+		require.Len(t, res, 2)
+		require.True(t, res[url.String()].GetSuccess())
+	}
+}
+
```
alessio-perugini, line 585: Here, if possible would be nice to extract the assertion code that is done by passing the `true` flag in the `HTTPServerFile`. Or make an abstraction that passes some assertion functions.

Another possible idea is by "proxing" the request to the file server:

```patch
diff --git a/internal/integrationtest/daemon/daemon_test.go b/internal/integrationtest/daemon/daemon_test.go
index 7d800c84..982ba8bb 100644
--- a/internal/integrationtest/daemon/daemon_test.go
+++ b/internal/integrationtest/daemon/daemon_test.go
@@ -20,6 +20,10 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"maps"
+	"net/http"
+	"net/http/httptest"
+	"strings"
 	"testing"
 	"time"
 
@@ -562,15 +566,41 @@ func TestDaemonUserAgent(t *testing.T) {
 	// Set up an http server to serve our custom index file
 	// The user-agent is tested inside the HTTPServeFile function
 	test_index := paths.New("..", "testdata", "test_index.json")
-	url := env.HTTPServeFile(8000, test_index, true)
+	url := env.HTTPServeFile(8000, test_index)
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// Test that the user-agent contains metadata from the context when the CLI is in daemon mode
+		userAgent := r.Header.Get("User-Agent")
+
+		require.Contains(t, userAgent, "cli-test/0.0.0")
+		require.Contains(t, userAgent, "grpc-go")
+		// Depends on how we built the client we may have git-snapshot or 0.0.0-git in dev releases
+		require.Condition(t, func() (success bool) {
+			return strings.Contains(userAgent, "arduino-cli/git-snapshot") ||
+				strings.Contains(userAgent, "arduino-cli/0.0.0-git")
+		})
+
+		proxiedReq, err := http.NewRequest(r.Method, url.String(), r.Body)
+		require.NoError(t, err)
+		maps.Copy(proxiedReq.Header, r.Header)
+
+		proxiedResp, err := http.DefaultTransport.RoundTrip(proxiedReq)
+		require.NoError(t, err)
+		defer proxiedResp.Body.Close()
+
+		// Copy the headers from the proxy response to the original response
+		maps.Copy(r.Header, proxiedReq.Header)
+		w.WriteHeader(proxiedResp.StatusCode)
+		io.Copy(w, proxiedResp.Body)
+	}))
+	defer ts.Close()
 
 	grpcInst := cli.Create()
 	require.NoError(t, grpcInst.Init("", "", func(ir *commands.InitResponse) {
 		fmt.Printf("INIT> %v\n", ir.GetMessage())
 	}))
 
-	// Set extra indexes
-	err := cli.SetValue("board_manager.additional_urls", `["http://127.0.0.1:8000/test_index.json"]`)
+	additionalURL := ts.URL + "/test_index.json"
+	err := cli.SetValue("board_manager.additional_urls", fmt.Sprintf(`["%s"]`, additionalURL))
 	require.NoError(t, err)
 
 	{
@@ -579,7 +609,7 @@ func TestDaemonUserAgent(t *testing.T) {
 		res, err := analyzeUpdateIndexClient(t, cl)
 		require.NoError(t, err)
 		require.Len(t, res, 2)
-		require.True(t, res[url.String()].GetSuccess())
+		require.True(t, res[additionalURL].GetSuccess())
 	}
 }
 
```

alessio-perugini (approved):
I don't want to block this over some minor refactoring in the integration test part.

If possible I would extract the hardcoded assertion logic in the HTTPServeFile.

The implementation is LGTM. I've also tested locally with the usage of network.proxy and without and it correctly sends the UserAgent

PR #2792, opened 2024-12-19
Upgrade some golang packages

cmaglie:

      before creating one)
- [X] The PR follows

Upgrade the following packages:
* `golang.org/x/*`
* `go.bug.st/serial`
* `github.com/go-git/go-git/v5`

Upgrading go-git required a change in the git-URL parsing function (actually it has been completely reworked).


internal/arduino/libraries/librariesmanager/install.go
```diff
@@ -249,25 +249,44 @@ func (lmi *Installer) InstallGitLib(gitURL string, overwrite bool) error {
...
-	} else if parsed, err := url.Parse(gitURL); parsed.String() != "" && err == nil {
-		i := strings.LastIndex(parsed.Path, "/")
-		res = strings.TrimSuffix(parsed.Path[i+1:], ".git")
-		rev = plumbing.Revision(parsed.Fragment)
-	} else {
-		return "", "", errors.New(i18n.Tr("invalid git url"))
-	}
-	return res, rev, nil
+		argURL = "https://github.com/" + strings.TrimPrefix(argURL, "git@github.com:")
+	}
```
alessio-perugini, line 264: Any reason why we are restricting only to `github` provider?
Others allow us to do the same thing like:
- GitLab (https://gitlab.com/painlessMesh/painlessMesh)
- bitbucket (https://bitbucket.org/jezhill/signalacquisition/src/main/)

Looking at the git docs under `The SSH Protocol` section, it should be a git standard.
https://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols.

cmaglie: I've added GitLab and bitbucket too.

By looking at the documentation:

> The SSH Protocol
> A common transport protocol for Git when self-hosting is over SSH. This is because SSH access to servers is already set up in most places — and if it isn’t, it’s easy to do. SSH is also an authenticated network protocol and, because it’s ubiquitous, it’s generally easy to set up and use.
>
> To clone a Git repository over SSH, you can specify an ssh:// URL like this:
>
> ```
> $ git clone ssh://[user@]server/project.git
> ```
> Or you can use the shorter scp-like syntax for the SSH protocol:
>
> ```
> $ git clone [user@]server:project.git
> ```
> In both cases above, if you don’t specify the optional username, Git assumes the user you’re currently logged in as.

It does say that we can use SSH but it does not mandate that the username must be `git@hostname` and neither that the HTTPS URL should match the address, I think we cannot generalize these exceptions.

alessio-perugini (approved):

PR #2951, opened 2025-07-11
Auto-import libraries based on sketch profile.

cmaglie:

      before creating one)
- [X] The PR follows

This PR changes the way libraries are imported when using sketch profiles.

The libraries specified on profiles are "loaded" in the global state. Still, they are imported and compiled only if referenced by the sketch (or indirectly by a library imported from the sketch) after the so-called "library discovery".

## What is the new behavior?

The libraries specified on profiles are always imported and compiled, even if not referenced directly or indirectly by the sketch.
This reduces the number of "library discovery" passes and slightly improves compilation times. Still, the main part of the library discovery phase remains.

It might introduce a breaking change. For example, if a library is listed in a sketch profile but not used in the sketch, after this update, the library will still be compiled, leading to a compilation failure.


alessio-perugini (approved):
LGTM

KurtE:
Quick note, yes this does break things.  That is some of Adafruit libraries say they depend on other libraries, not that their library uses them, but their examples do.
https://forum.arduino.cc/t/sketch-using-adafruit-ft6206-h-and-adafruit-st77xx-library-builds-with-ide-fails-with-app-lab/1426553?u=kurte

PR #2954, opened 2025-07-16
linker: allow multi-step link recipe (`c.combine`)

pillo79:
This PR introduces the possibility of using multiple commands for the `c.combine` pattern, and infrastructure that will enable using the same logic on any build step, if so decided. 

Two changes in behavior are introduced:

* When looking for multi-step recipes (e.g. hooks) in `RunRecipe()`, a key that exactly matches `prefix.suffix` is explicitly ignored if numbered entries are found, and only used as a fallback when no other entries are present. 
This change is consistent with current docs that require a `prefix.NNN.suffix` syntax for multi-step recipes. A property named `prefix.suffix` is undefined behavior, and is currently included among the numbered ones.

* Reuse the `RunRecipe` logic for the `recipe.c.combine` step to allow multi-stage linking. The above change in this context means that numbered steps have priority over the single property, but both can be specified to maintain backwards compatibility (the new numbered properties are ignored by the current IDE).

NO TESTS as I don't know where to add the test case... :innocent: BUT I tried with a core that defined these rules and got the exact behavior I described here. If someone can give me pointers I can add the proper test case.

Feel free to cherry-pick and/or edit and/or shoot down at will :sweat_smile:

### DIFF
--- a/docs/platform-specification.md
+++ b/docs/platform-specification.md
@@ -302,6 +302,10 @@ compiler.libraries.ldflags=
 recipe.c.combine.pattern="{compiler.path}{compiler.c.elf.cmd}" {compiler.c.elf.flags} -mmcu={build.mcu} -o "{build.path}/{build.project_name}.elf" {object_files} {compiler.libraries.ldflags} "{archive_file_path}" "-L{build.path}" -lm
 ```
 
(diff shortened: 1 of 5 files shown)


pillo79:
@cmaglie any update on when this will be reviewed and/or merged?

cmaglie (approved):

PR #2990, opened 2025-08-29
Cherry Pick to 0.29.x: Disable DTR clearing on 1200-bps touch (only on Windows) (#2234)

fabik111:
Cherry pick this fix for the version 0.29.x. This version must be updated since it's used in the `arduino-cloud-cli`
### Original PR
The reason why it was originally introduced:
https://github.com/arduino/Arduino/pull/2709/commits/a6909bdb49d99253b4e684365e72e5dce31a49a7

Why we are removing it now?
* Windows does preserve the state of the RTS/DTR bits on successive opening of the serial port.
* The serial library used in the Arduino IDE 1.8.x has a bug when trying to set DTR=false, on successive opening of the port the DTR line is set back high by the USB serial driver. This works differently from the serial library we use in the Arduino CLI, that sets DTR=false for good and this change is preserved on the successive opening of the port.
* Having the serial port left in a state with DTR=false may cause problems to tools uploading later.

It may probably completely removed, but for now, to reduce the testing surface, it will be disabled only for Windows.

      before creating one)
- [ ] The PR follows

### DIFF
--- a/arduino/serialutils/serialutils.go
+++ b/arduino/serialutils/serialutils.go
@@ -17,6 +17,7 @@ package serialutils
 
 import (
 	"fmt"
+	"runtime"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


cmaglie (approved):
Ignored CI, it will never pass on such an old commit.

Issue #2133, opened 2023-03-29
Support for custom libraries when using `sketch.yaml`


dmohns:
### Describe the request

In a project I am currently using different "kind" of libraries. 

1. Versioned libraries available in via Arduino Library Manager 
2. Non-versioned libraries available via their source code on Github
3. Non-versioned libraries available via their source code in local subfolder folder of my project (zipped or raw)

To ease collaboration on the project I would like to use a [Project Sketch File](https://arduino.github.io/arduino-cli/0.31/sketch-project-file/) and provide a set of dependencies that are confirmed working.

Unfortunately, the Project Sketch File only seem to support 1. (versioned libraries available in Arduino Library Manager). Scope of this request is to add functionality to supply Github and local folder libraries to the `libraries:` attribute on `sketch.yam` (in some fashion).

### Describe the current behavior

Currently, I have been able to compile my project using

```
arduino-cli compile --fqbn "<FQBN>" --libraries=Library/
```

However, if I tried to use `--profile` and `--libraries` the compile command fails

```
You cannot use the --libraries flag while compiling with a profile.
```

### Arduino CLI version

arduino-cli  Version: nightly-20230329 Commit: b69ed33 Date: 2023-03-29T01:34:59Z

### Operating system

macOS

### Operating system version

12.6.3

### Issue checklist

- [X] I searched for previous requests in [the issue tracker](https://github.com/arduino/arduino-cli/issues?q=)
- [X] I verified the feature was still missing when using the [nightly build](https://arduino.github.io/arduino-cli/dev/installation/#nightly-builds)
- [X] My request contains all necessary details

umbynos:
Seems a duplicate of https://github.com/arduino/arduino-cli/issues/108

jirutka:
> Seems a duplicate of #108

It’s not a duplicate, #108 is just one of the ways it can be implemented (and more complicated than requested here).

dmohns:
Yeah, just to clarify, this issue is about making `sketch.yaml` spec more flexible to support libraries from other sources than Arduino Library Manager.

One way that is backwards compatible would be

```
profiles:
  <PROFILE_NAME>:
    ....
    libraries:
      - <LIB_NAME> (<LIB_VERSION>)
```

to allow `<LIB_VERSION>` to be `<LIB_VERSION_OR_URL>` and take the form of VCS and local urls, for example

```
profiles:
  my_profile:
    ....
    libraries:
      - ArduinoIoTCloud (1.0.2)
[...]

PR #3028, opened 2025-10-21
Support for git-downloaded libraries in profiles.

cmaglie:

      before creating one)
- [X] The PR follows

This PR adds support for git-based libraries in profiles.

Sketch profiles do not support libraries downloaded from git.
Only libraries available on the Libraries Index or in the local storage could be included in a build made with profiles.

## What is the new behavior?

Now the profiles support a third kind of library, and a git reference could be indicated in the profiles, like in the following example:

```yaml
profiles:
  zero:
    fqbn: arduino:samd:arduino_zero_native
    platforms:
      - platform: arduino:samd (1.8.14)
    libraries:
      - git: https://github.com/cmaglie/FlashStorage.git#1.0.0

default_profile: zero
```

The specific git-revision of the library specified in the profile will be downloaded from git and cached in the data directory's `internal` folder. The downloaded copy is cached only the first time and will not be downloaded again.

If the git-reference is updated on the repository, the cached copy will become outdated, and the only way to refresh it is to delete the internal cache directory to force the Arduino CLI to download it again.

For example, a user trying to always utilize the latest `main` branch of a library might be tempted to write something like:
```yaml
    libraries:
      - git: https://github.com/cmaglie/FlashStorage.git#main
```
BTW, this won't work well because, after the first compile (and download from git), the Arduino CLI will never try again to refresh the library.

No

## Other information

Fix https://github.com/arduino/arduino-cli/issues/2133


internal/arduino/sketch/testdata/profiles/profile_with_libraries.yml
```diff
@@ -0,0 +1,13 @@
+profiles:
+  giga:
+    fqbn: arduino:mbed_giga:giga
+    platforms:
+      - platform: arduino:mbed_giga (4.3.1)
+    libraries:
+      - FlashStorage (1.2.3)
+      - dir: /path/to/system/lib
+      - dir: path/to/sketch/lib
+      - git: https://github.com/username/HelloWorld.git#v2.13
```
lucarin91: Why `#`? I think it is more common `@` in this case
```suggestion
      - git: https://github.com/username/HelloWorld.git@v2.13
```

cmaglie:
> We cannot guarantee that a tag hasn't moved and that someone else downloading that project will not have a completely different version of the library.

This is already happening with the unversioned platform reference in profiles https://github.com/arduino/arduino-cli/pull/2940.
In these cases, simply the profile will not guarantee a reproducible build.
Developers are requesting these features to make their lives easier; sometimes, they want to use a specific version of a library that is not included in the library index.

> If we want to go down this route, we should do mirroring of github repository ourselves, to make sure that the tag didn't change (exactly how go implement that).

Actually, Go is not mirroring; if you put a "moving" reference in the `go.mod` file, like a branch:
```
	github.com/arduino/go-paths-helper main
```
this is replaced by `go mod tidy` with a fixed tag:
```
	github.com/arduino/go-paths-helper v1.14.0
```
or if `main` points to a non-tagged revision, the timestamp+sha256sum of the revision is recorded:
```
	github.com/arduino/go-paths-helper v0.0.0-20171220113728-d1dd9e313b1b
```
(The SHA256 would be more than enough tho, since the integrity is validated with `go.sum` separately).

lucarin91:
> > If we want to go down this route, we should do mirroring of github repository ourselves, to make sure that the tag didn't change (exactly how go implement that).
> 
> Actually, Go is not mirroring; if you put a "moving" reference in the `go.mod` file, like a branch:
> ```
> 	github.com/arduino/go-paths-helper main
> ```
> this is replaced by `go mod tidy` with a fixed tag:
> ```
> 	github.com/arduino/go-paths-helper v1.14.0
> ```
> or if `main` points to a non-tagged revision, the timestamp+sha256sum of the revision is recorded:
> ```
> 	github.com/arduino/go-paths-helper v0.0.0-20171220113728-d1dd9e313b1b
> ```
> (The SHA256 would be more than enough tho, since the integrity is validated with `go.sum` separately).
> 

Check it here https://index.golang.org/. Go mirrors git libraries and it also serves a checksum database, to be sure that tags doesn't moves and you cannot update a tag breaking code of someone else. So as I said we could do this, but I would prefer to think that a bit more, to avoid this becoming a liability issue.

PR #3092, opened 2026-02-04
Add check for the presence of 'other' in licensed cache

cmaglie:

      before creating one)
- [X] The PR follows

This check is done before running the `licensed status`.
For a detailed explanation, see: https://github.com/arduino/tooling-project-assets/issues/683

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -65,12 +65,12 @@ tasks:
       - task: general:cache-dep-licenses
     cmds:
       - |
-        if grep --files-with-matches -r "license: other" .licenses/go; then \
-          echo "The libraries cache file above have the license set to 'other'. Please fix the cached license and retry."; \
-          exit 1; \
-        else \
-          licensed status; \
+        if grep --files-with-matches -r "license: other" .licenses; then
+          echo "The license cache files above have the license set to 'other'. Please manually define the license type and retry."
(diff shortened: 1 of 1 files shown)


Taskfile.yml
```diff
@@ -64,7 +64,13 @@ tasks:
...
       - task: general:cache-dep-licenses
     cmds:
-      - licensed status
+      - |
+        if grep --files-with-matches -r "license: other" .licenses/go; then \
+          echo "The libraries cache file above have the license set to 'other'. Please fix the cached license and retry."; \
+          exit 1; \
+        else \
+          licensed status; \
+        fi
```
per1234: There is no reason to make this conditional. It only adds complexity. Leave it as an element in the `cmds` sequence, as before.

Taskfile.yml
```diff
@@ -64,7 +64,13 @@ tasks:
     deps:
       - task: general:cache-dep-licenses
     cmds:
-      - licensed status
+      - |
+        if grep --files-with-matches -r "license: other" .licenses/go; then \
```
per1234:
```suggestion
        if grep --files-with-matches -r "license: other" .licenses; then \
```

The `.licenses/go` path is project-specific. The standard folder structure looks like `.licenses/<project identifier>/<source identifier>`, in order to support repositories that contain multiple modules (or the equivalent for other languages/frameworks). It will be better to make this code generally applicable so that it can be upstreamed to the assets repo verbatim.

per1234 (approved):
Thanks Cristian!

PR #3171, opened 2026-04-20
WIP: feat: implement pre_upload check and pattern

facchinm:

      before creating one)
- [ ] The PR follows

For multi binary uploads it might be useful to skip flashing a part of the binary (if unchanged, or just compatible). By adding this pair of rules (activated by .upload.use_pre_upload_check=true) we can give maximum flexibility for platform developers who need this kind of behaviour.

pre_upload_check.pattern should return != 0 if we want to execute the second rule before upload.pattern

### DIFF
--- a/commands/service_upload.go
+++ b/commands/service_upload.go
@@ -571,6 +571,8 @@ func (s *arduinoCoreServerImpl) runProgramAction(ctx context.Context, pme *packa
 		}
 	}
 
+	preUpload := uploadProperties.GetBoolean("upload.use_pre_upload_check")
+
 	// Run recipes for upload
 	toolEnv := pme.GetEnvVarsForSpawnedProcess()
 	if burnBootloader {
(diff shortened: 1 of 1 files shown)


commands/service_upload.go
```diff
@@ -571,6 +571,8 @@ func (s *arduinoCoreServerImpl) runProgramAction(ctx context.Context, pme *packa
 		}
 	}
 
+	preUpload := uploadProperties.GetBoolean("upload.use_pre_upload_check")
```
per1234, line 574: `upload.use_pre_upload_check` seems superfluous. Why not just use the existence of the pattern properties as the determining factor?

per1234 (changes requested):
Please add documentation for the new properties to the Arduino Platform Specification:

https://github.com/arduino/arduino-cli/blob/master/docs/platform-specification.md

commands/service_upload.go
facchinm: That could be an idea :slightly_smiling_face: I just eanted to keep theat part more isolated but it's indeed superfluous.

facchinm: I thought about this a bit and I think it might be better to keep some kind of board-depenedant flag (coming from boards.txt) otherwise the rule will apply to all the boards using the same upload tool (and it might be unwanted for some reason)

cmaglie:
Superseded by `scripting-tools` https://github.com/arduino/scripting-tools.

PR #3206, opened 2026-06-01
fix(compiler): Sketch may be wrongly considered up-to-date (-> not recompiled when needed)

### LINKED ISSUE #3202: Modified .ino is not recompiled, stale cached object file is reused
### Describe the problem

Arduino IDE reuses stale `.ino.cpp.o` after modifying sketch.

### To reproduce

Steps:
1. Compile sketch
2. Modify .ino
3. Add `#error TEST`
4. Save
5. Compile again

IDE reports:
```text
Using previously compiled file:
...Test.ino.cpp.o
```

and compilation succeeds using stale object files.

### Expected behavior

Sketch source code is recompiled after being modified.

In the demonstration described above, the expected result is for the compilation to fail due to the introduced `#error` directive.

### Arduino IDE version

2.3.9

### Operating system

Windows

### Operating system version

W11

### Additional context

Everything was fine before the update.

---

I'm using "**Arduino SAMD Boards (32-bits ARM Cortex-M0+)**" platform version 1.8.14.

---

Deleting the folder at the following path:

```text
C:\Users\<user>\AppData\Local\arduino\sketches
```

temporarily fixes the issue.

#### Additional reports

- https://forum.arduino.cc/t/2-3-9-recompile-errors/1446989
- https://forum.arduino.cc/t/errors-using-version-3-3-9/1447333/1
- https://forum.arduino.cc/t/errors-using-version-3-3-9/1447333/3
- https://github.com/arduino/arduino-cli/issues/3204
- https://github.com/arduino/arduino-ide/issues/2921

### Issue checklist

- [x] I searched for previous reports in [the issue tracker](https://github.com/arduino/arduino-ide/issues?q=)
- [x] I verified the problem still occurs when using the latest [nightly build](https://www.arduino.cc/en/software#nightly-builds)
- [x] My report contains all necessary details

cmaglie:

      before creating one)
- [X] The PR follows

Fixes a race condition in the builder.

The root cause is that [`sketchIsUnchanged` is computed](https://github.com/arduino/arduino-cli/blob/fb6bae107847ea597e13622d2f656006161a1c4b/internal/arduino/builder/internal/detector/detector.go#L336) after the [`preRunner` goroutines are started](https://github.com/arduino/arduino-cli/blob/fb6bae107847ea597e13622d2f656006161a1c4b/internal/arduino/builder/internal/detector/detector.go#L308-L322). A `preRunner` GCC task for mergedSketch would write `libsdetect.d` with a fresh timestamp, making the subsequent check incorrectly report "sketch unchanged."

Actually, the race condition will not be easily triggered because the prerunner will usually be scheduled *after* the check, since Go tends to keep the current thread running rather than immediately starting just-spawned goroutines. The race condition becomes more likely if the sketch has many libraries and the PC has a multi-core processor (increasing the likelihood that a concurrent GCC is started quickly).

The fix is relatively simple: compute `sketchIsUnchanged` before `preRunner` is started.

In some circumstances, likely with complex sketches, the sketch may not be recompiled after being modified.

## What is the new behavior?

The sketch is always recompiled if a modification occurs.

No

## Other information

Should fix #3202

### DIFF
--- a/internal/arduino/builder/internal/detector/detector.go
+++ b/internal/arduino/builder/internal/detector/detector.go
@@ -22,6 +22,7 @@ import (
 	"errors"
 	"fmt"
 	"maps"
+	"os"
 	"os/exec"
 	"regexp"
 	"slices"
@@ -320,6 +321,15 @@ func (l *SketchLibrariesDetector) findIncludes(
 			l.preRunner.Cancel()
 		}
 	}()
+
+	// Only for testing purposes
+	if libraryDetector_ForcePrerunnerPriority {
+		// When set to true, it forces the library detector pre-runner (that is normally scheduled
+		// randomly by the goroutine scheduler) to run with priority, in particular before
+		// the library detection starts.
+		fmt.Println("TESTING: Waiting for pre-runner start...")
+		time.Sleep(time.Second)
+	}
 
 	l.addIncludeFolder(buildCorePath)
 	if buildVariantPath != nil {
@@ -732,3 +742,6 @@ func LibrariesLoader(
 	resolver := librariesresolver.NewCppResolver(allLibs, targetPlatform, buildPlatform)
 	return lm, resolver, verboseOut.Bytes(), nil
 }
+
+// libraryDetector_ForcePrerunnerPriority is used for testing purposes.
+var libraryDetector_ForcePrerunnerPriority = os.Getenv("TESTING_LIBRARY_DETECTOR_FORCE_PRERUNNER_PRIORITY") == "1"
--- a/internal/arduino/builder/internal/detector/detector.go
+++ b/internal/arduino/builder/internal/detector/detector.go
@@ -225,9 +225,9 @@ func (l *SketchLibrariesDetector) IncludeFoldersChanged() bool {
 	return l.detectedChangeInLibraries
 }
 
-// IsSketchUnchanged returns true if the sketch source and all included library
+// IsSketchDepsUnchanged returns true if the sketch or any of its dependencies is up-to-date
 // headers are unchanged since the last preprocessing run.
-func (l *SketchLibrariesDetector) IsSketchUnchanged() bool {
+func (l *SketchLibrariesDetector) IsSketchDepsUnchanged() bool {
 	return l.sketchIsUnchanged
 }
 
--- a/internal/arduino/builder/internal/detector/detector.go
+++ b/internal/arduino/builder/internal/detector/detector.go
@@ -225,7 +225,8 @@ func (l *SketchLibrariesDetector) IncludeFoldersChanged() bool {
 	return l.detectedChangeInLibraries
 }
 
-// IsSketchUnchanged returns true if the sketch or any of its dependencies is up-to-date
+// IsSketchUnchanged returns true if the sketch source and all included library
+// headers are unchanged since the last preprocessing run.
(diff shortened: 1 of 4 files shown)


cmaglie:
## How to test the fix:

1. Download the Arduino CLI build here: https://github.com/arduino/arduino-cli/actions/runs/26755498463 (it's at the bottom of the page; be sure to be **logged in to GitHub** to download the artifacts).
2. Unpack the artifacts; you should find an `arduino-cli.exe`.

### Testing in the Arduino IDE

If you want to test the fix in the Arduino IDE, you must replace `arduino-cli.exe`/`arduino-cli` bundled in the IDE with the version downloaded above:
1. Close the Arduino IDE.
2. Find your Arduino IDE install folder (on Windows, it's usually `%LOCALAPPDATA%/Programs/Arduino IDE` or `C:/Programs/Arduino IDE`)
3. Navigate into `IDE_FOLDER/resources/app/lib/backend/resources/`
4. You should find an `arduino-cli.exe`, replace it with the one downloaded from GitHub.
5. Restart the IDE.

internal/arduino/builder/internal/detector/detector.go
```diff
@@ -304,6 +306,16 @@ func (l *SketchLibrariesDetector) findIncludes(
...
 
+	// Determine if the sketch is unchanged BEFORE starting the preRunner goroutines.
+	// The preRunner may start GCC tasks that write libsdetect.d as a side effect; if we
+	// checked after the preRunner, a fast GCC run could update libsdetect.d's timestamp
+	// and make the sketch appear unchanged even after a modification (issue #3202).
+	mergedSketchForCheck, err := l.makeSourceFile(sketchBuildPath, sketchBuildPath, paths.New(sketch.MainFile.Base()+".cpp.merged"))
+	if err != nil {
+		return err
+	}
+	l.sketchIsUnchanged, _ = mergedSketchForCheck.ObjFileIsUpToDate(logrus.WithField("runner", "sketchcheck"))
```
lucarin91: I am not sure if this change will introduce some other side effects, because now the `sketchIsUnchanged` variable will also be set if the `useCachedLibrariesResolution` is false. `sketchIsUnchanged` is then used that https://github.com/arduino/arduino-cli/blob/325e1d5384b8779e7cd99ee4e0a85801bd2b41fc/internal/arduino/builder/preprocess_sketch.go#L32, which, as far as I understand it, will skip some compilation steps.

internal/arduino/builder/internal/detector/detector.go
```diff
@@ -333,7 +354,6 @@ func (l *SketchLibrariesDetector) findIncludes(
 		if err != nil {
 			return err
 		}
```
lucarin91: I think this could be removed now

internal/arduino/builder/internal/detector/detector.go
cmaglie: `useCachedLibrariesResolution` is used by the language server to do a quick rebuild of the sketch, skipping the library discovery phase.

This setting will make rebuilds faster, so the language server responds more quickly, but if the `#include <...>` directives are edited, library suggestions may be incomplete until the next full compile.

BTW, since `sketchIsUnchanged` is orthogonal to `useCachedLibrariesResolution`, IMHO the current patch is correct.

lucarin91 (approved):

lucarin91 (approved):

PR #3234, opened 2026-07-07
ci: improve test robustness

pillo79:

This PR addresses two flaky tests in the CI testing workflow: 
* `TestUploadCancelation`, which relies on a racy timer to detect failures
* `TestArduinoCliDaemonCompileWithLotsOfOutput` is unnecessarily tight (normal test case completes anyway, failure hangs indefinitely).

* `TestUploadCancelation` verifies the existence of a file which is created and deleted in a 5 second window, expecting a kill command to complete in that time frame; otherwise, assumes it was not sent. 
* `TestArduinoCliDaemonCompileWithLotsOfOutput` should detect issue #2169 by running several compile jobs. Each is expected to take less than 20 seconds; otherwise, assumes it is hung. 

## What is the new behavior?

* `TestUploadCancelation` watches the file for constant updates until a longer 30 second timeout expires - the timer is never reached because it stops as soon as the file is no longer updated (unless the kill command had no effect). 
* `TestArduinoCliDaemonCompileWithLotsOfOutput` now waits up to 120 seconds for each compile; again, the timeout is never reached unless the CLI actually hangs.

No


pillo79:
Currently outstanding issues:

* Several tests rely on `board list` discovering a mocked board, and this does not always succeed due to the issue fixed in arduino/pluggable-discovery-protocol-handler#71. Waiting on that being merged to mark this as ready for review.

* The flakiness in `markdown:check-links` cannot be fully fixed at the task level due to Github network issues. Multiple retries with a simple loop + delay are performed on the same runner and always result in the same links being reported as invalid.

cmaglie (approved):

PR #3248, opened 2026-07-22
builder: handle ANSI-colored include diagnostics

### LINKED ISSUE #3020: fdiagnostics-color in extra_flags causes break in library resolver
### Describe the problem

When `-fdiagnostics-color=always` is added as part of the build extra flags, the library resolver/processor is unable to resolve any libraries.

This is because, if I copy the output to the clipboard, the resolver seems to include the ANSI escape code in the library name:

Image

### To reproduce

with an example sketch such as
```
#include <Arduino.h>
#include <SPI.h>

void setup(void){
    Serial.begin(115200);
    Serial.println("HI!");
}

void loop(void){

}
```

run
```
arduino-cli compile -v \
	--build-property "compiler.c.extra_flags=-fdiagnostics-color=always" \
	--build-property "compiler.cpp.extra_flags=-fdiagnostics-color=always" \
	--fqbn ArduinoCore-avr:avr:uno \
	testArduinoCliBug.ino
```

### Expected behavior

The ANSI sequence should be escaped/removed from the library

### Arduino CLI version

r2843.20e315c2f

### Operating system

Linux

### Operating system version

Arch Linux

### Additional context

From initial code research the most likely candidate for the fault is in `internal/arduino/builder/internal/detector/detector.go`, possibly in the function `IncludesFinderWithRegExp` or the thing calling it.

### Issue checklist

- [x] I searched for previous reports in [the issue tracker](https://github.com/arduino/arduino-cli/issues?q=)
- [x] I verified the problem still occurs when using the [nightly build](https://arduino.github.io/arduino-cli/dev/installation/#nightly-builds)
- [x] My report contains all necessary details

giri256:

- [x] The PR has no duplicates.
- [x] The PR follows the contributing guidelines.
- [x] Tests for the changes have been added.
- [x] Documentation is not required because this fixes internal diagnostic parsing without changing the public interface.
- [x] `UPGRADING.md` is not required because there is no breaking change.
- [x] `configuration.schema.json` is not affected.

Bug fix for library dependency detection during compilation.

When compiler diagnostics contain ANSI CSI color sequences, the missing-header parser includes those escape sequences in the extracted header name. For example, `SPI.h` can be returned with embedded control bytes, so the library resolver cannot match the include.

This is reproducible when `-fdiagnostics-color=always` is supplied through build properties.

Fixes #3020.

## What is the new behavior?

ANSI CSI sequences are removed only from the detector's internal copy of compiler stderr before parsing the missing include. The original compiler output remains untouched, so user-visible colored diagnostics are preserved.

A regression test covers a realistic GCC-colored missing-header diagnostic.

## Does this PR introduce a breaking change, and is titled accordingly?

No.

## Other information

Local verification:

- `go test ./internal/arduino/builder/internal/detector -count=1`
- `go test ./internal/arduino/builder/... -count=1`
- `go vet ./internal/arduino/builder/internal/detector`
- `git diff --check`

The first upstream run passed all unit, style, lint, artifact, macOS, Linux, and other Windows jobs. Its only failure was the unrelated Windows board integration mock not returning its synthetic serial port in `TestBoardListMock/ContainsMockedPorts`.

### DIFF
--- a/internal/arduino/builder/internal/detector/detector.go
+++ b/internal/arduino/builder/internal/detector/detector.go
@@ -648,10 +648,14 @@ func (l *SketchLibrariesDetector) failIfImportedLibraryIsWrong() error {
 	return nil
 }
 
-var includeRegexp = regexp.MustCompile(`(?ms)^\s*[0-9 |]*\s*#[ \t]*include\s*[<"](\S+)[">]`)
+var (
+	ansiControlSequenceRegexp = regexp.MustCompile("\x1b\\[[0-?]*[ -/]*[@-~]")
+	includeRegexp             = regexp.MustCompile(`(?ms)^\s*[0-9 |]*\s*#[ \t]*include\s*[<"](\S+)[">]`)
+)
 
 // IncludesFinderWithRegExp fixdoc
 func IncludesFinderWithRegExp(source string) string {
+	source = ansiControlSequenceRegexp.ReplaceAllString(source, "")
 	match := includeRegexp.FindStringSubmatch(source)
 	if match != nil {
 		return strings.TrimSpace(match[1])
--- a/internal/arduino/builder/internal/detector/detector_test.go
+++ b/internal/arduino/builder/internal/detector/detector_test.go
@@ -85,3 +85,14 @@ func TestIncludesFinderWithRegExpPaddedIncludes5(t *testing.T) {
 
 	require.Equal(t, "Foobar.h", include)
 }
+
+func TestIncludesFinderWithRegExpANSIColoredOutput(t *testing.T) {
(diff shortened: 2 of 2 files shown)


giri256:
The only failed code check is `test-integration (windows-latest, board)`. The log shows `TestBoardListMock/ContainsMockedPorts` received an empty detected-port list from the mocked serial discovery. The equivalent macOS/Linux board shards and every unit, formatting, lint, artifact, and other Windows job passed, and this test does not exercise the changed detector package. I cannot rerun upstream Actions as an external contributor; could a maintainer rerun the failed job?

cmaglie (approved):
Thanks!

PR #3270, opened 2026-08-11
builder: add library discovery macros

### LINKED ISSUE #3124: add library detection and version macros
### Describe the request

It would be extremely useful if the CLI could automatically generate, after the library discovery phase,  a set of macros that describe the presence in the build of each detected library and its version.

There are several reasons:
* libraries could auto-configure depending on the presence of other libs in the project
* versions would not need to be manually defined in code for each library
* it would be possible to test for any library version in the same way

The CLI does not currently provide this information. Only the `{includes}` variable in the sketch/libraries compilation step may offer any clue.

### Arduino CLI version

new ones 

### Operating system

N/A

### Operating system version

OS independent

### Additional context

A suitable solution would provide, when for example Bridge 1.2.3 is included, something like a `-DFOUND_BRIDGE_LIB=0x010203` in a specific `{lib_detect_macros}` variable for expansion in build commands. It may also be as a `@file` if this would generate too many macros.

The identifiers would be generated with unequivocal rules (for example by converting the library name to a valid uppercase C slug) and the versions would need to be easily sortable.

Notes:
- it would be desirable to have this at core compile time as well, if possible. Not a deal breaker if the core can refer to libraries (#3110); in that case the library dependent changes may be moved to a core-supplied lib.
- libs whose version can't be identified could have `1` as their version, which translates to the lowest possible non-zero version `0.0.1`
- multiple #defines can be added for the same lib, for example:
   - longer ones, including the author name, to remove potential ambiguities;
   - more info on discovery (e.g. suffixes `_IN_SKETCHBOOK`, `_IN_CORE`, `_SCORE` etc)

### Issue checklist

- [x] I searched for previous requests in [the issue tracker](https://github.com/arduino/arduino-cli/issues?q=)
- [x] I verified the feature was still missing when using the [nightly build](https://arduino.github.io/arduino-cli/dev/installation/#nightly-builds)
- [x] My request contains all necessary details

pillo79:

      before creating one)
- [x] The PR follows

This PR adds a set of library discovery macros, addressing the need described in #3124.

The set of libraries used in the build are not directly inspectable at compile time. This can be worked around in fragile ways, for example with `__has_include` or checking an include guard, but it's very library-specific.

## What is the new behavior?

A new default property `{build.library_discovery_macros}` is populated by the CLI and available while building all sketch and library files. For each discovered library two macros are defined:
 * `FOUND_<LIB>_LIB` is set to the hex representation the version code (`0x<major><minor><patch>`  as 3 hex bytes, each capped to `0xff`).
 * `FOUND_<LIB>_LIB_IN_<SOURCE>` " details the source of the library. Only one of the following is defined and set to `1`:
    * `IN_SKETCHBOOK`: user's `Arduino/libraries` folder, 
    * `IN_PLATFORM`: used core's `libraries/` folder,
    *  `IN_PROFILE`: specifically added by the used build profile ,
    * `IN_SPECIFIED_PATH`: specifically added via `--library` command line option

No.

### DIFF
--- a/docs/platform-specification.md
+++ b/docs/platform-specification.md
@@ -146,6 +146,21 @@ The following automatically generated properties can be used globally in all con
   intent, `-DARDUINO_LIB_DISCOVERY_PHASE` was added to `recipe.preproc.macros` during library discovery in Arduino
   Builder 1.5.3/Arduino CLI 0.10.0. That flag was replaced by the more flexible `{build.library_discovery_phase}`
   property.
+- `{build.library_discovery_flags}`: a series of `-D` flags, one set per library resolved during library discovery. It
+  is available to sketch and library compile recipes only; it is deliberately not exposed to core-compile recipes. This
+  property was added in Arduino CLI 1.6.0.<br /> For each library, the property contains a flag with the form
+  `-DFOUND_<SLUG>_LIB=0x<hex-version>`, and one with the form `-DFOUND_<SLUG>_LIB_<SOURCE>=1`.
+  - `<SLUG>` is the library name converted to a valid uppercase C identifier (runs of characters outside `[A-Za-z0-9_]`
+    become a single `_`).
+  - `<hex-version>` is the library's version, in the form of a 24-bit value assembled from the major/minor/patch
+    components (`MMmmpp`). Each component is clamped to `0xFF`. Pre-release/build metadata components are discarded. An
+    unknown or unparseable version is encoded as `1`.
+  - `<SOURCE>` identifies where the library was found:
+    - `IN_IDE`: Arduino IDE's built-in libraries.
+    - `IN_PLATFORM`: libraries bundled with the platform.
+    - `IN_PROFILE`: libraries specified as a dependency by the build profile.
+    - `IN_SKETCHBOOK`: libraries located under `directories.user`, or under a path specified via a `--libraries` flag.
+    - `IN_SPECIFIED_PATH`: libraries under a path specified via a `--library` flag.
 - `{compiler.optimization_flags}`: see ["Sketch debugging configuration"](#sketch-debugging-configuration) for details
 - `{extra.time.utc}`: Unix time (seconds since 1970-01-01T00:00:00Z) according to the machine the build is running on
 - `{extra.time.local}`: Unix time with local timezone and DST offset
--- a/internal/arduino/builder/archive_compiled_files.go
+++ b/internal/arduino/builder/archive_compiled_files.go
@@ -56,6 +56,9 @@ func (b *Builder) archiveCompiledFiles(archiveFilePath *paths.Path, objectFilesT
 
 	for _, objectFile := range objectFilesToArchive {
 		properties := b.buildProperties.Clone()
+		if b.pathIsCore(archiveFilePath) {
+			properties.Remove("build.library_discovery_flags")
+		}
 		properties.Set("archive_file", archiveFilePath.Base())
 		properties.SetPath("archive_file_path", archiveFilePath)
 		properties.SetPath("object_file", objectFile)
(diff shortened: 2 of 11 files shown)


docs/platform-specification.md
```diff
@@ -146,6 +146,17 @@ The following automatically generated properties can be used globally in all con
...
+  discovery, of the form `-DFOUND_<SLUG>_LIB=0x<hex-version>` plus an optional `-DFOUND_<SLUG>_LIB_<SOURCE>=1` suffix.
+  `<SLUG>` is the library name converted to a valid uppercase C identifier (runs of characters outside `[A-Za-z0-9_]`
+  become a single `_`). The hex version packs major/minor/patch into a 24-bit value (`0xMMmmpp`, each component clamped
+  to `0xFF`); an unknown or unparseable version is encoded as `1`. `<SOURCE>` identifies where the library was found:
+  `IN_SKETCHBOOK` (installed libraries, and paths passed via `--libraries`), `IN_PLATFORM` (bundled with the target or
+  referenced core), `IN_PROFILE` (declared in a sketch profile), `IN_IDE` (bundled with the IDE), or `IN_SPECIFIED_PATH`
+  (a single library path passed via `--library`); no suffix macro is emitted for an unrecognized location. It is
+  available to sketch and library compile recipes only; it is deliberately not exposed to core-compile recipes, since
+  arduino-cli's incremental build cache has no mechanism to detect a change in its value and force a core rebuild. This
+  property was added in Arduino CLI 1.6.0.
```
per1234:
```suggestion
- `{build.library_discovery_flags}`: a series of `-D` flags, one set per library resolved during library discovery. It
  is available to sketch and library compile recipes only; it is deliberately not exposed to core-compile recipes. This
  property was added in Arduino CLI 1.6.0.<br />
  For each library, the property contains a flag with the form `-DFOUND_<SLUG>_LIB=0x<hex-version>`, and one with the
  form `-DFOUND_<SLUG>_LIB_<SOURCE>=1`.
  - `<SLUG>` is the library name converted to a valid uppercase C identifier (runs of characters outside `[A-Za-z0-9_]`
    become a single `_`).
  - `<hex-version>` is the library's version, in the form of a 24-bit value assembled from the major/minor/patch
    components (`MMmmpp`). Each component is clamped to `0xFF`. Pre-release/build metadata components are discarded. An
    unknown or unparseable version is encoded as `1`.
  - `<SOURCE>` identifies where the library was found:
    - `IN_IDE`: Arduino IDE's built-in libraries.
    - `IN_PLATFORM`: libraries bundled with the platform.
    - `IN_PROFILE`: libraries specified as a dependency by the build profile.
    - `IN_SKETCHBOOK`: libraries located under `directories.user`, or under a path specified via a `--libraries` flag.
    - `IN_SPECIFIED_PATH`: libraries under a path specified via a `--library` flag.
```

- Adjust wording and structure for readability
- Use a bulleted list for documentation of each component to avoid a "wall of text" situation.
- Remove superfluous explanation of why property is not available for core compilation (this would make sense as a comment in the codebase, but not in the user-facing specification)
- Remove info re: lack of source flag for libraries from unrecognized locations. This should be an "impossible" event, and thus not worth documenting.
- Correct terminology ("arduino-cli" -> "Arduino CLI", "core" -> "platform", "sketch profile" -> "build profile").
- Document discarding of pre-release and build metadata components of library version.

per1234 (dismissed):

docs/platform-specification.md
pillo79: Thanks Per, applied! Suggestion on other name changes or logic improvements are welcome. I'd like to change SPECIFIED_PATH especially, but UNMANAGED doesn't look right to me and I found no better alternatives.

PR #3287, opened 2026-08-27
Add "upload-to-file" capability

cmaglie:

      before creating one)
- [X] The PR follows

This PR adds a new "upload-to-file" flag to the upload command. Instead of performing a normal upload, the upload command will write the resulting firmware executables into an archive we call a "firmware file". This archive will also contain the board's metadata, the uploader tools needed for the upload, and the recipes to upload the firmware.

In theory, only using the firmware file, we could make a hypothetical universal uploader which:

* Reads the required tools from the firmware file and downloads and installs them.
* Reads the identification properties from the board's metadata so it can identify the upload port through the corresponding pluggable discovery.
* Performs the upload using the board's recipes recorded in the firmware file.

To accomplish this the current Arduino platforms needs to be updated, in particular we need them to explicity declare:

* Which artifact files, obtained from the compilation, are actually needed for the upload.
* Which configuration files (from the platform/variant or from the compilation) are needed for the upload.
* Something else that I'm surely forgetting to list here.

For platforms that do not explicitly provide the above information, we may make some guesses, but I already know that some upload tools will fail because of missing configurations.

### DIFF
--- a/commands/service_board_details.go
+++ b/commands/service_board_details.go
@@ -20,9 +20,11 @@ import (
 
 	"github.com/arduino/arduino-cli/commands/cmderrors"
 	"github.com/arduino/arduino-cli/commands/internal/instances"
+	"github.com/arduino/arduino-cli/internal/arduino/cores"
 	"github.com/arduino/arduino-cli/internal/arduino/utils"
 	"github.com/arduino/arduino-cli/pkg/fqbn"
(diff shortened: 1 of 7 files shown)


PaulStoffregen:
Is there any preliminary documentation for format of this new this file?

cmaglie:
Not yet, we are iterating to balance some pros and cons.

